PR #15, opened 2018-10-08
Initial activation factory description

DefaultRyan:
Holding off on the function reference until Harry's pull request completes.


docs/XSPEC03 - Platform Abstraction Layer.md
```diff
@@ -470,3 +470,57 @@ Calling this function converts the mutable buffer to an immutable **XlangString*
...
+    GUID const& iid,
+    void** factory
+);
+```
+
+#### Parameters
+- class_name - The name of the class.
+- iid - The unique identifier (GUID) of the factory interface being requested.
+- factory - The out parameter receiving the factory
+-
```
kennykerr: Extra bullet

design_notes/XDN03 - Platform Support for Activation Factories.md
```diff
@@ -0,0 +1,76 @@
...
+* The PAL can not load the mapped component.
+* The PAL loads the component, but can not locate the exported **xlang_lib_get_activation_factory** function to retrieve the component's factories.
+* The component itself fails to return the requested factory.
+
+## Mapping classes to components
+
+This functionality is expected to evolve as xlang's application model matures, with new behaviors and abstractions being added as appropriate.
+
+Upon the first request for an activation factory, or in other words, the first time a class name is passed to **xlang_get_activation_factory**, the PAL will attempt to load the owning component in the following manner:
+* Obtain the innermost namespace scope from the class name
```
devhawk: This is basically the same algorithm windows uses to locate metadata at runtime in winmd files.

design_notes/XDN03 - Platform Support for Activation Factories.md
```diff
@@ -0,0 +1,76 @@
...
+* The PAL can not map the class name to a component.
+* The PAL can not load the mapped component.
+* The PAL loads the component, but can not locate the exported **xlang_lib_get_activation_factory** function to retrieve the component's factories.
+* The component itself fails to return the requested factory.
+
+## Mapping classes to components
+
+This functionality is expected to evolve as xlang's application model matures, with new behaviors and abstractions being added as appropriate.
+
+Upon the first request for an activation factory, or in other words, the first time a class name is passed to **xlang_get_activation_factory**, the PAL will attempt to load the owning component in the following manner:
```
kennykerr: This implies that the PAL may be stateful and cache the results of these requests. I don't think we should imply that or even suggest that as a good optimization. Caching within combase or within components in the OS has proven to be very ineffective compared with caching within a language projection.

DefaultRyan: There's no intention of caching the actual factories - just the results of the dll lookup. But I'll remove that wording to avoid sowing confusion.

PR #20, opened 2018-10-10
Added code owners to enforce reviewer policy

Scottj1s:
Looking for general agreement on this idea, which Kenny asked for.  GitHub code owners are required reviewers for any PRs affecting files under their respective paths of ownership.  Later rules overrule earlier ones, so the one on line 5 is the default.


Scottj1s:
I think the feature requires approval by the code owner (and not merely ensures they are automatically added as reviewers).  We'll experiment and if too onerous, turn it off.

PR #33, opened 2018-10-11
Fixed Case Insensitive Example

wizarrc:
The example provided would be allowed because the types are different. Fixed the example by only changing the case of the namespace while keeping the type the same.

### DIFF
--- a/docs/XSPEC01 - Type System Specification.md
+++ b/docs/XSPEC01 - Type System Specification.md
@@ -23,7 +23,7 @@ Except for interfaces, all Xlang types must have public visibility. Xlang interf
 
 Xlang does not support nested types. No Xlang type can enclose another type. No Xlang type can be nested inside another type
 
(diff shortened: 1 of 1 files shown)


docs/XSPEC01 - Type System Specification.md
```diff
@@ -23,7 +23,7 @@ Except for interfaces, all Xlang types must have public visibility. Xlang interf
 
 Xlang does not support nested types. No Xlang type can enclose another type. No Xlang type can be nested inside another type
 
-Xlang namespaces and type names are case preserving but insensitive. This means that you cannot have namespaces or type names that vary only by case. For example: you cannot have Foo.SomeType and foo.AnotherType nor can you have Foo.SomeType and Foo.someType.
```
BenJKuhn: This is a great catch. Can you also add a comment that "Foo.SomeType and foo.AnotherType" can exist and are considered part of the same namespace?

wizarrc: @BenJKuhn I added your suggestion.

Issue #37, opened 2018-10-15
Pythonic Python Experience


devhawk:
Currently, pywinrt (aka the python projection generation tool) emits code that is usable from python, but does not provide a pythonic experience - i.e. it does not feel natural and familiar to a python developer. This issue tracks the work we will do to generate a more pythonic experience by wrapping the generated native types in pure python.

At a minimum, expected to project the following WinRT constructs as the corresponding pythonic constructs:

| WinRT Construct   | Python Construct                        |
|-----------------  |-------------------                      |
| Namespaces        | pythonic experience TBD                 |    
| IStringable       | \_\_str__ support                       |
| IClosable         | with statement support                  |
| Collections       | sequence protocol and \_\_len__ support |
| IAsync*           | await support                           |
| get/put methods   | property support                        |
| add/remove methods | pythonic experience TBD                |

nvictor:
so happy to see the wip of the python projection!

PR #38, opened 2018-10-15
Python Milestone: Event support

devhawk:
This PR adds event support to the Python projection tool. 

* Events are projected directly as add and remove methods. Clearly, this is not pythonic, but "natural and familar" python developer experience will be addressed in the future (Tracked by #37)

Example:
```python
def onMapChanged(sender, args): 
  print("added key {}".format(args.get_Key())

m = _pyrt.StringMap()
m.Insert("king", "arthur") # nothing prints, event not connected yet
token = m.add_MapChanged(onMapChanged)
m.Insert("sir", "robin") # prints "added key sir"
m.remove_MapChanged(token)
m.Insert("tim", "the enchanter") # nothing prints, event remvoed
```

Other changes
* Reworked projection code to share more generator logic between classes and interfaces 
* Support projecting methods on interfaces from required interfaces
* Added support for enum values in structs
* Stopped special casing properties - projecting instead as get/put methods. As with events, this is clearly not pythonic but will be addressed later (Tracked by #37)
* optimized Python signature for special WinRT methods (prop get/set, event add/remove). 

Note, code generated by the test_python build target [attached](https://github.com/Microsoft/xlang/files/2480484/devhawk-pyevents2-output.zip). Please let me know if there is a better way to review the generated code. (Gist? Dedicated repo? Orphan branch in this repo?)


src/library/impl/meta_reader/table.h
```diff
@@ -334,8 +334,7 @@ namespace xlang::meta::reader
 
         bool operator==(row_base const& other) const noexcept
         {
-            XLANG_ASSERT(m_table == other.m_table);
-            return m_index == other.m_index;
+            return (m_table == other.m_table) && (m_index == other.m_index);
```
kennykerr: Why this change? Rows are basically iterators and the assertion says that comparing iterators from different tables is an invalid operation. This is much the same as comparing iterators from different containers in the C++ standard library.

devhawk: when collecting required interfaces, I needed a way to [test if two given types were the same.](https://github.com/Microsoft/xlang/blob/1376a084e300b9f512d939825293199bd96987c4/src/tool/python/helpers.h#L214). If there's a more appropriate way, please let me know.

src/tool/python/code_writers.h
```diff
@@ -97,106 +97,84 @@ namespace xlang
...
 
-            w.write_each<write_getset_table_prop_row>(get_instance_properties(type));
-            //w.write_each<write_class_getset_table_row_event>(get_instance_events(type));
+        w.write("\nstatic PyGetSetDef @_getset[] = {\n", type.TypeName());
 
-            w.write_each< write_getset_table_field_row>(type.FieldList());
-            w.write("    { nullptr }\n};\n");
+        for (auto&& field : type.FieldList())
+        {
+            // TODO: remove const_cast once pywinrt is updated to target Python 3.7.
```
zooba: Any reason why we wouldn't target both? It may mean generating some ugly `#if` checks (or typedefs), but it's worth it for the broader compatibility.

src/tool/python/code_writers.h
```diff
@@ -635,44 +569,43 @@ static PyType_Spec @_Type_spec =
...
-                prop.Parent().TypeName(),
-                property_methods.set.Name(),
-                bind<write_winrt_wrapper>(prop.Parent()),
-                bind<write_property_set_body>("self->obj", prop.Name(), prop.Type().Type()));
+            w.write(R"(    else if (arg_count == -1)
+    {
+        return nullptr; 
+    }
+
+    PyErr_SetString(PyExc_RuntimeError, "Invalid parameter count");
```
zooba: Would suggest matching the usual error type and message (though if/when you switch to `PyArg_ParseTupleAndKeywords` for parsing, you'll get this for free).

src/test/python/testcollections.py
```diff
@@ -0,0 +1,52 @@
...
+                self.assertEqual(sender.get_Size(), 2)
+                self.assertTrue(sender.HasKey("dr"))
+                self.assertTrue(sender.HasKey("hello"))
+                
+                loop.call_soon_threadsafe(asyncio.Future.set_result, future, True)
+
+            m = _pyrt.StringMap()
+            m.Insert("hello", "world")
+            token = m.add_MapChanged(onMapChanged)
+            m.Insert("dr", "who")
```
zooba: The callback should trigger here synchronously as well, right? Or does it only work in an async context?

devhawk: Events aren't guaranteed to come in on the same thread, though i think this one might. However, I wrote this test asynchronously so I could await on the event firing. Let me know if there's a more pythonic way of doing that

src/tool/python/code_writers.h
devhawk: It already compiles for both. Haven't addressed packaging yet so I'll keep generating 3.6 compatible code for now

src/tool/python/code_writers.h
devhawk: It looks like `PyArg_ParseTupleAndKeywords` uses `PyErr_BadInternalCall` which seems like a mistake. Will switch over to setting `PyExc_TypeError` instead of `PyExc_RuntimeError` to mirror the behavior of `PyErr_BadArgument` (but calling directly so I can pass my own message instead of the hardcoded one passed by `PyErr_BadArgument`)

devhawk:
I spun up a repo to store pywinrt generated code. Generated code for this PR is here: https://github.com/devhawk/pywinrt-output/tree/events

src/tool/python/helpers.h
```diff
@@ -188,52 +201,232 @@ namespace xlang
...
+        case TypeDefOrRef::TypeSpec:
+            collect_required_interfaces(w, sigs, index.TypeSpec().Signature().GenericTypeInst());
+            break;
+        }
+    }
+
+    auto get_required_interfaces(TypeDef const& type)
+    {
+        writer w;
+        w.debug_trace = true;
```
kennykerr: Probably shouldn't leave this on in committed code as it can get quite noisy.

src/tool/python/helpers.h
```diff
@@ -188,52 +201,232 @@ namespace xlang
...
+                method_map[method.Name()].push_back(method_info{ method, std::vector<std::string> {} });
+            }
+        }
+        else if (category == category::interface_type)
+        {
+            for (auto&& info : get_required_interfaces(type))
+            {
+                for (auto&& method : info.type.MethodList())
+                {
+                    //method_map.try_emplace(method.Name(), std::vector<method_info> {});
```
kennykerr: Remove comment?

src/tool/python/code_writers.h
zooba: It should only do that if _you_ provide the wrong arguments. It shouldn't do that when the caller provides the wrong ones (otherwise all of Python would be raising the wrong error).

PR #41, opened 2018-10-19
Activation factories, v1

DefaultRyan:
This contains the initial implementation of activation factories as described in the design doc. There's only a Windows implementation right now, but I've attempted to structure the code to easily accommodate other platforms using dlopen and dlsym instead of LoadLibrary and GetProcAddress.


src/platform/activation_abi.cpp
```diff
@@ -0,0 +1,69 @@
...
+    void** factory
+) XLANG_NOEXCEPT
+try
+{
+    if (!class_name)
+    {
+        xlang::throw_result(xlang_error_invalid_arg);
+    }
+
+    auto const encoding = xlang_get_string_encoding(class_name);
```
kennykerr: Looking forward to revisiting the string question now that we've made the decision to split from WinRT. This is a good example of what developers are likely to do all over the place in the hopes of optimizing string performance. _Not_ picking an encoding does not produce a developer-friendly platform.

BenJKuhn: Agreed, this is a bit awkward, but I also feel that this layer is an outlier. Most people will be dealing with their native string type for the programming language / application, and the conversion won't be as visible. As I noted to Ryan, this layer could accept raw character strings as well since this is only called through C. It's not a projected API.

At the end of all the processing, it eventually calls LoadLibraryW, which requires at least a substring to be converted to UFF-16 anyway. On other platforms, it might be UTF-8 at the bottom. As such, I think the code (at least for Windows) should be simplified to just do processing in UTF-16, though the implementation could be templatized to prefer UTF-8 for platforms where the module loading function uses that format.

src/platform/activation_abi.cpp
```diff
@@ -0,0 +1,69 @@
...
+        for (auto current_namespace = enclosing_namespace(to_string_view<char_type>(class_name));
+            !current_namespace.empty();
+            current_namespace = enclosing_namespace(current_namespace))
+        {
+            xlang_pfn_lib_get_activation_factory pfn = try_get_activation_func(current_namespace);
+            if (pfn)
+            {
+                void* factory{};
+                xlang_result result = (*pfn)(class_name, iid, &factory);
+                if (result == xlang_error_ok)
```
Scottj1s: fyi - the PAL doc has a some mixed case error definitions (e.g., Xlang_INVALID_ARG)

src/platform/activation_abi.cpp
```diff
@@ -0,0 +1,69 @@
...
+                {
+                    return factory;
+                }
+                else if (result != xlang_error_class_not_available)
+                {
+                    throw_result(result);
```
Scottj1s: what errors do you imagine should short-circuit out enclosing namespace checks?  is this a definitive "we have the factory, but it doesn't work"?

src/platform/activation_abi.cpp
Scottj1s: At a minimum, a type function to convert encoding to filesystem char type "en passant" would be nice - probably many opportunities to reuse it.  I still like the idea of tailoring string conversion to both projection and component/PAL (and minimizing it), rather than just "picking one"

src/platform/platform_activation.h
```diff
@@ -0,0 +1,32 @@
+#pragma once
+
+#include "pal.h"
+#include <string_view>
+
+namespace xlang::impl
+{
+#if XLANG_PLATFORM_WINDOWS
```
Scottj1s: should this be in pal.h?

src/platform/activation_abi.cpp
```diff
@@ -0,0 +1,69 @@
...
+#include "platform_activation.h"
+
+namespace xlang::impl
+{
+    template <typename char_type>
+    void* get_activation_factory(
+        xlang_string class_name,
+        xlang_guid const& iid)
+    {
+        for (auto current_namespace = enclosing_namespace(to_string_view<char_type>(class_name));
```
Alovchin91: Apparently, here you try to get the enclosing namespace, by cutting the class name from `class_name` right away.

But [XDN03 - Platform Support for Activation Factories](/Microsoft/xlang/blob/master/design_notes/XDN03%20-%20Platform%20Support%20for%20Activation%20Factories.md) design note has the following example:

> ...imagine an app requests a factory for the class *MyComponent.Feature.Widget*...
> * The PAL searches for the library *MyComponent.Feature.Widget.(dll/so)* (on Windows/Linux, respectively).
> * This library is not found in the application's directory, so it moves up one namespace level ...

Is it the design note that should be updated, or should you initialize `current_namespace` with just `to_string_view<char_type>(class_name)`?

src/platform/win32_activation.cpp
```diff
@@ -0,0 +1,69 @@
...
+
+namespace xlang::impl
+{
+    xlang_pfn_lib_get_activation_factory try_get_activation_func(
+        std::basic_string_view<wchar_t> module_namespace)
+    {
+        HMODULE module{};
+
+        constexpr auto file_ext{ L".dll"sv };
+        if (module_namespace.size() + file_ext.size() < MAX_PATH)
```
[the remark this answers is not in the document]
Scottj1s: or just use _malloca if you want a variable buffer that can be allocated from stack or heap as appropriate

src/test/components/Component/module.cpp
```diff
@@ -0,0 +1,39 @@
...
+        *factory = nullptr;
+
+        char16_t const* buffer{};
+        uint32_t length{};
+        winrt::check_hresult(xlang_get_string_raw_buffer_utf16(class_name, &buffer, &length));
+        std::u16string_view name{ buffer, length };
+        
+        if (name == u"Component.Class"sv)
+        {
+            // TODO: Unify xlang guid and winrt guid
```
Scottj1s: have we given any thought to this?  usings in pal.h?

src/platform/activation_abi.cpp
devhawk: Created Issue #53 it track

src/platform/win32_activation.cpp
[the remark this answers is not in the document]
devhawk: opened #54 to track

src/platform/platform_activation.h
devhawk: PAL.h has xlang_get_activation_factory. Pretty sure this file is explicitly part of the implementation.

src/test/components/Component/module.cpp
devhawk: Since this code would eventually be generated automatically by the language projection, much as C++/WinRT generates WINRT_GetActivationFactory implementation today, does it matter?

src/platform/activation_abi.cpp
devhawk: Good catch @Alovchin91. @DefaultRyan is out of the office currently so I'm finishing off this PR. I'm pretty sure XDN03 is wrong here, but I opened #55 for @DefaultRyan to look at when he's back in the office.

PR #43, opened 2018-10-19
Python/WinRT: Python Iterator support and _from function

devhawk:
This PR adds two major projection features:

* the ability to pass [Python Iterator types](https://docs.python.org/3/library/stdtypes.html#iterator-types) for WinRT IIterable<T> parameters 
   * Passing [Python Sequence Types](https://docs.python.org/3/library/stdtypes.html#sequence-types-list-tuple-range)as WinRT IVector and [Python Mapping Types](https://docs.python.org/3/library/stdtypes.html#mapping-types-dict) as WinRT IMap will come in a later update
* Exposes a static ~~_as~~ _from method on projected classes and concrete interfaces to enable type casting (i.e. QueryInterface)

Other changes
* reworked python type mapping to eliminate linker errors when a projected type has parameters of an unprojected type
   * UniversalApiContract is large, so for projection developer inner loop it is often preferable to only project a small subset of the overall contract. This feature enables more of UniversalApiContract to be projected piecemeal *during projection development*
* refactored WinRT <--> Python converter code in pybase.h to enable specializations (used to enable projecting Python iterators as IIterable<T>)
* fixed array parameter categorization and updated generation of array parameters to actually compile.
   * Note, methods with array parameters still don't work, but at least they compile so you can generate code for a type that has array parameters and all the non-array parametered methods will still work as expected 

Sample projected code (with diff vs. previous Python PR!) is available here: https://github.com/devhawk/pywinrt-output/tree/iterables

### DIFF
--- a/src/test/python/testasfunction.py
+++ b/src/test/python/testasfunction.py
@@ -1,0 +1,17 @@
+import find_projection
+import _pyrt
+import unittest
+
+class TestQueryInterface(unittest.TestCase):
+    def test_as_function(self):
+        propset = _pyrt.PropertySet()
+        propset.Insert("strmap", _pyrt.StringMap())
+        self.assertTrue(propset.HasKey("strmap"))
+        o = propset.Lookup("strmap")
+        strmap = _pyrt.StringMap._as(o)
+        self.assertEqual(type(strmap), _pyrt.StringMap)
+
+if __name__ == '__main__':
+    _pyrt.init_apartment()
+    unittest.main()
+    _pyrt.uninit_apartment()
--- a/src/tool/python/code_writers.h
+++ b/src/tool/python/code_writers.h
@@ -518,18 +518,20 @@ static PyType_Spec @_Type_spec =
 
     void write_type_query_interface(writer& w, TypeDef const& type)
     {
-        if (is_ptype(type))
+        if (is_ptype(type) || type.Flags().Abstract())
         {
             return;
         }
 
         auto format = R"(
-static PyObject* %__query_interface(PyObject* /*unused*/, PyObject* args)
+static PyObject* %__query_interface(PyObject* /*unused*/, PyObject* arg)
 {
-    return nullptr;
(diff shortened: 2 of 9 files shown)


src/test/python/testasfunction.py
```diff
@@ -0,0 +1,17 @@
...
+import _pyrt
+import unittest
+
+class TestQueryInterface(unittest.TestCase):
+    def test_as_function(self):
+        propset = _pyrt.PropertySet()
+        propset.Insert("strmap", _pyrt.StringMap())
+        self.assertTrue(propset.HasKey("strmap"))
+        o = propset.Lookup("strmap")
+        strmap = _pyrt.StringMap._as(o)
```
BenJKuhn: Semantics from _as seem backwards. As on comptr, etc. convert to a type specified in a template parameter. I would think...

`strmap = _pyrt.StringMap._from(o);`

… reads better. Thoughts?

devhawk: Yup, makes sense. Will change.

src/tool/python/code_writers.h
```diff
@@ -1317,6 +1379,7 @@ static PyObject* @_get_%(%* self, void* /*unused*/)
                 field.Parent().TypeName(),
                 field.Name(),
                 bind<write_winrt_wrapper>(field.Parent()),
+                bind<write_struct_property_get_variable>(field),
                 bind<write_struct_property_get_variable>(field));
```
zooba: Is this an intentional duplicate?

devhawk: Yes, but it seems kind of ugly looking at it again. I switched away from using if constexpr branches in favor of just having converter<T> type specializations. However, that meant I couldn't use type deduction anymore. The variable is duplicated here because I'm using declype instead of specifying the type directly.

example: `return py::converter<decltype(self->obj.Height)>::convert(self->obj.Height);`

I kept the convert_to<T> helper, I'm thinking I should have kept the convert<T> helper as well. Thoughts?

PR #44, opened 2018-10-22
Typo

Suniti13:


devhawk:
I think succinctly is correct here.

Issue #47, opened 2018-10-24
implement convert_string in PAL for non-windows platforms


devhawk:
The two convert_string PAL functions (UTF-8 -> UTF16 and UTF16 -> UTF8) are only implemented for windows. The version in common_string_convert.cpp simply throw today.

DefaultRyan:
#81 addressed this.

PR #48, opened 2018-10-24
Fix ubuntu build

devhawk:
changes to enable xlang tools, PAL and to compile on Ubuntu 18.04 with clang 6. Tested via Windwos Susbsytem for Linux.

Note, convert_string functions in PAL remain unimplemented on non-windows platforms. As such, 4 of the PAL tests fail on Ubuntu. #47 opened to track the issue, assigned to @DefaultRyan.

### DIFF
--- a/src/platform/common_string_convert.cpp
+++ b/src/platform/common_string_convert.cpp
@@ -1,0 +1,39 @@
+#include "pal_internal.h"
+#include "string_convert.h"
+
+namespace xlang::impl
+{
+    uint32_t get_converted_length(char16_t const* input_str, uint32_t input_length)
+    {
+        return convert_string(input_str, input_length, nullptr, 0);
+    }
+
+    uint32_t get_converted_length(xlang_char8 const* input_str, uint32_t input_length)
+    {
+        return convert_string(input_str, input_length, nullptr, 0);
+    }
+
+    uint32_t convert_string(
+        char16_t const* input_str,
+        uint32_t input_length,
+        xlang_char8* output_buffer,
+        uint32_t buffer_length)
+    {
+        static_assert(sizeof(xlang_char8) == sizeof(char));
+ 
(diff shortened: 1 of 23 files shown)


src/platform/common_string_convert.cpp
```diff
@@ -0,0 +1,39 @@
...
+
+    uint32_t convert_string(
+        char16_t const* input_str,
+        uint32_t input_length,
+        xlang_char8* output_buffer,
+        uint32_t buffer_length)
+    {
+        static_assert(sizeof(xlang_char8) == sizeof(char));
+ 
+        // TODO: implement convert_string for non-windows platforms (#47)
```
Scottj1s: Remove comment?

src/platform/common_string_convert.cpp
```diff
@@ -0,0 +1,39 @@
...
+    
+    uint32_t convert_string(
+        xlang_char8 const* input_str,
+        uint32_t input_length,
+        char16_t* output_buffer,
+        uint32_t buffer_length)
+    {
+        static_assert(sizeof(xlang_char8) == sizeof(char));
+
+        // TODO: implement convert_string for non-windows platforms (#47)
```
Scottj1s: ditto

src/platform/common_string_convert.cpp
devhawk: I don't think it's a good idea to remove the TODO comment

src/platform/common_string_convert.cpp
devhawk: ditto :)

devhawk:
Note, I'm merging this PR w/o @DefaultRyan approval, even though he's the PAL code owner. He's unavailable right now + he actually did most of the PAL changes in a different branch that I cherry picked into this one.

PR #51, opened 2018-10-24
Add abi executable for generating MIDL compatible headers

dunhor:
This is part 1. It's mostly functional. Not 100% finished and somewhat rough at parts, but per Ben's request... The only known missing part is generic *_impl definitions, which I'm working through now. Thus far I've been using the 'Windows.AI.MachineLearning', 'Windows.ApplicationModel.Background', 'Windows.Foundation', and 'Windows.UI.Notifications' namespaces for validation along with a custom written idl for validating. That's quite a tedious process, but, with the exception of generics in the Windows.Foundation.h file, the results are effectively equivalent to MIDL. The only known differences:

* MIDL recursively processes contract version definitions; here we only declare what exists in our namespace dependencies
* MIDL forward declares class dependencies (e.g. 'CoreWindow') for function arguments. Since the actual argument is the default interface (e.g. 'ICoreWindow'), there's no real need to forward declare the class, so we don't do so here
* We forward declare all class types, enums, and structs defined in this namespace. Technically unnecessary and MIDL will only do so for those used as function arguments (see above), but also not harmful
* Formatting and ordering are differently, obviously. Formatting was kept as close as possible for easier validation, but sometimes it was better to change it. I'm also using a modified version of MIDL for validating that more closely matches the formatting here

Also adds WinDBG javascript visualizers for a number of xlang::meta::reader types


.gitignore
```diff
@@ -26,13 +26,13 @@ bld/
 
 # Visual Studio 2015/2017 cache/options directory
 .vs
-.vscode
+.vscode
```
dunhor: I assume I removed trailing spaces here, but anyone know how to validate? GitHub doesn't seem to have the option to show whitespace, and in particular whitespace changes

dunhor: Well, at least you can hide whitespace changes...

src/tool/abi/abi_writer.cpp
```diff
@@ -0,0 +1,819 @@
+#include "pch.h"
+
+#include <cctype>
+#include <cstring>
+#include <iostream> // TODO: REMOVE
```
dunhor: nts

src/tool/abi/abi_writer.cpp
```diff
@@ -0,0 +1,819 @@
...
+
+    // NOTE: MIDLRT will declare any generic type that references a type in this namespace, even if no function in this
+    //       namespace references that generic type. We won't replicate that here for simplicity since most, if not all
+    //       consumers that wish to reference such a type will include the header that contains such a function
+}
+
+void writer::initialize_dependencies(TypeDef const& type)
+{
+    using namespace std::literals;
+
```
dunhor: nts

src/tool/abi/abi_writer.cpp
```diff
@@ -0,0 +1,819 @@
...
+    for (auto const& type : m_members.delegates)
+    {
+        // Generics require definitions
+        if (distance(type.GenericParam()) == 0)
+        {
+            write_forward_declaration(*this, type, m_genericArgStack);
+        }
+    }
+
+    bool const isFoundationNamespace = m_namespace == foundation_namespace;
```
dunhor: nts

src/tool/abi/abi_writer.cpp
```diff
@@ -0,0 +1,819 @@
...
+    for (auto const& sig : type.GenericArgs())
+    {
+        xlang::visit(sig.Type(),
+            [&](coded_index<TypeDefOrRef> const& t)
+            {
+                handle_typedef_or_ref(t);
+            },
+            [&](GenericTypeInstSig const& t)
+            {
+                // TODO: push?
```
dunhor: >// TODO: push? [](start = 16, length = 14)

nts

src/tool/abi/code_writers.h
```diff
@@ -0,0 +1,788 @@
...
+    using namespace xlang::text;
+
+    auto const typeCategory = get_category(type);
+    auto const categoryString = [&]()
+    {
+    switch (typeCategory)
+    {
+        case category::interface_type: return "Interface"sv;
+        case category::class_type: return "Class"sv;
+        case category::enum_type: // TODO: return "Enum"sv; ??? 'Struct' is what MIDL does...
```
dunhor: >// TODO: return "Enum"sv; ??? 'Struct' is what MIDL does... [](start = 34, length = 59)

nts

src/tool/abi/main.cpp
```diff
@@ -0,0 +1,177 @@
...
+
+            if (f.includes(members))
+            {
+                group.add([&]()
+                {
+                    write_abi_header(ns, c, members, config);
+                });
+            }
+        }
+
```
dunhor: join

Scottj1s:
> src/test/python/__pycache__/

did you mean to remove this?

---
Refers to: .gitignore:340 in 6360495. [](commit_id = 6360495e0b8a169bfadd1971de901758e4bf84a2, deletion_comment = True)

Scottj1s:
> src/test/python/__pycache__/

nm - CodeFlow wasn't showing me iteration 25 (and it doesn't scroll!)

---
In reply to: [432838941](https://github.com/Microsoft/xlang/pull/51#issuecomment-432838941) [](ancestors = 432838941)

---
Refers to: .gitignore:340 in 6360495. [](commit_id = 6360495e0b8a169bfadd1971de901758e4bf84a2, deletion_comment = True)

src/tool/abi/main.cpp
```diff
@@ -0,0 +1,179 @@
...
+{
+    auto out{ absolute(args.value("output", "abi")) };
+    create_directories(out);
+    out += path::preferred_separator;
+    return out.string();
+}
+
+void print_usage()
+{
+    puts("Usage...");
```
Scottj1s: >puts("Usage..."); [](start = 4, length = 17)

don't forget

src/tool/abi/abi_writer.h
```diff
@@ -0,0 +1,260 @@
...
+
+    template <typename Int>
+    void write(format_hex<Int> val)
+    {
+        write_value(val.value);
+    }
+
+    void write(indent value)
+    {
+        for (std::size_t i = 0; i < (m_indentation + value.additional_indentation); ++i)
```
Scottj1s: >m_indentation [](start = 37, length = 13)

I think Harry is refactoring to move indent guard into the base library

src/tool/abi/main.cpp
dunhor: I was wanting to follow existing format to be consistent, but the existing tools just do the same thing, so I copied :/

---
In reply to: [227974213](https://github.com/Microsoft/xlang/pull/51#discussion_r227974213) [](ancestors = 227974213)

src/tool/abi/abi_writer.h
dunhor: Cool 👍🏼

---
In reply to: [227979165](https://github.com/Microsoft/xlang/pull/51#discussion_r227979165) [](ancestors = 227979165)

src/tool/abi/main.cpp
kennykerr: I think we'll probably just write a universal usage generator rather than maintaining various individual blurbs. This is a reasonable placeholder until we get that working.

src/tool/abi/common.h
```diff
@@ -0,0 +1,342 @@
...
+    }
+};
+
+namespace xlang
+{
+    template <typename... T> visit_overload(T...) -> visit_overload<T...>;
+}
+
+template <typename Visitor>
+inline auto visit(xlang::meta::reader::coded_index<xlang::meta::reader::TypeDefOrRef> const& type, Visitor&& visitor)
```
kennykerr: Function templates are implicitly inline.

src/tool/abi/common.h
```diff
@@ -0,0 +1,342 @@
...
+    using namespace xlang::meta::reader;
+    switch (type.type())
+    {
+    case TypeDefOrRef::TypeDef: return visitor(type.TypeDef()); break;
+    case TypeDefOrRef::TypeRef: return visitor(type.TypeRef()); break;
+    case TypeDefOrRef::TypeSpec: return visitor(type.TypeSpec().Signature().GenericTypeInst()); break;
+    }
+}
+
+inline constexpr std::pair<std::string_view, std::string_view> decompose_type(std::string_view typeName) noexcept
```
kennykerr: constexpr functions are implicitly inline.

src/scripts/mdExts.js
```diff
@@ -0,0 +1,3953 @@
+"use strict";
```
kennykerr: This is spectacular. You should consider refactoring this into a standalone winmd reader library for JavaScript. 😉

src/tool/abi/common.h
dunhor: heh, habit mostly. Shouldn't be harmful

PR #60, opened 2018-10-25
Testing, C++/WinRT compat, optimizations, and instance-based fast ABI

kennykerr:
This update lays the groundwork for the more fine-grained approach to testing. The idea is that the `component.idl` will be fleshed out to cover all of the code gen scenarios that all xlang tools will be validated against. The `test/cpp` project has quite a feature layers or dimensions of testing and validating against the resulting metadata. I don't expect all tools to provide this level of testing, but at the very least each tool should validate that they can consume/call the types from component.idl. 

This update also introduces improved parity with cppwinrt.exe as well as the basis for the fast ABI (now based on exclusive interface folding).

You'll also notice a few new optimizations that substantially reduces the size of call site shims (consumers) and uses the `take_ownership_from_abi` marker.


src/CMakeLists.txt
```diff
@@ -78,9 +78,14 @@ if (MSVC)
         string(APPEND CMAKE_EXE_LINKER_FLAGS_${build_type} " /LTCG:INCREMENTAL /OPT:REF")
     endforeach()
     
-    add_compile_options(/await)
     add_compile_options(/permissive-)
-    string(APPEND CMAKE_CXX_FLAGS_DEBUG " /JMC")
```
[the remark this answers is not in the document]
kennykerr: I was trying to limit the differences between compilers in the cmake file as much as possible. clang-cl attempts to emulate the cl command line but it doesn't do so very reliably. For example, it doesn't accept /JMC. 

I'd prefer to just fork and call clang directly with clang-specific args but cmake/ninja won't accept that at all for some reason.

src/library/cmd_reader.h
```diff
@@ -172,5 +155,22 @@ namespace xlang::cmd
 
             return files;
         }
+
+    private:
```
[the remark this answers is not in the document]
kennykerr: No, this works. Notice that I removed the auto return type. That's why clang was complaining - it couldn't deduce the return type without a definition.

src/CMakeLists.txt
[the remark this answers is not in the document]
devhawk: https://stackoverflow.com/a/38174328

src/library/cmd_reader.h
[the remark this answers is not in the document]
devhawk: cool. Lets make sure we're testing on Clang/Linux (and clang/windows once we get that up and working) manually until we get automatic builds of those configurations working

src/library/impl/meta_reader/schema.h
```diff
@@ -48,7 +48,7 @@ namespace xlang::meta::reader
 
         auto Flags() const
         {
-            return TypeAttributes{ get_value<uint32_t>(0) };
+            return TypeAttributes{{ get_value<uint32_t>(0) }};
```
devhawk: Does this change to using the initalizer list ctor? Looking at the TypeAttributes class, I don't get why you made this change

src/library/cmd_reader.h
```diff
@@ -172,5 +155,22 @@ namespace xlang::cmd
...
+
+        std::vector<option>::const_iterator find(std::vector<option> const& options, std::string_view const& arg)
+        {
+            for (auto current = options.begin(); current != options.end(); ++current)
```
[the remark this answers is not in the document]
kennykerr: This is needed because `find` returns the iterator. Range-based for loops hide the iterator from view.

src/library/impl/meta_reader/schema.h
kennykerr: Hallway conversation: the code is basically initializing the base where that field actually resides. VC allows this but Clang (in strict mode) does not.

src/CMakeLists.txt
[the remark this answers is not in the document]
kennykerr: I think that depends on the Clang compiler, or at least the build configuration, that VS shipped (which is no longer supported). Anyway, I couldn't get it to work. If you can figure out how to call clang directly I'd love that.

src/library/impl/meta_reader/schema.h
kennykerr: This basically initializes the base type where the field actually resides. VC allows the former but Clang, in strict mode, does not.

src/tool/cpp/strings/base_string.h
```diff
@@ -199,7 +203,7 @@ WINRT_EXPORT namespace winrt
 
         bool empty() const noexcept
         {
-            return 0 == size();
+            return !m_handle;
```
kennykerr: Nice improvement suggested by Raymond Chen to improve inlining.

src/test/cpp/compare/out/winrt/Component.Fast.h
```diff
@@ -0,0 +1,304 @@
+﻿// WARNING: Please don't edit this file. It was generated by C++/WinRT v2.0.000000.0
+#pragma once
+#include "winrt/impl/Component.2.h"
+#include "winrt/impl/Component.Fast.2.h"
+#include "winrt/Component.h"
+namespace winrt::impl
+{
+    template <typename D> hstring consume_Component_Fast_IFastClass<D>::First() const
+    {
+        void* result;
```
kennykerr: This is an example of the call site optimization that reduces the size of the shims by around 35% by eliminating the exception frame previously required when the result variable matched the return type.

src/CMakeLists.txt
```diff
@@ -80,15 +80,18 @@ if (MSVC)
...
+        add_compile_options(-Wno-delete-non-virtual-dtor -mcx16 -fno-delayed-template-parsing -Xclang -fcoroutines-ts)
+    else()
+        add_compile_options(/await)
+    endif()
 
     # Explicitly configure _DEBUG preprocessor macro 
     string(APPEND CMAKE_CXX_FLAGS_DEBUG " /D_DEBUG")
 else()
     add_compile_options(-stdlib=libc++)
-    add_compile_options(-fcoroutines-ts)
```
devhawk: Why did you remove coro support for linux but leave it in for windows?

src/test/cpp/CMakeLists.txt
```diff
@@ -1,18 +1,59 @@
...
-target_sources(test_cpp PUBLIC main.cpp)
-add_dependencies(test_cpp test_il)
+target_sources(cpp_test PUBLIC main.cpp)
+
+file(TO_NATIVE_PATH ${CMAKE_CURRENT_BINARY_DIR}/../../tool/cpp/cpp.exe cpp_exe)
+file(TO_NATIVE_PATH ${CMAKE_CURRENT_SOURCE_DIR} project_folder)
+file(TO_NATIVE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/generated" generated_files)
+file(TO_NATIVE_PATH "${generated_files}/attributes.winmd" attributes_winmd)
+file(TO_NATIVE_PATH "${generated_files}/component.winmd" component_winmd)
+file(TO_NATIVE_PATH "${project_folder}/compare/out/winrt/Component.h" compare_component_h)
```
Scottj1s: What is strategy behind checking in all the generated compare* files?  I'm missing what they're being compare with.

kennykerr: The compare folder is what cpp.exe generates for the component so that any changes to the generated output can be compared with previous commits. That way we can easily tell when we've made a change to the code gen that affected the output and that may not necessarily trigger a unit test failure. This is exceedingly useful as I make changes to the compiler that I *think* should not impact code gen.

src/CMakeLists.txt
kennykerr: Testing with non-Windows now...

Scottj1s:
Is it a temporary strategy?  How will it scale with?

  *   Automatic updating of the version string (e.g, base.h)
  *   Additional test collateral

Maybe a better long-term strategy, as we release binaries, is to run the last version of the tool and compare outputs without checking them into source control.

src/test/cpp/CMakeLists.txt
Scottj1s: Is it a temporary strategy?  How will it scale with?

  *   Automatic updating of the version string (e.g, base.h)
  *   Additional test collateral

Maybe a better long-term strategy, as we release binaries, is to run the last version of the tool and compare outputs without checking them into source control.

src/tool/cpp/code_writers.h
```diff
@@ -2053,23 +2379,407 @@ int32_t WINRT_CALL WINRT_GetActivationFactory(void* classId, void** factory) noe
...
-    inline void write_component_g_h(writer& w, TypeDef const& type)
+    void write_component_base_type_parameter(writer&, TypeDef const&)
+    {
+
+    }
+
+    void write_component_base_type_argument(writer&, TypeDef const&)
+    {
+
+    }
```
Scottj1s: todo?

src/tool/cpp/code_writers.h
```diff
@@ -2053,23 +2379,407 @@ int32_t WINRT_CALL WINRT_GetActivationFactory(void* classId, void** factory) noe
...
+                w.write(", @", interface_name);
+            }
+        }
+    }
+
+    void write_component_module_lock(writer&, TypeDef const&)
+    {
+
+    }
+
```
Scottj1s: todo?

src/tool/cpp/main.cpp
```diff
@@ -31,7 +32,9 @@ namespace xlang
             { "include", 0 },
             { "exclude", 0 },
             { "root", 0, 1 },
-            { "base", 0, 0 }
+            { "base", 0, 0 },
+            { "prefix", 0, 0 },
+            { "overwrite", 0, 0 }
         };
```
Scottj1s: these are redundant

Scottj1s:
>         {

do we need a version override, if we're going to check in generated files?

---
Refers to: src/tool/cpp/main.cpp:20 in f33f125. [](commit_id = f33f12585b504b4de875e2bc8a1f6ac81ee0771a, deletion_comment = False)

src/tool/cpp/settings.h
```diff
@@ -14,9 +14,12 @@ namespace xlang
         bool component{};
         std::string component_folder;
         std::string component_name;
-        std::string component_pch{ "pch.h" };
+        std::string component_pch;
+        bool component_prefix{};
 
         bool verbose{};
+        bool overwrite{};
+        bool uniform{};
```
Scottj1s: component_overwrite for consistency?

src/test/cpp/CMakeLists.txt
kennykerr: * Version string should not be checked in so that shouldn't affect the output. 
* Additional tests would intentionally produce new code to compare. 

Once code gen is stable - I've reached parity with C+/WinrT - and tests provide reasonable code coverage, changes to the compare folder should be minimal/none, but provide a very convenient early warning system in the dev inner loop that something has changed. Pulling it out of the repo makes it an outer loop step that must be manually triggered and compared with an external tool. We tried that for the last rev of C++/WinRT and it was very inconvenient.

src/tool/cpp/strings/base_activation.h
```diff
@@ -21,6 +19,11 @@ WINRT_EXPORT namespace winrt
     }
 }
 
+#ifdef __clang__
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wdeprecated-declarations"
```
Scottj1s: >Wdeprecated- [](start = 35, length = 12)

what was deprecated?

src/tool/cpp/main.cpp
kennykerr: What's redundant?

src/tool/cpp/strings/base_com_ptr.h
```diff
@@ -8,6 +8,10 @@ WINRT_EXPORT namespace winrt
 
         com_ptr(std::nullptr_t = nullptr) noexcept {}
 
+        com_ptr(take_ownership_from_abi_t, void* ptr) noexcept : m_ptr(static_cast<type*>(ptr))
```
Scottj1s: >take_ownership_from_abi_t [](start = 16, length = 25)

this could come in handy with the new error object.  ;-)

kennykerr: yep 😉

src/tool/cpp/settings.h
kennykerr: good idea

src/tool/cpp/main.cpp
Scottj1s: prefix and overwrite are already listed on lines 28, 29

---
In reply to: [229060880](https://github.com/Microsoft/xlang/pull/60#discussion_r229060880) [](ancestors = 229060880)

kennykerr: Ah. I'll add a debug assertion checking for dupes.

src/tool/cpp/strings/base_activation.h
```diff
@@ -192,10 +199,6 @@ namespace winrt::impl
         template <typename F>
         auto call(F&& callback)
         {
-#ifdef WINRT_DIAGNOSTICS
-            get_diagnostics_info().add_factory<Class>();
-#endif
-
             {
```
Scottj1s: this was removed, but diagnostics_info is still around?

src/tool/cpp/strings/base_error.h
```diff
@@ -177,8 +175,6 @@ WINRT_EXPORT namespace winrt
             WINRT_VERIFY_(impl::error_ok, WINRT_GetRestrictedErrorInfo(m_info.put_void()));
         }
 
-        impl::bstr_handle m_debug_reference;
-        uint32_t const m_debug_magic{ 0xAABBCCDD };
         hresult m_code{ impl::error_fail };
```
Scottj1s: removed because of BSTR alone?  too bad

src/test/cpp/CMakeLists.txt
Scottj1s: version string will soon be injected into the tool by the nightly build, to uniquely stamp the exe.  that will cause diffs every night, yes?

---
In reply to: [229060580](https://github.com/Microsoft/xlang/pull/60#discussion_r229060580) [](ancestors = 229060580)

src/test/cpp/compare/out/winrt/Component.Fast.h
Scottj1s: sweet

src/test/cpp/CMakeLists.txt
kennykerr: Yes, but the build need not commit that version string so it should not affect the commit history.

src/tool/cpp/strings/base_activation.h
kennykerr: Clang has deprecated _ReadWriteBarrier but VC has not. Previously we disabled the warning for the entire header. Here I'm just limiting it to these two functions. Separately I'll see about replacing it with a standard alternative.

src/tool/cpp/strings/base_activation.h
kennykerr: Removed it all.

src/tool/cpp/strings/base_error.h
kennykerr: Restored this. It provides little value as it is rather obscurely supported in Visual Studio, but I'll leave it there until we have a better debugging story.

src/tool/cpp/strings/base_activation.h
DefaultRyan: Check out the PAL's atomic_reference_count. I think I've demonstrated that we can replace all that with features from the <atomic> header.

kennykerr: In general yes, but activation needs 128 bit atomics, which I don't think are available in a standard form. Some of this stuff can be made more portable. For now I'm just shooting for parity. Once I have parity with C++/WinRT I'll switch gears and look at portability again.

src/tool/cpp/code_writers.h
kennykerr: Removed placeholder code.

src/tool/cpp/code_writers.h
kennykerr: Removed placeholder code.

src/tool/cpp/strings/base_activation.h
Scottj1s: it was still in base.h - was that generated before removal?

src/test/cpp/CMakeLists.txt
Scottj1s: I don't see how that would make a difference.  The build will produce a versioned cpp.exe, which in turn will produced version headers, even with the version number being volatile (not committed).

kennykerr: That's of no consequence since the build won't be committing its changed.

src/tool/cpp/strings/base_activation.h
kennykerr: Please check again. I don't see it in the PR.

src/test/cpp/CMakeLists.txt
Scottj1s: offline conversation cleared things up - this is a visual check only (not an automated diff)

PR #61, opened 2018-10-25
Update build scripts

devhawk:
* Get rid of PowerShell build scripts 
* Add src/scripts/windows/build.cmd and src/scripts/ubuntu/build.sh scripts to automate command line builds on Windows and Ubuntu
   * Note, other Linux flavors probably work w/ Ubuntu scripts, but have only been tested on Ubuntu 18.04 
* move wslConfigureBionic.sh script to src/scripts/ubuntu folder and rename it simply configure.sh

Note, since this is changing the build scripts, it is expected to break CI builds. Will work with @Scottj1s to make sure CI system gets updated.


devhawk:
Overriding CI build requirement as this commit changes the build scripts. Tested locally, @Scottj1s to fix CI build

PR #79, opened 2018-10-31
Python/WinRT: module namespace support + initial python code module generation

devhawk:
This PR adds two major features. 

First, we now project WinRT types in a python module named for the type's namespace.  

Previously:
``` python
import _pyrt
u = _pyrt.Uri("http://microsoft.com")
```

Now:
``` python
import pyrt.windows.foundation as wf
u = wf.Uri("http://microsoft.com")
```

Second, we have started projecting a pure Python module code in addition to the native module code. Eventually, this will provide a more pythonic developer experience, but today simply imports the native types into the corresponding python module. Example generated [__init__.py](https://github.com/devhawk/pywinrt-output/blob/namespaces/generated/pyrt/windows/foundation/__init__.py) code.

``` python
from pyrt import _import_ns

_internal = _import_ns("Windows.Foundation")

Uri = _internal.Uri
WwwFormUrlDecoder = _internal.WwwFormUrlDecoder
# ad nauseum for the remaining projected types in Windows.Foundation namespace
```

Additional changes:
* added static modifier to class constructor to eliminate linker issue for types with same name but different namespace
* updated python test cmake file to support compiling + installing the test projection in addition to generating it
* updated tests to work with generated or installed projection package

Sample projected code is available here: https://github.com/devhawk/pywinrt-output/tree/namespaces


kennykerr:
`pyrt.Windows.Foundation` - this seems a bit cryptic. Obviously, this is Python-specific. Much like in C++/WinRT, the "root namespace" should be either `winrt` or `xlang` depending on what you're targeting. So in this case it would be `winrt.Windows.Foundation`.

devhawk:
pyrt is the default generated module name. It can be specified on the command line of pywinrt (see https://github.com/Microsoft/xlang/blob/devhawk/namespace/src/test/python/CMakeLists.txt#L8 and https://github.com/Microsoft/xlang/blob/devhawk/namespace/src/test/python/CMakeLists.txt#L73). 

@BenJKuhn suggested "pyrt" as the default, so I'd rather not change it right now. Besides, while I can see the value of "winrt" as the  module name for "the inbox WinRT APIs", it doesn't seem like xlang should be the module name for "arbitrary projected xlang library".

devhawk:
Closed by mistake.

zooba:
Agreed that the module name ought to be an option when generating, and maybe we want to skip the first namespace name? Then you could have "import windows.foundation…" directly (lower_case module names are more Pythonic). I actually reserved the [windows](https://pypi.org/project/windows/) package on PyPI for the day when we can ship this :)

src/test/python/complete_winml_tutorial.py
```diff
@@ -0,0 +1,149 @@
...
+@timed_op
+def load_model(model_path):
+    if type(model_path) != str:
+        model_path = str(model_path)
+    return winml.LearningModel.LoadFromFilePath(model_path)
+
+@timed_op
+async def load_image_file(file_path):
+    if type(file_path) != str:
+        file_path = str(file_path)
```
zooba: The best way to do this pattern is `os.fspath(file_path)`, which was specifically added to convert pathlib objects to str/bytes, but an unconditional `str(file_path)` is also fine.

Being able to do this implicitly would be nice too - most Python file APIs run their parameters through `fspath` (either directly or using the private C version).

src/test/python/find_projection.py
```diff
@@ -1,13 +1,5 @@
-from inspect import currentframe, getframeinfo
-from pathlib import Path
-import sys
-
-filename = getframeinfo(currentframe()).filename
```
zooba: I know it's gone now, but `__file__` contains the same value as you were getting here.

devhawk:
@zooba I'll change the casing of the modules to lower case for this PR, but I'm going to hold off on skipping the top level module name (aka pyrt in this case) until my next block of work.

src/test/python/complete_winml_tutorial.py
devhawk: I'll look at changing the sample code to use os.fspath. But I'm not sure that we be implicitly converting arbitrary python objects to strings at the WinRT layer - i.e. auto converting model_path to a str *in the projection*.

src/test/python/find_projection.py
devhawk: Now he tells me :)

zooba:
@devhawk 
> I'm going to hold off on skipping the top level module name (aka pyrt in this case)

I was thinking the top-level namespace (so "Windows" in this case), on the basis that it most naturally fits that top top-level module name will be the same, but might also need disambiguation. An option to skip it may be enough, since at least that will discourage people from building their own tricks to automatically import things.

src/test/python/complete_winml_tutorial.py
zooba: I'd consider `pathlib.*Path` objects for the same reason you'd do it for `datetime` - it's in the core library. There's a safe protocol for it though, so it probably fits neatly into the "when it's not a str, before raising see if it has `__fspath__` that returns str before raising". (Doing the same thing with `__str__` is not safe, since that's on every object.)

zooba: For a bit more background, `pathlib`'s classes were _nearly_ subclasses of `str`. Some people's preference to use `bytes` for paths, as well as the difficulty in making existing operators return the correct type, prevented it. Plus a whole lot of people just thought it was dirty (not me :) )

devhawk:
@zooba For WinRT namespace segments that have a capital letter in the middle (such as Windows.AI.MachineLearning), is a reasonable approach simply to lower case the entire path? (windows / ai / machinelearning) or do I need to add underscore or hyphen at the word break. 

I see examples of all three approaches in the python standard library:
* ensurepip, configparser
* pydoc_data, py_compile
* site-packages

devhawk:
FYI, https://github.com/devhawk/pywinrt-output/tree/namespaces has been updated w/ casing change

src/test/python/complete_winml_tutorial.py
devhawk: Done. THanks @zooba

src/test/python/complete_winml_tutorial.py
```diff
@@ -0,0 +1,142 @@
...
+
+    op.put_Completed(callback)
+
+    return await future
+
+def run_async_code(code):
+    loop = asyncio.get_event_loop()
+    loop.run_until_complete(code())
+    loop.close()
+
```
Scottj1s: this all looks like a generally useful module

src/test/python/complete_winml_tutorial.py
```diff
@@ -0,0 +1,142 @@
...
+def load_model(model_path):
+    return winml.LearningModel.LoadFromFilePath(os.fspath(model_path))
+
+@timed_op
+async def load_image_file(file_path):
+    from pyrt.windows.storage import StorageFile
+    from pyrt.windows.graphics.imaging import BitmapDecoder
+    from pyrt.windows.media import VideoFrame
+
+    file = await wrap_async_op(StorageFile.GetFileFromPathAsync(os.fspath(file_path)))
```
Scottj1s: >wrap_async_op [](start = 17, length = 13)

is it possible to project each *Async winrt method as an "async def" so it can be awaited directly?  be nice to remove this boilerplate wrapping

src/tool/python/settings.h
```diff
@@ -6,7 +6,7 @@ namespace xlang
     {
         std::set<std::string> input;
 
-        std::string output_folder;
+        std::experimental::filesystem::path output_folder;
```
Scottj1s: >std::experimental::filesystem [](start = 8, length = 29)

how come you're not using <filesystem> ?

Scottj1s: <filesystem> (wow, CodeFlow just snipped that out)

src/test/python/complete_winml_tutorial.py
devhawk: Yes, that's the plan - part of my next task. More details in #37

src/test/python/complete_winml_tutorial.py
devhawk: run_async_code is a Python 3.6 version of the [asyncio.run method](https://docs.python.org/3/library/asyncio-task.html#asyncio.run) from Python 3.7. Example Code: https://docs.python.org/3/library/asyncio-task.html#coroutines

timed_op is useful. I wrote the original version *years ago* when I was working on IronPython (though I did have to add async support for this code). Nothing to do w/ WinRT/Xlang tho.

wrap_async_op addressed below

src/tool/python/settings.h
devhawk: I think there was a clang/ubuntu issue. Once we get Ubuntu build up and running, we should change this. Filed #85 to track

devhawk:
FYI, I used my admin privileges to merge this because the CLA bot seemed to have never run after merging in latest changes from master.

PR #80, opened 2018-10-31
Removing hardcoded arch in build script

phoyd:
Let build.sh read the target arch from clang -v.


phoyd:
'awk' is listed as "tools you can expect to find on any machine" for autoconf (https://www.gnu.org/software/autoconf/manual/autoconf-2.65/html_node/Limitations-of-Usual-Tools.html) and part of the POSIX (http://pubs.opengroup.org/onlinepubs/9699919799/utilities/awk.html) specification. I've never seen a Linux/BSD/OSX install without awk - it's one of the bread and butter utilities for system scripts. Even 'busybox' carries an implementation. 'awk' is safe assumption. 

OTOH, I don't think that we need a build script at all, because everything should be in the CMakeLists.txt. If the CMake build works, then I can open the project with full support in VSCode or QtCreator. Build instructions and dependencies should be in the documentation.

devhawk:
OK, no need to add awk to configure script. I think we'll hold off on folder renaming for now. When we get to the point that we are validating on multiple Linux distros, we can revisit.

Issue #23, opened 2018-10-10
Azure Pipelines builds aren't public


akoeplinger:
The badge in the readme links to https://microsoft.visualstudio.com/Dart/_build/latest?definitionId=31784 which is only accessible inside of MSFT.

udnaan:
Ditto. In order for the badge to inspire confidence, others should be able to inspect the build.

devhawk:
Working on Azure pipeline VM definitions for  xlang. For now, they are in my fork of https://github.com/Microsoft/azure-pipelines-image-generation but we will add them to the xlang repo after we get them up and running. 

Anyone interested in taking an early look, here are the [Packer files](https://www.packer.io/)

* [Linux xlang build](https://github.com/devhawk/azure-pipelines-image-generation/blob/xlang/images/linux/xlang-build.json)
* [Windows xlang build](https://github.com/devhawk/azure-pipelines-image-generation/blob/xlang/images/win/xlang-build.json)

devhawk:
Linux CI builds are enabled and public: https://dev.azure.com/msft-xlang/public/_build/latest?definitionId=1

Public Windows CI builds coming soon, so leaving this issue open

devhawk:
I will close this when #92 gets merged

PR #86, opened 2018-11-05
Component optimizations

kennykerr:
This update adds support for a number of additional opt-in optimizations for C++/WinRT (that will be on by default for xlang). 

* Uniform construction
* Direct implementation access
* Type erased factories

The first two allow a component direct access to implementation types while using only the projected types. No need to use make/make_self/get_self if you simply want to use the public API surface. Your calls  will compile down to direct calls into the implementation that may even be entirely inlined.

The last one avoids the #include dependencies in module.g.cpp so that it need not be recompiled every time any one implementation class happens to change and thus improved build performance.

These optimizations are turned on using the /opt flag as they are breaking changes that the component developer may need to make minor changes to support. New projects (templates) should use /opt by default and xlang will have these on by default.


src/test/cpp/CMakeLists.txt
```diff
@@ -2,7 +2,28 @@ cmake_minimum_required(VERSION 3.9)
...
-target_sources(cpp_test PUBLIC pch.cpp activation.cpp test/Async.cpp test/Collections.cpp test/Edge.cpp test/Fast.cpp test/Enum.cpp test/Structs.cpp "${CMAKE_CURRENT_SOURCE_DIR}/generated/module.g.cpp")
+target_sources(cpp_test PUBLIC 
+    "${CMAKE_CURRENT_SOURCE_DIR}/generated/module.g.cpp"
+    pch.cpp
+    activation.cpp
+    test/Async.cpp
+    test/Collections.cpp
+    test/Edge.cpp
+    test/Fast.cpp
+    test/Enum.cpp
```
devhawk: We should alphabetize lists like this so we can more easily pick out changes in the future (though module.g.cpp and pch.cpp should probably be at the top, regardless)

src/test/cpp/compare/out/module.g.cpp
```diff
@@ -11,6 +12,77 @@
...
+    if (winrt::get_module_lock())
+    {
+        return false;
+    }
+
+    winrt::clear_factory_cache();
+    return true;
+}
+
+void* WINRT_CALL winrt_get_activation_factory(std::wstring_view const& name)
```
devhawk: what happens if this function throws?

src/tool/cpp/settings.h
```diff
@@ -17,12 +17,15 @@ namespace xlang
...
+        std::string component_lib;
+        bool component_opt{};
 
         bool verbose{};
-        bool uniform{};
 
         std::set<std::string> include;
         std::set<std::string> exclude;
+
+        meta::reader::filter filter;
```
devhawk: :thumbs_up: I'm going to do this in pywinrt

devhawk:
FYI, I queued a build for this PR. Progress is being made on #23 but manual build kick off is required in the short term :(

src/test/cpp/implementation/Component.Fast.FastClass.h
```diff
@@ -51,3 +51,4 @@ namespace winrt::Component::Fast::factory_implementation
     {
     };
 }
+
```
Scottj1s: curious - hand edited or generated?

Scottj1s: curious - hand edited or generated?

kennykerr: Hand rolled before I had any code gen working.

JohnTasler:
The Direct implementation access concerns me a bit. It seems it could easily allow a developer to forget to add their methods to the IDL, since they're accessible without explicitly using make_self/get_self.

Issue #50, opened 2018-10-24
stop disabling Wreturn-type-c-linkage in pal.h


devhawk:
operator|= and operator&=give a return-type-c-linkage warning when compiling with clang on Ubuntu. these operators are already in a #ifdef __cplusplus block, so I think the warning is spurious. If there a way to modify these functions to avoid the clang warning? I'm OK leaving the #pragma that disables the warning in place (though the TODO comment needs to be updated)

Issue #49, opened 2018-10-24
stop disabling -Wexceptions for xlang::to_result


devhawk:
xlang::to_result is marked no_except, but it throws and doesn't catch unconditionally. Can we rework this function to eliminate the #pragma to ignore the compiler warning on clang?

PR #90, opened 2018-11-08
Fix some clang warnings

DefaultRyan:


DefaultRyan:
Fixes #50 and #49

src/platform/published/pal.h
```diff
@@ -241,6 +207,31 @@ extern "C"
 #endif
 
 #ifdef __cplusplus
+
+inline constexpr xlang_string_encoding operator|(xlang_string_encoding lhs, xlang_string_encoding rhs) noexcept
```
kennykerr: `constexpr` implies `inline`.

PR #91, opened 2018-11-08
Strip trailing whitespace from python command-line output.

DefaultRyan:
Was breaking VS IDE.


src/test/python/CMakeLists.txt
```diff
@@ -3,6 +3,7 @@ cmake_minimum_required(VERSION 3.9)
 execute_process(
     COMMAND py -c "import sys; vi = sys.version_info; print('cp{0}{1}-{2}'.format(vi.major, vi.minor, 'win_amd64' if sys.maxsize > 2**32 else 'win32'))"
     OUTPUT_VARIABLE python_version_info
+    OUTPUT_STRIP_TRAILING_WHITESPACE
```
devhawk: Nice, didn't know about this cmake feature. As part of my devhawk/python branch, I'm changing this cmake script so that I no longer need this python code, but thanks for getting this fixed while I work on build stuff

PR #94, opened 2018-11-12
Add ABI executable

dunhor:
The goal is to be as close to MIDL (on IDL files that were generated with WINMDIDL) as possible. Ideally they would be identical, but that's fairly difficult to do. In general, the known differences are:

* The order of declarations/definitions is similar, but still quite different. The primary difference is with generics, which MIDL/WINMDIDL sort very weirdly. I try and mimic as much as I can to make validation easier, but there's a point where it's not worth it and would require a fair amount of unnecessary work to do
* MIDL seems to generate generics in places where it's not strictly necessary. E.g. if one argument is in the target namespace, then the generic seems to be generated for that header, even if it's not used. This is not a behavior that is mimicked. This seems to occasionally cause build breaks when integrating with existing code
* MIDL recursively defines contract values. E.g. if A includes B includes C, then A.h will have contract version definitions for contracts defined in C. This only goes a single namespace "deep" but includes its dependencies, so those contracts will still get defined, and since those contracts shouldn't be used in the active header, there should be no noticeable difference

On my main OS SSD, it takes ~636 ms to process the entire Windows SDK using the Clang built binary and ~613 ms with the MSVC built binary.


src/tool/abi/abi_writer.h
```diff
@@ -0,0 +1,251 @@
...
+    {
+        return m_config;
+    }
+
+    void write_code(std::string_view value)
+    {
+        xlang::text::bind_list("::", namespace_range{ value })(*this);
+    }
+
+    static void write_uppercase(writer& w, std::string_view str)
```
DefaultRyan: write_uppercase and write_lowercase run the risk of misbehaving with non-ASCII characters. @kennykerr we should add some types to our test component with names like Çömpôñénť.

dunhor: That's a good point, though MIDL is currently a very unhappy camper when I try and use non-ANSI characters

dunhor: Changed characters and got it to compile and output a .winmd and .h (though, with an assertion failure), but the .h is bogus...

kennykerr:
Thanks! Will dig into this PR soon.

src/tool/abi/main.cpp
```diff
@@ -0,0 +1,313 @@
...
+{
+    auto out{ absolute(args.value("output", "abi")) };
+    create_directories(out);
+    out += path::preferred_separator;
+    return out.string();
+}
+
+void print_usage()
+{
+    puts("Usage...");
```
Scottj1s: need to implement

src/tool/abi/metadata_cache.cpp
```diff
@@ -0,0 +1,680 @@
...
+
+        auto getVersion = [&](interface_type const* iface)
+        {
+            if (auto ver = version_attribute(iface->type()))
+            {
+                return *ver;
+            }
+
+            xlang::throw_invalid("Interface '", iface->clr_full_name(), "' has no version attribute, but is trying to "
+                "be used as a fast ABI iterface for class '", type.clr_full_name(), "' which has no contract attribute");
```
Scottj1s: typo - iterface

src/tool/abi/sha1.h
```diff
@@ -0,0 +1,177 @@
...
+    std::size_t shift = (sizeof(T) - 1) * 8;
+    T mask = static_cast<T>(0xFF) << shift;
+    for (std::size_t i = 0; i < sizeof(T); ++i)
+    {
+        *(target++) = static_cast<std::uint8_t>((value & mask) >> shift);
+        mask >>= 8;
+        shift -= 8;
+    }
+
+    return target;
```
Scottj1s: why returned?

Scottj1s: I see - updated in loop

src/tool/abi/main.cpp
dunhor: This came up on my aborted PR. See: https://github.com/Microsoft/xlang/pull/51

src/tool/abi/sha1.h
dunhor: 👍🏻

src/tool/abi/types.h
```diff
@@ -0,0 +1,952 @@
...
+        return result;
+    }
+
+    [[noreturn]]
+    virtual void write_cpp_forward_declaration(writer&) const override
+    {
+        // Fast ABI types should only ever be used as base classes, which need to be defined, so we should never find
+        // ourselves in a scenario where we are trying to forward declare one
+        XLANG_ASSERT(false);
+        xlang::throw_invalid("Attemptint to forward declare a fast ABI type");
```
Scottj1s: typo

PR #96, opened 2018-11-13
Defaultryan/pal abi refresh

DefaultRyan:
This refactors the PAL to be more compatible with the upcoming error handling strategy. The actual error type/interface still needs to be put in, but at this point, it should be a minor change. More importantly, this should largely unblock ongoing work on updating the xlang-specific C++ projection tool.


src/platform/activation_abi.cpp
```diff
@@ -17,12 +18,12 @@ namespace xlang::impl
...
                 void* factory{};
-                xlang_result result = (*pfn)(class_name, iid, &factory);
-                if (result == xlang_error_ok)
+                xlang_error_info* result = (*pfn)(class_name, iid, &factory);
+                if (result == nullptr)
                 {
                     return factory;
                 }
-                else if (result != xlang_error_class_not_available)
+                else if (result->error_code() != xlang_error_class_not_available)
```
kennykerr: I'm a little concerned about the use of exceptions in the PAL. The PAL functions are meant to be very low-level routines that language projections will use to implement high-level abstractions. Having those lower-level implementations throw exceptions can cause unintended performance hickups in otherwise non-failure code paths. I recently mentioned this to Scott in the context of `QueryInterface`. You really don't want a `QueryInterface` implementation throwing internally. Likewise, the activation functions should probably not throw in the event that failure is not considered fatal at the language level. I can imagine a language projection where multiple activation paths might be considered. This happens today with WinRT. It's possible that such a scenario may not be common in xlang but it's worth considering. Still, in this example it seems that the implementation could just as easily use a nullptr return value to signal failure to find the class. That way, any "TryCreate" function in the language or app can avoid an exception under the hood.

src/platform/pal_error.h
```diff
@@ -0,0 +1,91 @@
...
+    {
+        struct error_info : xlang_error_info
+        {
+            explicit error_info(xlang_result result) noexcept
+                : m_result(result)
+            {}
+
+            int32_t XLANG_CALL QueryInterface(xlang_guid const& id, void** object) noexcept override
+            {
+                if (id == xlang_unknown_guid)
```
kennykerr: The `if (id == xlang_unknown_guid)` and `if (id == xlang_error_info_guid)` branches could be combined. Not a big deal either way though.

src/platform/pal_error.h
```diff
@@ -0,0 +1,91 @@
...
+                    *object = static_cast<xlang_unknown*>(this);
+                }
+                else if (id == xlang_error_info_guid)
+                {
+                    *object = static_cast<xlang_error_info*>(this);
+                }
+                else
+                {
+                    *object = nullptr;
+                    return 0x80004002;
```
kennykerr: Either an implementation constant or a comment for clarity.

src/platform/pal_error.h
```diff
@@ -0,0 +1,91 @@
...
+{
+    namespace impl
+    {
+        struct error_info : xlang_error_info
+        {
+            explicit error_info(xlang_result result) noexcept
+                : m_result(result)
+            {}
+
+            int32_t XLANG_CALL QueryInterface(xlang_guid const& id, void** object) noexcept override
```
kennykerr: These should all probably be `final`.

DefaultRyan: Will do, but these are merely placeholder implementations for now. :)

kennykerr: Ah right.

src/platform/activation_abi.cpp
DefaultRyan: Returning nullptr to indicate a failure to find the class still needs to be distinguished from other, more "exceptional" cases, such as the component failing to allocate memory for the factory, or an incorrect GUID being supplied (causing QueryInterface to fail).

As I read your comment, I'm wondering if you're advocating more of a "try_get_activation_factory" function in the PAL?

kennykerr: Chatted offline - we're on the same page.

PR #103, opened 2018-11-30
C++/WinRT is now fully generic

kennykerr:
Thanks to the xlang metadata reader, C++/WinRT now generates all parameterized, or generic, interfaces from metadata. Interfaces like `Windows.Foundation.Collections.IVector<T>` are now generated from metadata rather than hardcoded in `base.h` - I originally hand-rolled all of the generic interfaces because I did not have a sufficiently intelligent parser that could give me the necessary generic type information. This meant that `base.h` had a lot of hand-rolled code that was very hard to maintain. That has now been replaced by generated code, cutting the size of `base.h` in half. It also highlighted some areas where the hand-rolled code had not kept up with the optimizations in generated code. 

This now also paves the way for different C++ language projections, such as the xlang and noexcept variants, because the generated code will handle much of the differences. It also means that we can support open generics if so desired. 

Beyond that, this update includes several bug fixes.


kennykerr:
@devhawk - Harry, can you get VS 15.9 on the build machine? I believe that’s why this is failing – hitting some parser bugs in 15.8.

devhawk:
> @devhawk - Harry, can you get VS 15.9 on the build machine? I believe that’s why this is failing – hitting some parser bugs in 15.8.

Build containers have VS Build Tools, not full VS, but yes I can refresh the build image. Will also setup a task to re-build the build container every week.

src/library/impl/meta_reader/cache.h
```diff
@@ -135,8 +135,6 @@ namespace xlang::meta::reader
 
         void remove_legacy_cppwinrt_foundation_types()
```
Scottj1s: maybe rename function per it's now narrower purpose - something like remove_projected_numeric_types

src/test/cpp/compare/out/winrt/Component.Async.h
```diff
@@ -9,25 +9,25 @@ namespace winrt::impl
     {
         void* operation;
         check_hresult(WINRT_SHIM(Component::Async::IClassStatics)->Action(&operation));
-        return { take_ownership_from_abi, operation };
+        return { operation, take_ownership_from_abi };
```
Scottj1s: I like the trailing tag better

src/test/cpp/component.idl
```diff
@@ -192,4 +192,36 @@ namespace Component
...
+            void SyncCall(
+                String a,
+                Windows.Foundation.IReference<int> b,
+                Windows.Foundation.Collections.IIterable<int> c,
+                Windows.Foundation.Collections.IVectorView<int> d,
+                Windows.Foundation.Collections.IMapView<int, int> e,
+                Windows.Foundation.Collections.IVector<int> f,
+                Windows.Foundation.Collections.IMap<int, int> g);
+
+            Windows.Foundation.IAsyncAction NotSyncCall(
```
Scottj1s: curious name, why not AsyncCall?

src/test/cpp/pch.h
```diff
@@ -2,10 +2,11 @@
...
 #include "winrt/base.h"
+using namespace std::literals;
 
 template<bool B> bool static_require()
 {
     return B;
 }
 
-#define STATIC_REQUIRE(...) REQUIRE(static_require<__VA_ARGS__>())
+#define STATIC_REQUIRE(...) { constexpr bool require = __VA_ARGS__; REQUIRE(static_require<require>()); }
```
Scottj1s: some clang limitiation?

src/library/impl/meta_reader/cache.h
kennykerr: 👍

src/test/cpp/compare/out/winrt/Component.Async.h
kennykerr: Yep. I also realized that `hresult_error` already has an attach semantic and its trailing, so this is more consistent.

src/test/cpp/component.idl
kennykerr: Because the old compiler looked for the substr "Async" and I wanted to make sure that would fail the new test as we now explicitly look for one of the the IAsyncXxx return type. I can probably clean up the naming now that this old code is gone. 👍

src/test/cpp/pch.h
kennykerr: No, the Catch REQUIRE macro falls over with complex expressions. This prevents that from happening.

src/tool/cpp/cppx/strings/base_abi.h
```diff
@@ -1,386 +1 @@
```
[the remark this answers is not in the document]
kennykerr: No, I needed to leave one to keep the cmake "Strings" machinery from falling over...

PR #106, opened 2018-11-30
update windows build container

devhawk:
Removing seperate install of Windows 10 SDK. As of 15.9.1, VS Build Tools includes Windows 10 SDK v17763 (aka RS5 SDK). [Verified manually](https://dev.azure.com/msft-xlang/public/_build/results?buildId=135) that this new image works with our existing build infrastrucutre 

~~Note, the WIndows SDK is a required depenency of the cmake tooling (aka Microsoft.VisualStudio.Component.VC.CMake.Project). I think we should be specifying the Windows SDK on the build tools command line, but the docs are unclear on the topic. Waiting on feedback from VS team before committing.~~

VS team verified Windows 10 SDK v17763 component ID is same between VS and VS Build Tools, though it's not listed on the [VS Build tool component ID list page](https://docs.microsoft.com/en-us/visualstudio/install/workload-component-id-vs-build-tools?view=vs-2017) (currently). Manually added to dockerfile to ensure correct installation in case vs build tool dependency tree changes in the future

### DIFF
--- a/src/scripts/windows/Dockerfile
+++ b/src/scripts/windows/Dockerfile
@@ -5,7 +5,6 @@
 FROM microsoft/dotnet-framework:4.7.2-runtime-windowsservercore-1803
 
 ADD https://aka.ms/vs/15/release/vs_buildtools.exe C:\TEMP\vs_buildtools.exe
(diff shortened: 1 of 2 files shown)


devhawk:
FYI, this change will get picked up as part of the new Sunday night build container refresh

PR #107, opened 2018-12-10
C++/WinRT: Improved support for COM-style single interface inheritance

kennykerr:
As much as we like WinRT, C++/WinRT is also used for a lot of COM-only APIs. This update makes it possible to implement COM servers where their exists an interface hierarchy. This is not required for WinRT, but is required for some COM implementations and is also needed for the upcoming WinRT fast ABI.

### DIFF
--- a/src/library/impl/meta_reader/cache.h
+++ b/src/library/impl/meta_reader/cache.h
@@ -133,7 +133,7 @@ namespace xlang::meta::reader
             return m_namespaces;
         }
 
-        void remove_legacy_cppwinrt_foundation_types()
+        void remove_cppwinrt_foundation_types()
         {
             auto remove = [&](auto&& ns, auto&& name)
             {
--- a/src/test/cpp/compare/out/winrt/base.h
+++ b/src/test/cpp/compare/out/winrt/base.h
@@ -710,6 +710,12 @@ WINRT_EXPORT namespace winrt
         return impl::guid_storage<default_interface<T>>::value;
     }
 
+    template <typename T>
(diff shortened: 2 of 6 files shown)


src/test/cpp/compare/out/winrt/base.h
```diff
@@ -1768,7 +1768,6 @@ WINRT_EXPORT namespace winrt
 namespace winrt::impl
 {
     inline constexpr hresult error_ok{ 0 }; // S_OK
-    inline constexpr hresult error_false{ 1 }; // S_FALSE
```
kennykerr: Removed unused internal constant.

src/library/impl/meta_reader/cache.h
```diff
@@ -133,7 +133,7 @@ namespace xlang::meta::reader
             return m_namespaces;
         }
 
-        void remove_legacy_cppwinrt_foundation_types()
+        void remove_cppwinrt_foundation_types()
```
kennykerr: Renamed as per preview PR feedback.

src/test/cpp/compare/out/winrt/base.h
```diff
@@ -710,6 +710,12 @@ WINRT_EXPORT namespace winrt
         return impl::guid_storage<default_interface<T>>::value;
     }
 
+    template <typename T>
+    bool is_guid_of(guid const& id) noexcept
```
kennykerr: Simply hoisted this out of the `impl` namespace.

PR #109, opened 2018-12-14
Improve C++/WinRT support for internal OS use of WIL 

kennykerr:
Currently, OS developers using both WIL and C++/WinRT must remember to include <wil/cppwinrt.h> to ensure that all static libs can interoperate properly when errors propagate through both WIL and C++/WinRT code paths.

Michael Novak and I have agreed on a new approach for bridging WIL and C++/WinRT using a global function pointer. In his words, “It’s the most reliable way I’ve found to deal with ‘optional’ functionality given the various integration problems and ordering that stem from linking disparately compiled libs together.  WIL does this in multiple places today to handle everything from exceptions-disabled to crt differences to store compatibility, etc.”

Once wil/cppwinrt.h has been updated (in the OS repo), I will remove the WINRT_EXTERNAL_CATCH_CLAUSE macro from C++/WinRT.

### DIFF
--- a/src/test/cpp/test/CustomError.cpp
+++ b/src/test/cpp/test/CustomError.cpp
@@ -7,7 +7,7 @@ using namespace Windows::Foundation;
 namespace
 {
     // Some custom exception type unknown to C++/WinRT
-    struct WilException
+    struct CustomError
     {
     };
 
@@ -16,7 +16,7 @@ namespace
         hstring ToString()
         {
             // Throw custom exception inside C++/WinRT projection
-            throw WilException();
+            throw CustomError();
         }
     };
 
@@ -29,7 +29,7 @@ namespace
         {
             throw;
         }
-        catch (WilException)
+        catch (CustomError)
         {
             return 0x80000018; // E_ILLEGAL_DELEGATE_ASSIGNMENT
         }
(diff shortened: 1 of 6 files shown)


src/test/cpp/test/Wil.cpp
```diff
@@ -0,0 +1,53 @@
+#include "pch.h"
+#include "winrt/Windows.Foundation.h"
+
+using namespace winrt;
+using namespace Windows::Foundation;
+
+namespace
+{
+    // Some custom exception type unknown to C++/WinRT
+    struct WilException
```
Scottj1s: I know WIL is the prime beneficiary, but a more general name would be clearer (e.g., CustomException)

kennykerr: Sure. 😉

kennykerr:
test

kennykerr:
test2

src/test/cpp/test/CustomError.cpp
```diff
@@ -0,0 +1,53 @@
...
+        {
+            // Throw custom exception inside C++/WinRT projection
+            throw CustomError();
+        }
+    };
+
+    // Global handler to translate custom exception
+    int32_t WINRT_CALL handler(void* address) noexcept
+    {
+        REQUIRE(address);
```
kennykerr: test

Issue #113, opened 2019-01-02
winrt::capture helper must support projected types as well


kennykerr:
The capture helpers currently only support com_ptr types and not projected types. This impacts a number of interop scenarios where COM interfaces are used to create WinRT types. Those callers must manually stitch up the calls, which is rather error prone.

Issue #112, opened 2019-01-02
winrt::delegate is missing constructors for get_strong and get_weak support


kennykerr:
This particular delegate didn't receive the `get_strong/get_weak` overloads that all of the other delegates received in order to create a delegate holding a strong/weak reference automatically. This is an inconsistency and makes it harder to code these patterns correctly.

PR #114, opened 2019-01-02
C++/WinRT improved capture helpers and variadic delegates

kennykerr:
This update fixes the limitation with the `winrt::capture` helpers by supporting projected types as well. The comes up now and then with the WinRT interop APIs when they return a projected type.

This update also adds support for `get_strong/get_weak` when creating variadic a `winrt::delegate`.

I have also ported the original capture tests from cppwinrt and fixed some minor issues building xlang with VS 2019.

Issues: #113, #112

### DIFF
--- a/src/library/impl/base.h
+++ b/src/library/impl/base.h
@@ -24,6 +24,7 @@
 #include <list>
 #include <map>
 #include <optional>
+#include <string>
 #include <string_view>
 #include <variant>
 #include <vector>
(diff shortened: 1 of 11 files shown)


src/library/impl/base.h
```diff
@@ -24,6 +24,7 @@
 #include <list>
 #include <map>
 #include <optional>
+#include <string>
```
kennykerr: The xlang libraries use `std::to_string`, which is included with `<string>`.

PR #116, opened 2019-01-08
C++/WinRT coroutine locals may return/complete before locals are destroyed

kennykerr:
@JamesMcNellis recently discovered that the traditional way of implementing a coroutine type may allow locals within the coroutine to be destroyed after the coroutine returns/completes (rather than prior to final suspension). @GorNishanov clarified that the resumption of any waiter should be deferred until final suspension to avoid this problem (among other benefits).

This update fixes this problem for C++/WinRT coroutines and also ports over all of the existing coroutine tests from out internal repo. New tests have been added to detect this issue as well. 

I wish I could port over VSO's vastly superior PR viewer as well. 🙁

### DIFF
--- a/src/test/cpp/test/AsyncLocal.cpp
+++ b/src/test/cpp/test/AsyncLocal.cpp
@@ -1,0 +1,81 @@
+#include "catch.hpp"
+#include <windows.h>
+#include "winrt/Windows.Foundation.h"
+
+using namespace winrt;
+using namespace Windows::Foundation;
+
+namespace
+{
+    //
+    // Checks that coroutine locals are destroyed prior to notifying waiters.
+    //
+
+    struct Local
+    {
+        bool& destroyed;
+
+        ~Local()
+        {
+            Sleep(100);
+            REQUIRE(!destroyed);
+            destroyed = true;
+        }
+    };
+
+    IAsyncAction Action(HANDLE event, bool& destroyed)
+    {
+        co_await resume_on_signal(event);
+        Local local{ destroyed };
+        Sleep(100);
+    }
+
+    IAsyncActionWithProgress<int> ActionWithProgress(HANDLE event, bool& destroyed)
+    {
+        co_await resume_on_signal(event);
+        Local local{ destroyed };
+        Sleep(100);
+    }
+
+    IAsyncOperation<int> Operation(HANDLE event, bool& destroyed)
+    {
+        co_await resume_on_signal(event);
+        Local local{ destroyed };
+        Sleep(100);
+        co_return 1;
+    }
+
+    IAsyncOperationWithProgress<int, int> OperationWithProgress(HANDLE event, bool& destroyed)
+    {
+        co_await resume_on_signal(event);
+        Local local{ destroyed };
+        Sleep(100);
+        co_return 1;
+    }
+
+    template <typename F>
+    void Check(F make)
+    {
+        handle start{ CreateEvent(nullptr, true, false, nullptr) };
+        handle completed{ CreateEvent(nullptr, true, false, nullptr) };
+        bool destroyed = false;
+        auto async = make(start.get(), destroyed);
+
+        async.Completed([&](auto&&...)
+            {
+                REQUIRE(destroyed);
+                SetEvent(completed.get());
+            });
+
+        SetEvent(start.get());
+        REQUIRE(WaitForSingleObject(completed.get(), 1000) == WAIT_OBJECT_0);
+    }
+}
+
+TEST_CASE("AsyncLocal")
+{
+    Check(Action);
+    Check(ActionWithProgress);
+    Check(Operation);
+    Check(OperationWithProgress);
+}
--- a/src/tool/cpp/cppwinrt/strings/base_await.h
+++ b/src/tool/cpp/cppwinrt/strings/base_await.h
@@ -671,9 +671,27 @@ namespace winrt::impl
         {
         }
 
+        auto GetResults()
+        {
+            slim_lock_guard const guard(m_lock);
+
+            if (m_status == AsyncStatus::Completed)
+            {
+                return static_cast<Derived*>(this)->get_return_value();
+            }
+
+            rethrow_if_failed();
+            WINRT_ASSERT(m_status == AsyncStatus::Started);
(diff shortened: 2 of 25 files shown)


src/test/cpp/test/AsyncLocal.cpp
```diff
@@ -0,0 +1,81 @@
...
+            REQUIRE(!destroyed);
+            destroyed = true;
+        }
+    };
+
+    IAsyncAction Action(HANDLE event, bool& destroyed)
+    {
+        co_await resume_on_signal(event);
+        Local local{ destroyed };
+        Sleep(100);
```
GorNishanov: Can you clarify why this Sleep is important?

kennykerr: I'll remove those. They aren't needed. My original repro needed a sleep due to some race that prevented the repro without the sleep, but I see the simpler repro works without the sleep now.

src/tool/cpp/cppwinrt/strings/base_await.h
```diff
@@ -671,77 +672,100 @@ namespace winrt::impl
...
+                void await_resume() const noexcept
                 {
-                    m_status = AsyncStatus::Canceled;
                 }
-                catch (...)
+
+                bool await_suspend(std::experimental::coroutine_handle<>) const noexcept
                 {
-                    m_status = AsyncStatus::Error;
+                    uint32_t const remaining = promise->subtract_reference();
```
GorNishanov: Should `subtract_reference` be after `promise->set_completed()`?
As soon as I gave up my reference, I assume that the promise may be destroyed by someone else concurrently which decremented its own reference and noticed that the count went to zero.

src/tool/cpp/cppwinrt/strings/base_std_async_action.h
```diff
@@ -4,51 +4,12 @@ WINRT_EXPORT namespace std::experimental
...
-                    return;
-                }
-
-                this->rethrow_if_failed();
-                WINRT_ASSERT(this->m_status == AsyncStatus::Started);
-                throw winrt::hresult_illegal_method_call();
-            }
-
-            void return_void()
+            void return_void() const noexcept
```
GorNishanov: Nice! So much shorter

src/tool/cpp/cppwinrt/strings/base_std_async_operation.h
```diff
@@ -4,52 +4,18 @@ WINRT_EXPORT namespace std::experimental
...
-
-                if (this->m_status == AsyncStatus::Completed)
-                {
-                    return m_result;
-                }
-
-                this->rethrow_if_failed();
-                WINRT_ASSERT(this->m_status == AsyncStatus::Started);
-                throw winrt::hresult_illegal_method_call();
+                return m_result;
```
GorNishanov: Making this `return std::move(m_result);` might help to avoid extra copy.

src/tool/cpp/cppwinrt/strings/base_std_async_operation_with_progress.h
```diff
@@ -4,65 +4,32 @@ WINRT_EXPORT namespace std::experimental
...
-
-                if (this->m_status == AsyncStatus::Completed)
-                {
-                    return m_result;
-                }
-
-                this->rethrow_if_failed();
-                WINRT_ASSERT(this->m_status == AsyncStatus::Started);
-                throw winrt::hresult_illegal_method_call();
+                return m_result;
```
GorNishanov: `return std::move(m_result);` ?

src/tool/cpp/cppwinrt/strings/base_std_async_operation_with_progress.h
```diff
@@ -4,65 +4,32 @@ WINRT_EXPORT namespace std::experimental
...
             }
 
-            void return_value(TResult const& result)
+            void return_value(TResult const& result) noexcept
```
GorNishanov: `noexcept` assumes that `m_result = result` never throws. Is this valid assumption?

src/tool/cpp/cppwinrt/strings/base_std_async_operation.h
```diff
@@ -4,52 +4,18 @@ WINRT_EXPORT namespace std::experimental
...
             }
 
-            void return_value(TResult const& result)
+            void return_value(TResult const& result) noexcept
```
GorNishanov: `noexcept` here assumes that `m_result = result;` can never throw. Is this a valid assumption?

src/tool/cpp/cppwinrt/strings/base_std_async_action.h
kennykerr: 👍

src/tool/cpp/cppwinrt/strings/base_std_async_operation.h
kennykerr: Good idea. Will do.

src/tool/cpp/cppwinrt/strings/base_std_async_operation.h
kennykerr: Yes, since `result` is guaranteed to be a WinRT type and WinRT types in C++/WinRT all provide noexcept copy semantics.

src/tool/cpp/cppwinrt/strings/base_std_async_operation_with_progress.h
kennykerr: Yes, will do.

src/tool/cpp/cppwinrt/strings/base_std_async_operation_with_progress.h
kennykerr: Same as above. Since result is guaranteed to be a WinRT type and WinRT types in C++/WinRT all provide noexcept copy semantics.

src/tool/cpp/cppwinrt/strings/base_std_async_operation.h
kennykerr: I'm also going to change `return_value` to use `TResult&&` and `std::forward` to avoid a copy on the way in.

PR #118, opened 2019-01-10
C++/WinRT events don't gracefully handle invalid tokens

kennykerr:
The `winrt::event<Delegate>` implementation does not gracefully handle the case where its `remove` method is called with an invalid token value (a value not present in the array). This update brings the implementation in line with WRL, which handles this scenario.

Tests have been added to validate this scenario.

### DIFF
--- a/src/test/cpp/CMakeLists.txt
+++ b/src/test/cpp/CMakeLists.txt
@@ -16,6 +16,7 @@ target_sources(cpp_test PUBLIC
     test/Edge.cpp
     test/Enum.cpp
     test/Events.cpp
+    test/EventsInvalid.cpp
     test/Fast.cpp
     test/FinalRelease.cpp
     test/Names.cpp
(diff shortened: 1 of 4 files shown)


JohnTasler:
>         slim_mutex m_change;

Why is this needed? Isn't a swap within the scope of a "change"?

---
Refers to: src/tool/cpp/cppwinrt/strings/base_events.h:502 in 36d83b4. [](commit_id = 36d83b4539190a41bc197b108d45eeb29db9c366, deletion_comment = False)

kennykerr:
@JohnTasler GitHub’s PR viewer is horrible. Unfortunately, if you leave a comment on a line of code outside of the PR then I cannot reply to it.

Briefly, the `m_change` lock ensures that add/remove are synchronized while `m_swap` only protects swapping the targets array. This improves concurrency because the latter happens more frequently and we don’t want it to block on the former.

JohnTasler:
Sorry. I wanted to grok and review the whole class. I'm using CodeFlow.

---
In reply to: [453295948](https://github.com/Microsoft/xlang/pull/118#issuecomment-453295948) [](ancestors = 453295948)

PR #119, opened 2019-01-10
XDN05 - xlang Binary Interface

devhawk:
Start fleshing out XDN05, which covers the application binary interface exposed by xlang objects


design_notes/XDN05 - xlang Binary Interface.md
```diff
@@ -12,6 +12,233 @@ status: draft
...
+As stated above, interfaces are identified by a GUID. This GUID is passed by reference as the first
+parameter of QueryInterface. The pointer to the requested interface - if available - is returned
+via a void** out parameter. It is the responsibility of the calling code to cast this pointer to
+the correct COM interface pointer type.
+
+Methods on COM interfaces (including xlang ABI interfaces) never throw C++ exceptions, due to the
+need for a stable contract across compilers. QueryInterface returns a 32-bit signed integer as an
+error code known. On Windows, this code is known as an
+[HRESULT](https://docs.microsoft.com/en-us/windows/desktop/com/error-handling-in-com).
+Generally with HRESULTs, non-negative integers indicate errors, positive integers and zero indicate
```
arcadiogarcia: I think here you meant
```suggestion
Generally with HRESULTs, negative integers indicate errors, positive integers and zero indicate
```

devhawk: Yup you're right. Thanks!

design_notes/XDN05 - xlang Binary Interface.md
```diff
@@ -12,6 +12,233 @@ status: draft
...
+
+## Overview
+
+xlang components are intended to be authored and built independently from their clients - both
+applications as well as other xlang components. Because an xlang component may be built at a
+different time and potentially using a different compiler from the calling code, xlang needs a
+stable contract - known as an application binary interface or ABI - to enable interoperability.
+
+xlang uses COM interfaces as the stable contract for xlang object types. Flat C functions are also
+used in xlang component binary loading as well as in xlang's
```
kennykerr: Remove line feed.

kennykerr: As the GitHub PR/diff viewer doesn't provide a more readable way to view the markdown, being a bit more careful about the source may help reviews.

design_notes/XDN05 - xlang Binary Interface.md
```diff
@@ -12,6 +12,233 @@ status: draft
...
+to later versions of a given component. By inheriting from another interface, the COM interface is
+defining a single longer array of function pointers, where the first set of function pointers are
+defined by the base interface while the next set of function pointers are defined by the derived
+interface. Multiple levels of COM interface single inheritance is supported.
+
+> Note, while xlang uses COM interfaces as a stable contract between components, xlang is
+explicitly avoiding the rest of the COM standard, including concepts such as apartments, monikers
+and marshalling. These constructs are tightly bound to the Windows platform and so have little use
+in the xlang project. Likewise, xlang is explicitly avoiding Windows technologies that build on COM
+such as DCOM, MTS, COM+, ActiveX and the Windows Runtime. Windows Runtime is an inspiration for
```
kennykerr: Such statements continue to confuse/concern developers. While the xlang project defines a new PAL/ABI that is distinct from the Windows Runtime, the xlang project also provides libraries and tools that target and support the Windows Runtime.

design_notes/XDN05 - xlang Binary Interface.md
```diff
@@ -12,6 +12,233 @@ status: draft
...
+requested interface is not supported.
+
+As stated above, interfaces are identified by a GUID. This GUID is passed by reference as the first
+parameter of QueryInterface. The pointer to the requested interface - if available - is returned
+via a void** out parameter. It is the responsibility of the calling code to cast this pointer to
+the correct COM interface pointer type.
+
+Methods on COM interfaces (including xlang ABI interfaces) never throw C++ exceptions, due to the
+need for a stable contract across compilers. QueryInterface returns a 32-bit signed integer as an
+error code known. On Windows, this code is known as an
```
kennykerr: known?

design_notes/XDN05 - xlang Binary Interface.md
```diff
@@ -12,6 +12,233 @@ status: draft
...
+[HRESULT](https://docs.microsoft.com/en-us/windows/desktop/com/error-handling-in-com).
+Generally with HRESULTs, negative integers indicate errors, positive integers and zero indicate
```
kennykerr: Generally?

design_notes/XDN05 - xlang Binary Interface.md
```diff
@@ -12,6 +12,233 @@ status: draft
...
+
+COM objects manage their own lifetime via an internal reference count. When this count goes to
+zero, the COM object cleans itself up. IUnknown provides methods to increment and decrement the
+object's reference count. If a language projection makes a copy of a COM interface pointer, it must
+call AddRef. When client code is finished using a given pointer to a COM interface, the language
+projection must call Release.
+
+The return value of these function is specified to be the new reference count of the object.
+However the COM standard itself warns that this value is intended to be used only for test purposes.
+In practice on Windows, the value returned from these two functions is often inaccurate. The one
```
kennykerr: I don't think we need to cast judgment on the accuracy of various implementations. ATL/WRL & C++/WinRT actually provide accurate results in practice. The more subtle point that the COM spec makes (in a subtle way) is that the usefulness of that "accurate" value is questionable. If any point needs to be made its that a COM interface pointer is invalid once Release has been called no matter what Release returns.

design_notes/XDN05 - xlang Binary Interface.md
```diff
@@ -12,6 +12,233 @@ status: draft
...
+scenario where you can trust the return value is when Release returns zero - in this case, you can
+reliably assume the object has cleaned itself up.
+
+## IXlangObject
+
+In addition to the core functionality described by IUnknown, ABI interfaces for xlang types derive
+from IXlangObject. This interface adds additional core functionality needed for xlang.
+
+``` cpp
+enum ObjectInfoCategory
```
kennykerr: XlangObjectInfoCategory

design_notes/XDN05 - xlang Binary Interface.md
```diff
@@ -12,6 +12,233 @@ status: draft
...
+{
+    StringRepresentation,
```
kennykerr: StringName

design_notes/XDN05 - xlang Binary Interface.md
```diff
@@ -12,6 +12,233 @@ status: draft
...
+    TypeName,
+    MemoryUsage
```
kennykerr: Size?

design_notes/XDN05 - xlang Binary Interface.md
```diff
@@ -12,6 +12,233 @@ status: draft
...
+
+### GetInfo
+
+In WinRT, the IInspectable interface provides property get methods that retrieved information about
+the object needed in specific runtime scenarios. Over time, the list of information categories that
+might be useful at runtime grew. For example, knowing the memory footprint of a given WinRT object
+would allow for garbage collected languages like C# make better scheduling decisions. But since
+IInspectable is the base COM interface for every interface in WinRT, it's impossible to extend.
+
+For xlang, instead of individual property get methods, IXlangObject exposes a single GetObjectInfo
```
kennykerr: GetInfo?

design_notes/XDN05 - xlang Binary Interface.md
```diff
@@ -12,6 +12,233 @@ status: draft
...
+method. The method takes an enum value indicating the object info being requested and a void** to
+hold the returned value. xlang objects are not required to support all object information
+categories. Similar to QueryInterface, an object can simply return an error code indicating that a
+requested information category is not supported by the object.
```
kennykerr: It must then also return a nullptr through its void** parameter.

design_notes/XDN05 - xlang Binary Interface.md
```diff
@@ -12,6 +12,233 @@ status: draft
...
+to uninformed decision making regarding collection scheduling. If these systems could retrieve
+the memory usage of arbitrary xlang objects, their scheduling heuristics would be better and the
+system would run more smoothly overall.
+
+The MemoryUsage information category retrieves the memory usage in bytes of the object in question.
+The property type for MemoryUsage is an unsigned 32-bit integer.
+
+## Weak References
+
+One issue faced in COM based systems is circular references - i.e. two objects who each hold a
```
kennykerr: Not just COM. Any reference-counting model. e.g. std::shared_ptr.

design_notes/XDN05 - xlang Binary Interface.md
```diff
@@ -12,6 +12,233 @@ status: draft
...
+pointer to the other. In this case, neither reference count will ever go to zero and neither
+object will ever be cleaned up. To avoid this issue, xlang supports weak references.
+
+A weak reference to an xlang object does not keep it alive. In other words, when all non-weak
+reference count of an object goes to zero, it cleans itself up even if there are outstanding weak
+references to it. As such, weak reference holders must check to see if a given reference is still
+valid via the IWeakReference::Resolve method.
```
kennykerr: You should introduce the interface before referring to its member.

design_notes/XDN05 - xlang Binary Interface.md
```diff
@@ -12,6 +12,233 @@ status: draft
...
+
+Weak references are retrieved for an arbitrary xlang object by querying for the
+IWeakReferenceSource interface and calling GetWeakReference. If an object does not support weak
+references, it simply returns the no interface error code from query interface.
```
kennykerr: ... it simply rejects the query.

design_notes/XDN05 - xlang Binary Interface.md
```diff
@@ -12,6 +12,233 @@ status: draft
...
+
+Note, because IWeakReferenceSource and IWeakReference are ABI interfaces that represents a proxy to
```
kennykerr: Another way to put it is that because weak references are not part of the projected surface of any API. Specifically, IXlangObject is only the base interface for projected types and a weak reference is not a projected type.

design_notes/XDN05 - xlang Binary Interface.md
```diff
@@ -12,6 +12,233 @@ status: draft
...
+an xlang object that may no longer exist, none of the IXlangObject methods are relevant. As such,
+this is one of the few places where an xlang ABI interface inherits directly from IUnknown.
+
```
kennykerr: You should also include the GUID values and note that this definitions is shared with COM and WinRT.

design_notes/XDN05 - xlang Binary Interface.md
kennykerr: I tried the unified view for markdown and its more readable - thanks.

design_notes/XDN05 - xlang Binary Interface.md
```diff
@@ -12,6 +12,233 @@ status: draft
...
+[COM interfaces](https://docs.microsoft.com/en-us/windows/desktop/com/com-objects-and-interfaces)
+as well as the definition of fundamental COM interface
+[IUnknown](https://docs.microsoft.com/en-us/windows/desktop/com/using-and-implementing-iunknown)
+as a foundation for stable contracts that both components and language projections can depend on.
+
+A COM interface pointer is basically a pointer to an array of function pointers, similar to how
+most C++ compilers implement
+[virtual method tables](https://en.wikipedia.org/wiki/Virtual_method_table).
+Each member of a COM interface is stored in an array slot as a function pointer. Each function in
+the array takes the array pointer as it's first parameter. While this design is intrinsic to most
```
Scottj1s: its

design_notes/XDN05 - xlang Binary Interface.md
```diff
@@ -12,6 +12,233 @@ status: draft
...
+the correct COM interface pointer type.
+
+Methods on COM interfaces (including xlang ABI interfaces) never throw C++ exceptions, due to the
+need for a stable contract across compilers. QueryInterface returns a 32-bit signed integer as an
+error code known. On Windows, this code is known as an
+[HRESULT](https://docs.microsoft.com/en-us/windows/desktop/com/error-handling-in-com).
+Generally with HRESULTs, negative integers indicate errors, positive integers and zero indicate
+success. QueryInterface in particular should always return 0 on success.
+
+QueryInterface may return the following error codes:
```
Scottj1s: if it's just a list of one now, collapse into a single sentence "... such as ..."

design_notes/XDN05 - xlang Binary Interface.md
Scottj1s: It's shorter but less precise

design_notes/XDN05 - xlang Binary Interface.md
```diff
@@ -12,6 +12,233 @@ status: draft
...
+enum ObjectInfoCategory
+{
+    StringRepresentation,
+    TypeName,
+    MemoryUsage
+};
+
+struct IXlangObject
+{
+    int32_t GetInfo(ObjectInfoCategory info_category, void** info);
```
Scottj1s: GetProperty?

design_notes/XDN05 - xlang Binary Interface.md
devhawk: I don't think StringName is the right choice here. StringValue maybe?

I picked "StringRepresenation" beause .NET, Java and JavaScript describe the ToString as returning a "string represetnation". Python calls it "string version",  Obj-C calls it "description".

design_notes/XDN05 - xlang Binary Interface.md
```diff
@@ -12,6 +12,233 @@ status: draft
...
+};
+```
+
+### GetInfo
+
+In WinRT, the IInspectable interface provides property get methods that retrieved information about
```
Scottj1s: retrieve

design_notes/XDN05 - xlang Binary Interface.md
devhawk: MemorySize or just Size?

design_notes/XDN05 - xlang Binary Interface.md
```diff
@@ -12,6 +12,233 @@ status: draft
...
+the object needed in specific runtime scenarios. Over time, the list of information categories that
+might be useful at runtime grew. For example, knowing the memory footprint of a given WinRT object
```
Scottj1s: grew -> has grown

design_notes/XDN05 - xlang Binary Interface.md
```diff
@@ -12,6 +12,233 @@ status: draft
...
+would allow for garbage collected languages like C# make better scheduling decisions. But since
```
Scottj1s: to make

design_notes/XDN05 - xlang Binary Interface.md
```diff
@@ -12,6 +12,233 @@ status: draft
...
+IInspectable is the base COM interface for every interface in WinRT, it's impossible to extend.
+
+For xlang, instead of individual property get methods, IXlangObject exposes a single GetObjectInfo
+method. The method takes an enum value indicating the object info being requested and a void** to
+hold the returned value. xlang objects are not required to support all object information
+categories. Similar to QueryInterface, an object can simply return an error code indicating that a
```
Scottj1s: specify that error?  e.g., E_NOTIMPL

design_notes/XDN05 - xlang Binary Interface.md
```diff
@@ -12,6 +12,233 @@ status: draft
...
+requested information category is not supported by the object.
+
+Having a single GetInfo method has two primary benefits over the IInspectable approach. Because
+enums can be additively versioned over time, it is possible to add new ones over time. Since
+supporting a given information category is optional, adding new ones would not break existing
+objects. Furthermore, it reduces the size of virtual function table that every xlang object needs
```
Scottj1s: the

design_notes/XDN05 - xlang Binary Interface.md
```diff
@@ -12,6 +12,233 @@ status: draft
...
+about an object at runtime. In order to do this, xlang object must have a mechanism to find the
+metadata for arbitrary xlang objects.
+
+The TypeName information category retrieves the fully namespace qualified name of the type in
+question. The property type for TypeName is an xlang_string.
+
+While all information categories are optional, it is highly recommended that all objects support
+TypeName in order to be usable from dynamic languages.
+
+#### Memory Usage
```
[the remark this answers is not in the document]
devhawk: Agreed. I like ObjectSize.

design_notes/XDN05 - xlang Binary Interface.md
BrentRe: It's a bug if an AddRef implementation doesn't return the correct value. What confuses folks is when proxy & stubs are involved. A subsequent AddRef on a proxy doesn't forward the AddRef to the object because there's no need to do so. So the physical reference count on an object will only include one for each referencing proxy, not one per AddRef on the proxy.

With COM, you typically don't know whether you are calling AddRef on the actual instance or calling AddRef on the proxy. Thus the questionable usefulness of the returned value.

design_notes/XDN05 - xlang Binary Interface.md
BenJKuhn: This is something that few people will ever type. Let's make this something a bit longer & narrower. E.g. GetObjectState or similar.

design_notes/XDN05 - xlang Binary Interface.md
devhawk: You're right, but is that really imporant to include here?

design_notes/XDN05 - xlang Binary Interface.md
devhawk: artifact of having GetWeakReference on IXlangObject originally. Will rework

design_notes/XDN05 - xlang Binary Interface.md
```diff
@@ -12,6 +12,233 @@ status: draft
...
+question. The property type for StringRepresentation is an xlang_string.
+
+While all information categories are optional, it is highly recommended that all objects support
+StringRepresentation. By default, this method can return the type's name, however it can be
+implemented in whatever manner makes most sense for the underlying object.
+
+#### Type Information
+
+Dynamic languages such as JavaScript and Python often need to be able to discover information
+about an object at runtime. In order to do this, xlang object must have a mechanism to find the
```
Scottj1s: "xlang must have ..."

design_notes/XDN05 - xlang Binary Interface.md
```diff
@@ -12,6 +12,233 @@ status: draft
...
+have a mechanism to retrieve the string representation of an object. 
+
+The StringRepresentation information category retrieves the string representation of the object in
+question. The property type for StringRepresentation is an xlang_string.
+
+While all information categories are optional, it is highly recommended that all objects support
+StringRepresentation. By default, this method can return the type's name, however it can be
+implemented in whatever manner makes most sense for the underlying object.
+
+#### Type Information
```
Scottj1s: this section hints at a reflection facility, making xlang 'usable from dynamic languages'.  but type name alone is hardly sufficient.  what's the use case other than downcasting (to statically known types)?

design_notes/XDN05 - xlang Binary Interface.md
BrentRe: I don't think we want a request for an unsupported category to cause an exception.  That leads to a horrible developer experience, especially as we're saying objects don't need to support all categories.

Thus we should never return a failure result.

design_notes/XDN05 - xlang Binary Interface.md
```diff
@@ -12,6 +12,233 @@ status: draft
...
+The MemoryUsage information category retrieves the memory usage in bytes of the object in question.
+The property type for MemoryUsage is an unsigned 32-bit integer.
+
+## Weak References
+
+One issue faced in COM based systems is circular references - i.e. two objects who each hold a
+pointer to the other. In this case, neither reference count will ever go to zero and neither
+object will ever be cleaned up. To avoid this issue, xlang supports weak references.
+
+A weak reference to an xlang object does not keep it alive. In other words, when all non-weak
```
Scottj1s: "the non-weak reference count ..."

design_notes/XDN05 - xlang Binary Interface.md
kennykerr: Brent, this function is never projected. A language projection calls it internally and thus there's no concern about an exception. Of course, I'm happy with a simpler "bool GetObjectInfo(category, void**) signature as well. An error code is overrated here.

design_notes/XDN05 - xlang Binary Interface.md
devhawk: Going with ObjectSize as per [this comment thread](#discussion_r247176143)

design_notes/XDN05 - xlang Binary Interface.md
devhawk: I like making it longer but I don't like calling it object "state". How about GetObjectInfo?

MikeHillberg: Why are we assuming it will only be used for getting?

design_notes/XDN05 - xlang Binary Interface.md
devhawk: Rephrased to indicate there are valid scenarios where the ref count returned may not be accurate, using tear-off interfaces as an example.

design_notes/XDN05 - xlang Binary Interface.md
devhawk: I like making it a boolean

design_notes/XDN05 - xlang Binary Interface.md
devhawk: You're right, the name alone is not sufficient - you also need metadata. GetRuntimeClassName was added to enable JS to locate metadata for a given type on the fly. 

I don't follow your use case question. Can you rephrase?

devhawk: I've added text to indicate the name is not sufficient, that a dynamic language projection would need a mechanism to find (by name) and consume the metadata for this type, and that we need an XDN for this.

design_notes/XDN05 - xlang Binary Interface.md
devhawk: "xlang dynamic language projections must have..."

design_notes/XDN05 - xlang Binary Interface.md
devhawk: Good suggestion.

Would you say IActivationFactory (as well as custom activation factory interfaces) should inherit from IXlangObject, IUnknown, or some other base interface specific to activation factory interfaces?

design_notes/XDN05 - xlang Binary Interface.md
devhawk: I'm a little worried about folks abusing this method for unintended usage as it is. Adding a set method as well (or supporting set by passing in a value via the info param) seems like asking for trouble.

design_notes/XDN05 - xlang Binary Interface.md
devhawk: The xlang project in the large, sure. But is that true of the xlang abi? 

I want to avoid folks thinking we're porting over a bunch of Windows COM infrastructure to non-windows platforms. Can you suggest alternative phrasing?

design_notes/XDN05 - xlang Binary Interface.md
devhawk: leaving as StringRepresenation for now. Please feel free to unresolve this comment and suggest other altenatives if you are so inclined.

design_notes/XDN05 - xlang Binary Interface.md
```diff
@@ -12,6 +12,233 @@ status: draft
...
+[.NET](https://docs.microsoft.com/en-us/dotnet/api/system.object.tostring?view=netframework-4.7.2),
+[Java](https://docs.oracle.com/javase/7/docs/api/java/lang/Object.html#toString()) and
+[Objective-C](https://developer.apple.com/documentation/objectivec/1418956-nsobject/1418746-description)
+have a mechanism to retrieve the string representation of an object. 
+
+The StringRepresentation information category retrieves the string representation of the object in
+question. The property type for StringRepresentation is an xlang_string.
+
+While all information categories are optional, it is highly recommended that all objects support
+StringRepresentation. By default, this method can return the type's name, however it can be
```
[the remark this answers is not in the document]
devhawk: will rephrase. I also reordered the object info categories to have TypeName first so I can more easily refer back to that in this section.

design_notes/XDN05 - xlang Binary Interface.md
BenJKuhn: GetObjectInfo seems fine to me. For setting values, we could always add an enum value that supplies an object that then has a setter (e.g. GetObjectInfo(ResourceSettigns, &objectResourceSettings); objectResourceSettings->SetResourcePool(1);).

devhawk: I still don't like the idea of setting things via GetObjectInfo, though if nessassary I kinda like the "get a setter object" approach - I don't want that pattern to be straightforward

design_notes/XDN05 - xlang Binary Interface.md
kennykerr: I don't see any problem with factory interfaces inheriting from IXlangObject, but there is no real need for them to do so, other than that it complicates the IDL/winmd authoring/generation process as you would have to distinguish between factory and non-factory interfaces. I'd rather not create another sub-category of interface.

design_notes/XDN05 - xlang Binary Interface.md
```diff
@@ -12,6 +12,249 @@ status: draft
...
+The return value of these function is specified to be the new reference count of the object. However
+the COM standard itself warns that this value is intended to be used only for test purposes. In
+practice on Windows, the value returned from these two functions is not guaranteed to be the object's
+true reference count. For example, COM object can choose to implement an "tear-off" interfaces.
+A tear-off interface is implemented on a separate object, which is constructed when the interface
+in question is queried. When AddRef or Release is called on the tear-off interface, it will return
+the reference count of the tear off interface object *not* the reference count of the original
+object.
+
+## IXlangObject
```
bbowman: Should IXlangObject not be named with Object since it can be used to represent interfaces as well as objects?

bbowman: Yeah understood but the terms across languages are of course overloaded and confusing which is why I'm wondering if there is something better (there might not be). I *think* we are probably fine since Java uses Object as its root and ObjC has NSObject so there is some prior art.

design_notes/XDN05 - xlang Binary Interface.md
```diff
@@ -12,6 +12,233 @@ status: draft
...
+call AddRef. When client code is finished using a given pointer to a COM interface, the language
+projection must call Release.
+
+The return value of these function is specified to be the new reference count of the object.
+However the COM standard itself warns that this value is intended to be used only for test purposes.
+In practice on Windows, the value returned from these two functions is often inaccurate. The one
+scenario where you can trust the return value is when Release returns zero - in this case, you can
+reliably assume the object has cleaned itself up.
+
+## IXlangObject
```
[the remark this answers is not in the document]
bbowman: Point still stands I guess on if QI for IComparable vs GetInfo or a top level IsEquals on IXlangObject is worth the vtable size vs perf tradeoff.

design_notes/XDN05 - xlang Binary Interface.md
```diff
@@ -12,6 +12,233 @@ status: draft
...
+Methods on COM interfaces (including xlang ABI interfaces) never throw C++ exceptions, due to the
+need for a stable contract across compilers. QueryInterface returns a 32-bit signed integer as an
+error code known. On Windows, this code is known as an
+[HRESULT](https://docs.microsoft.com/en-us/windows/desktop/com/error-handling-in-com).
+Generally with HRESULTs, negative integers indicate errors, positive integers and zero indicate
+success. QueryInterface in particular should always return 0 on success.
+
+QueryInterface may return the following error codes:
+
+- 0x80004002 (E_NOINTERFACE) indicates the requested interface is not supported by this object.
```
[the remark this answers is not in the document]
bbowman: I suppose a follow question is, do we ever expect custom QI implementations that may return errors besides success and no interface? If its a true binary state would this return be better as a bool (we don't have RPC and other things that need to intercede here so I'm not sure if we want an "always return an error code" style everywhere)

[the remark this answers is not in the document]
kennykerr: We need to retain IUnknown as-is for broad compatibility and interoperability.

design_notes/XDN05 - xlang Binary Interface.md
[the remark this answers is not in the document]
devhawk: It's not just Objective C and Java. C# has [System.Object.Equals virtual method](https://docs.microsoft.com/en-us/dotnet/api/system.object.equals) and Python has [__eq__](https://docs.python.org/3/reference/datamodel.html#object.__eq__), so there's lots of prior art.

design_notes/XDN05 - xlang Binary Interface.md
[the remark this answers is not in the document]
bbowman: hooray baggage :-) Figured this would be the case but want to make sure we are probing our assumptions since we are getting a somewhat "clean slate" here.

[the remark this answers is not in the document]
kennykerr: I'd love a time machine... 😉

design_notes/XDN05 - xlang Binary Interface.md
BenJKuhn: It's worth keeping in mind that this is infrastructure only seen by projection authors. End users won't deal with the interface, for the most part, so the naming here does not need to be a coherent ideal name for all possible languages.

design_notes/XDN05 - xlang Binary Interface.md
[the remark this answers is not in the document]
devhawk: I opened #133 to track supporting object equality in the ABI

devhawk:
FYI, while this document is not done @BenJKuhn has suggested we get what is reasonably settled checked in. I opened #133 to track ABI support for object equality + added empty stub sections for activation factories and class inheritance. Otherwise, this is ready for check-in with approval

Issue #65, opened 2018-10-26
[Suggestion] Introduce Int8 MIDL type


Alovchin91:
Since this project is not aiming to preserve binary compatibility with Windows Runtime, I would like to suggest introducing signed 8-byte `Int8` primitive type into Xlang type system.

Current MIDL 3.0 doesn’t support Int8 type, and only has unsigned 8-bit integer type. But introducing the signed one would make a lot of sence if you expect to support UTF-8 strings in Xlang ABI.

This would require changes to MIDL 3.0 definition language of course.

kennykerr:
A signed 8-bit integer was not considered useful by the original WinRT architects. We may revisit that since the underlying ECMA-335 spec certainly supports it. Thanks for the suggestion!

Alovchin91:
@BrentRe Basically I currently use them to return strings (`sbyte*` arrays) allocated in a specific memory pool (not with .NET Marshal) back to the C# caller. They are easily converted to managed strings using just a String constructor.

That might not be a viable case for Xlang, but I think other users could also give some insights into how they could use them.

fdwr:
The lack of actual signed 8-bit quantities in WinRT was annoying for implementing WinML tensors of that data type (see the INT8 data here for the ONNX Machine Learning format, https://github.com/onnx/onnx/blob/master/onnx/onnx.proto). Tensors are multidimensional arrays of data types, and we shouldn't need to reinterpret_cast int8_t to an array of unsigned char (which semantically doesn't really make sense) when it should be possible to just pass the array_view<int8_t> directly to WinRT API's. The data type is actually used, as quantized tensors greatly reduce multimegabyte model files. I wouldn't want to see XLang have the same limitation, especially given it's intended to have even broader interop.

>UTF-8 encodes all 1,112,064 valid code points in Unicode
Yeah, I see no use for signed bytes as code points and would discourage that (Unicode should always be unsigned), but I see other numeric processing uses outside machine learning too (e.g. signed 8-bit types used in older audio formats, like 8-bit PCM and telephony from ITU-T Recommendation G.711).

They are a basic data type in:
- Python numpy int8 - https://docs.scipy.org/doc/numpy/user/basics.types.html
- C++ int8_t - https://en.cppreference.com/w/cpp/types/integer
- Java byte (though arguably awful that Java made 'byte' signed by default, rather than a distinct sbyte type).
[...]

Issue #53, opened 2018-10-24
Revisit string encoding decision


devhawk:
WinRT uses UTF-16 encoded strings. Since xlang is not attempting to be binary compatible with WinRT, we can revisit this decision. In particular, we can consider several options

1. sticking with UTF16 encoded strings
2. changing to UTF8 encoded strings
3. Update string type + ABI to support UTF8 and UTF16 with conversion between the encodings

phoyd:
UTF-16 always feeled to me like kind of legacy encoding, an echo from the time before Unicode, when UCS-2 was used for multilingual text and then upgraded to UTF-16 to support the full code point range, while breaking as little client code as possible. It doesn't make anything easier, because it is still not a fixed width encoding and I haven't seen much of it outside of the Windows-APIs. Most of Unix/Internet development has settled with UTF-8 and forcing UTF-16 to them will raise eyebrows.

devhawk:
Looks like JNI supports both 8 and 16 bit strings. [NewString](https://docs.oracle.com/javase/7/docs/technotes/guides/jni/spec/functions.html#wp4925) constructs a jstring from an array of 16-bit jchars while [NewStringUTF](https://docs.oracle.com/javase/7/docs/technotes/guides/jni/spec/functions.html#wp17220) constructs a jstring from an array of 8-bit chars.

kennykerr:
The question is what encoding does it use internally to ensure good performance (and avoid copies).

DHowett-MSFT:
JNI uses "Modified UTF-8", which is unfortunately just far enough away from UTF-8 as to be an annoyance.

zooba:
Internally, CPython uses adaptive allocation so that it does "direct" encoding in code units with no surrogates or multi-element characters. But other implementations just use UTF-8 and are fine.

Part of the problem with both UTF-8 and UTF-16 is that malloc() and (Unicode-aware) strlen() do not use the same values, and the semantics of how they fail to give the same result are different between contexts (in Python, you'll get a reallocation/copy into a larger unit size, while in Windows it will pretend that surrogate pairs are independent characters unless you happen to get unlucky and hit something that properly validates the string).

So the simplest option to me seems to be choosing one representation for the ABI and let projections handle this as efficiently as they can, since they are in the best place to do this (for example, keeping the original around for round-tripping, lazily generating the language-specific representation, etc.).

[...]

PR #120, opened 2019-01-11
XDN03 - xlang type system

devhawk:
Note, this XDN is not complete. I'm putting it into PR to enable comments on the content that is available.


bbowman:
Are we sure we want to support unsigned fundamental types (particularly 64 bit versions)? Some languages (like Java) don’t have a strong unsigned data story.

design_notes/XDN03 - xlang Type System.md
```diff
@@ -15,3 +15,142 @@ status: draft
...
+Single     | a 32-bit IEEE 754 floating point number
+Double     | a 64-bit IEEE 754 floating point number
+Char16     | a 16-bit non-numeric value representing a UTF-16 code unit
+Boolean    | an 8-bit Boolean value
+String     | an immutable sequence of Char16s used to represent text
+Guid       | A 128-bit standard [universally unique identifier](https://en.wikipedia.org/wiki/Universally_unique_identifier)
+Object     | An xlang object of unknown type
+
+> Note: string encoding for xlang is currently under review. In particular, it has not been decided
+if xlang will support UTF-8 strings in some capacity. A design note will be commissioned for this
```
zooba: It would be nice to _only_ support UTF-8 and do silent conversions for back-compat. I think I heard .NET Core is getting a UTF-8 string type, so it'd be real nice to have a way to keep it consistent.

design_notes/XDN03 - xlang Type System.md
```diff
@@ -15,3 +15,142 @@ status: draft
...
+decision.
+
+## Enums
+
+An enum type is a distinct value type with a set of named value. Each named value in an enum
+corresponds to an constant integer value.
+
+Each enum type has an underlying integral type. The only legal underlying integral types for enums
+in xlang are Int32 and UInt32.
```
zooba: Any reason to keep both? Why not just Int32 for normal and UInt32 for flags? Or disallow math operations and define per-language casting rules?

design_notes/XDN03 - xlang Type System.md
```diff
@@ -15,3 +15,142 @@ status: draft
...
+interface's object members is unrelated to any other class's implementation of the same interface.
+
+Interfaces may optionally inherit from a single other interfaces. All of the type members of the
+base interface are considered part of the derived interface as well. Any class implementing an
+interface that uses inheritance must implement all of the type members specified by both the base
+and derived interfaces.
+
+> Note, WinRT interfaces support inheritance from multiple other interfaces. This approach leads to
+inefficiencies at the ABI layer that xlang wishes to avoid by constraining interfaces to single
+inheritance. This decision may be revisited as the xlang project progresses.
```
zooba: Is this dealt with if interface inheritance is really just a marker that requires a concrete object to also implement those? That could be a compile time check. But what do we lose?

design_notes/XDN03 - xlang Type System.md
zooba: Just saw the next paragraph, but apparently can't delete this comment from my phone.

design_notes/XDN03 - xlang Type System.md
devhawk: Yeah, I'd like to only support UTF-8 as well. Design note for string encoding will dig into this further. 

I seem to remember something about a UTF-8 string type for .NET, but not sure now where I saw it.

design_notes/XDN03 - xlang Type System.md
devhawk: In WinRT, interface inheritance isn't really inheritance - it's more of a marker. If Interface A requires Interface B, what that really means at the ABI layer is that you have two ABI interfaces, both of which inherit from IInspectable. This leads to both runtime overhead that stems from having to calling QueryInterface as well as memory overhead that stems from types having multiple interface vtables.

By limiting interfaces at the type system level to single inheritance, we can implement them at the ABI layer using a single composite virtual method table.

devhawk:
>           Are we sure we want to support unsigned fundamental types (particularly 64 bit versions)? Some languages (like Java) don’t have a strong unsigned data story.

devhawk:
@bbowman We debated about unsigned types back when we were designing WinRT, but I wasn't involved in the discussion. @BrentRe might have more context on that debate.

design_notes/XDN03 - xlang Type System.md
kennykerr: https://github.com/dotnet/corefxlab/issues/2350

Scottj1s: We will need both.  UTF16 is a requirement for languages like Java, which use it natively.  Otherwise, we unnecessarily convert back and forth.

design_notes/XDN03 - xlang Type System.md
```diff
@@ -15,3 +15,142 @@ status: draft
...
+Guid       | A 128-bit standard [universally unique identifier](https://en.wikipedia.org/wiki/Universally_unique_identifier)
+Object     | An xlang object of unknown type
+
+> Note: string encoding for xlang is currently under review. In particular, it has not been decided
+if xlang will support UTF-8 strings in some capacity. A design note will be commissioned for this
+decision.
+
+## Enums
+
+An enum type is a distinct value type with a set of named value. Each named value in an enum
```
Scottj1s: values

design_notes/XDN03 - xlang Type System.md
```diff
@@ -15,3 +15,142 @@ status: draft
...
+corresponds to an constant integer value.
```
Scottj1s: a

design_notes/XDN03 - xlang Type System.md
```diff
@@ -15,3 +15,142 @@ status: draft
...
+
+Each enum type has an underlying integral type. The only legal underlying integral types for enums
+in xlang are Int32 and UInt32.
+
+An enum type with an underlying integral type of UInt32 is considered a Flags enum. Flags enums
+are intended to be treated as a bit field. Language projections can and should provide logical
+operators such as and, or and not for Flags enum values.
+
+Enums are additively versionable. Subsequent versions of a given enum may add new named values (and
```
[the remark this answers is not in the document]
Scottj1s: agree.  even a case like this needs more fleshing out.  I read that I'd have to code defensively against unexpected additions to an enum I'm consuming (e.g., passed in to me)

design_notes/XDN03 - xlang Type System.md
Scottj1s: You may want to hoist that comment into th DN, to clarify the distinction between logical interface "requirements" and binary interface (vtable) inheritance, and the benefit of implementing the latter

design_notes/XDN03 - xlang Type System.md
```diff
@@ -15,3 +15,142 @@ status: draft
...
+Interfaces may optionally inherit from a single other interfaces. All of the type members of the
+base interface are considered part of the derived interface as well. Any class implementing an
+interface that uses inheritance must implement all of the type members specified by both the base
+and derived interfaces.
+
+> Note, WinRT interfaces support inheritance from multiple other interfaces. This approach leads to
+inefficiencies at the ABI layer that xlang wishes to avoid by constraining interfaces to single
+inheritance. This decision may be revisited as the xlang project progresses.
+
+Interfaces are not versionable. Once they have been defined, they may never be changed.
```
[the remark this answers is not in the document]
Scottj1s: isn't that effectively a new interface, so they really are immutable?  maybe just semantics

design_notes/XDN03 - xlang Type System.md
```diff
@@ -15,3 +15,142 @@ status: draft
...
+interface specifies the signatures for a group of object members (aka methods, properties and
+events) but does not specify an implementation for those object members.
+
+Interfaces can be implemented by multiple different classes. Each class implementation of a given
+interface's object members is unrelated to any other class's implementation of the same interface.
+
+Interfaces may optionally inherit from a single other interfaces. All of the type members of the
+base interface are considered part of the derived interface as well. Any class implementing an
+interface that uses inheritance must implement all of the type members specified by both the base
+and derived interfaces.
```
Scottj1s: maybe clarify: "specified by the interface's base interfaces".  i.e., anything further derived from that interface is of no consequence to the class

design_notes/XDN03 - xlang Type System.md
```diff
@@ -15,3 +15,142 @@ status: draft
...
+#### Parameterized Interface Examples
+
+For example, xlang's foundation types includes a set of parameterized interfaces to represent
+collections.
+
+- `IIterable<T>` represents a collection that can be enumerated
+- `IVector<T>` represents a collection that can be randomly accessed
+- `IMap<K,V>` represents a collection that maps from keys to values
+
+Both `IVector<T>` and `IMap<K,V>` can be enumerated. `IVector<T>`'s derives from `IIterable<T>`.
```
Scottj1s: delete 's

design_notes/XDN03 - xlang Type System.md
```diff
@@ -15,3 +15,142 @@ status: draft
...
+`IIterable` is instantiated as parameterized type `IKeyValuePair<K,V>`, using the type parameters
+defined as part of of `IMap<K,V>`.
+
+Using the same collection interfaces defined above, you could define an non parameterized interface
+`IPropertySet` that inherits from `IMap<String, Object>`. Note the use of concrete types in the
+declaration of IMap, where in the previous example we used type parameters.
+
+## Delegates
+
+## Classes
```
Scottj1s: make sure we discuss interface member collisions (cowboy-artist), maybe mention explicit implementation as a projection language feature

design_notes/XDN03 - xlang Type System.md
```diff
@@ -15,3 +15,142 @@ status: draft
...
+## Fundamental Types
+
+The xlang Type system includes a core set of built-in primitive types.
+
+xlang Type | Type Description
+---------- | ----------------------------------------------------------
+Int16      | a 16 bit signed integer
+Int32      | a 32 bit signed integer
+Int64      | a 64 bit signed integer
+UInt8      | an 8 bit unsigned integer
```
kennykerr: There have been requests for Int8. https://github.com/Microsoft/xlang/issues/65

design_notes/XDN03 - xlang Type System.md
```diff
@@ -15,3 +15,142 @@ status: draft
...
+
+Parameterized interfaces can inherit from both parameterized and non-parameterized interfaces.
+When inheriting from a parameterized interface, the type parameters of the base interface must
+be specified, though they may be specified in terms of the derived interface's type parameters.
+
+Non-parameterized interfaces can only inherit from concrete interfaces. This includes
+non-parameterized interfaces as  well as parameterized interfaces where all of the type parameters
+have been fully specified.
+
+Array types may not be used as type parameters.
```
kennykerr: There are no array types. There are array parameters, but they are not types (in metadata).

design_notes/XDN03 - xlang Type System.md
```diff
@@ -15,3 +15,142 @@ status: draft
...
+
+> Note, an xlang's type system interface is a different concept from
+[COM Interfaces](XDN05%20-%20xlang%20Binary%20Interface.md#com-interfaces) in xlang's
+[ABI](XDN05%20-%20xlang%20Binary%20Interface.md).
+So while an xlang ABI interface must inherit from
+[IXlangObject](XDN05%20-%20xlang%20Binary%20Interface.md#ixlangobject), interfaces in the xlang
+type system do not. Inheriting from IXlangObject is an implementation detail that is not directly
+surfaced in xlang's type system.
+
+An interface is a type that acts as a contract that can be implemented by multiple classes. An
```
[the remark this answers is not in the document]
devhawk: THis XDN is nowhere near done yet. We can review section ordering once more of the main doc is sketched in

design_notes/XDN03 - xlang Type System.md
devhawk: will do, but I hate this sentence. "a distinct value type with a set of named values". Suggestions?

design_notes/XDN03 - xlang Type System.md
devhawk: There will be an entire section on interface members that includes a section on overloading. There's a [similar section](https://github.com/Microsoft/xlang/blob/master/docs/XSPEC01%20-%20Type%20System%20Specification.md#method-overloading) in the WinRT Type system doc.

design_notes/XDN03 - xlang Type System.md
devhawk: Nice, I like that way of phrasing it.

design_notes/XDN03 - xlang Type System.md
BrentRe: We should add Int8. It's very useful for Machine Learning APIs

design_notes/XDN03 - xlang Type System.md
```diff
@@ -15,3 +15,142 @@ status: draft
...
+
+## Enums
+
+An enum type is a distinct value type with a set of named value. Each named value in an enum
+corresponds to an constant integer value.
+
+Each enum type has an underlying integral type. The only legal underlying integral types for enums
+in xlang are Int32 and UInt32.
+
+An enum type with an underlying integral type of UInt32 is considered a Flags enum. Flags enums
```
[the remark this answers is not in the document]
BrentRe: We definately need unsigned integers. We had lots of debate about that back in the WinRT days.

Some languages, e.g., JavaScript, will simply convert them to a larger signed numeric type. Int64, signed or unsigned, is a problem for JavaScript because it intrinsicly doesn't support that large of a numeric value.

design_notes/XDN03 - xlang Type System.md
bbowman: @DHowett-MSFT might be able to provide some insight on how this sort of thing was handled in Objective-C. I believe it *mostly* hides the internal representation from the developer except for exposing a very few "raw" functions that operate on the individual characters etc. This would, however, mean building a more full featured string class as part of the "foundation classes"

design_notes/XDN03 - xlang Type System.md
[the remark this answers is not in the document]
bbowman: I'm not sure if Brent is trying to reply to me or this thread but to add some context to my comment, I definitely think unsigned is important, but I'm concerned about exactly the boundaries you describe. In Rome, I forget where exactly, but we ran into the widening issue that translating a uint64 into java was unwieldy at best (there is a BigInteger class but that feels like a hack). We got away with it because it wasn't intended for arithmetic operations so the sign didn't actually matter (i think it was an id maybe) but I want to make sure that we don't promise something at the type system level that is going to project poorly (or at least make it easy to know at author time that it may not work the best)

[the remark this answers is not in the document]
kennykerr: We don't want the type system to artificially limit API authors. An API author designing an API for Java or JavaScript may naturally avoid certain types when his target audience is JavaScript or some other limited language, but we should not restrict the type system as a whole to the lowest common denominator.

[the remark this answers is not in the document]
BrentRe: We ran into the [U]Int64 & JavaScript issue back in the Win8 days. There are APIs that absolutely require 64-bit integers (e.g. file system APIs). We cannot cripple such APIs by lowering the type system to the least common denominator of all languages.

Yes, this means that some APIs will potentially by unusable (or usable with restrictions). Better than the alternative.

design_notes/XDN03 - xlang Type System.md
```diff
@@ -15,3 +15,142 @@ status: draft
...
+Guid       | A 128-bit standard [universally unique identifier](https://en.wikipedia.org/wiki/Universally_unique_identifier)
+Object     | An xlang object of unknown type
+
+> Note: string encoding for xlang is currently under review. In particular, it has not been decided
+if xlang will support UTF-8 strings in some capacity. A design note will be commissioned for this
+decision.
+
+## Enums
+
+An enum type is a distinct value type with a set of named values. Each named value in an enum
```
Scottj1s: enumerators, if you don't like values?

design_notes/XDN03 - xlang Type System.md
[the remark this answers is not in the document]
bbowman: Yeah I understand the desire for the flexibility and I think we are ok as long as there is an annotation or warning or something that we can use in metadata to say "not projectable everywhere" or something. Just want to call it out so we don't inadvertently lead devs down a path they can't get out of without breaking changes. 

Assuming that an author is thinking about *all* target languages at the time of authoring the API may be a bit ambitious as projects evolve etc. (For instance Rome would love to support JavaScript someday but that was not on the mind of the api developers at the very beginning of our cross plat effort)

design_notes/XDN03 - xlang Type System.md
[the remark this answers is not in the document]
bbowman: By identifier I assume you mean the GUID and by change I assume you mean non additive change (where fast ABI would still work for new and old alike). How would backcompat work in practice? I guess you would need to change the name of the old interface to something else so that you could continue to support older consumers while also implementing the new GUID for your interface?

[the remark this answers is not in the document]
kennykerr: COM interfaces are immutable full stop. There can be no change whatsoever. Adding to the end of an interface is a change and is not permitted by COM. You may decide to "change" the source code in some way but you are effectively dropping the old interface and minting a new one. If the interface is private to your project that's probably fine, but if that interface has been published in any way you are violating one of the central rules of COM.

design_notes/XDN03 - xlang Type System.md
devhawk: will do

design_notes/XDN03 - xlang Type System.md
devhawk: Looks like JNI supports both 8 and 16 bit strings as well. [NewString](https://docs.oracle.com/javase/7/docs/technotes/guides/jni/spec/functions.html#wp4925) constructs a jstring from an array of 16-bit jchars while [NewStringUTF](https://docs.oracle.com/javase/7/docs/technotes/guides/jni/spec/functions.html#wp17220) constructs a jstring from an array of 8-bit chars.

kennykerr: https://stackoverflow.com/a/9699138/500701

design_notes/XDN03 - xlang Type System.md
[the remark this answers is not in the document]
devhawk: I think it's important to state that enums (and classes when I get that section written) *can* version over time while structs, delegates and interfaces *can't*. But I totally agree that we need an XDN that covers *how* versioning works. I will add a not to that effect.

design_notes/XDN03 - xlang Type System.md
devhawk: Rephrased the note and added a link to issue #53

design_notes/XDN03 - xlang Type System.md
devhawk: removed "distinct value type"

design_notes/XDN03 - xlang Type System.md
```diff
@@ -15,3 +15,142 @@ status: draft
...
+corresponds to a constant integer value.
+
+Each enum type has an underlying integral type. The only legal underlying integral types for enums
+in xlang are Int32 and UInt32.
+
+An enum type with an underlying integral type of UInt32 is considered a Flags enum. Flags enums
```
devhawk: @jonwis : So I can't have a non-flags enum that is also UInt32 for making compat with existing code easier?  Can I specify the underlying type? What if I specify the underlying type as int32 but mark it with the flags attribute?

@kennykerr : There have also been requests for 64-bit enums and unsigned enums that aren't flags (addresses).

design_notes/XDN03 - xlang Type System.md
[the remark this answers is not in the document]
devhawk: FYI, since this thread has jumped topics, I started a [new conversation](#r247248656) to address @jonwis and @kennykerr  comments about enums.

design_notes/XDN03 - xlang Type System.md
devhawk: @jonwis midlrt enforces the constraint that non-flags enums are int32, flags enums are unit32

design_notes/XDN03 - xlang Type System.md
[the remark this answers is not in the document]
devhawk: xlang type system interfaces are *not* COM interfaces, though they are *implemented* as COM interfaces at the ABI layer. However, the same immutability rules @kennykerr describes for COM interfaces need to apply to xlang interfaces.

@jonwis, I'll add some text about identifiers. In MIDL3, isn't this identifier generated from the type name?

design_notes/XDN03 - xlang Type System.md
devhawk: cutting the array types line. When I get to writing up the section on parameters, I'll talk about arrays there.

design_notes/XDN03 - xlang Type System.md
devhawk: good suggestion @Scottj1s

design_notes/XDN03 - xlang Type System.md
devhawk: rephrased. Feel free to unresolve this conversation if you don't like my new phrasing.

design_notes/XDN03 - xlang Type System.md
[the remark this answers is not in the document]
devhawk: I'm going to leave this conversation open for now, but I'm basically with @BrentRe that we need unsigned and 64 bit ints, even though various languages may have issue with them. @bbowman In Java's case, aren't there a bunch of helper methods for dealing with unsigned ints?

design_notes/XDN03 - xlang Type System.md
devhawk: Lets use #53 for further discussion

bbowman:
Will there be a separate design note for the "Foundation" / standard library types? I see some mention of IVector, IPropertySet, etc. here but no mention of Asyncs, Uris, or other types that might reasonably be expected to be a foundational runtime class available to all authors?

design_notes/XDN03 - xlang Type System.md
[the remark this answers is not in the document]
bbowman: There are some helpers like https://docs.oracle.com/javase/8/docs/api/java/lang/Integer.html#divideUnsigned-int-int- but it requires the consumer of the value to understand its unigned and never treat it otherwise. There is also the BigInteger class that can go beyond 64bit but also doesn't "feel" like a primitive does. 

I still want to hear how we handle the casewhich you say "even though various languages may have issue with them" Perhaps there needs to separate design note on language inconsistencies and how that appears in metadata / the authoring experience.

Somewhat tangentially, will there be any annotation support at the metadata level?

devhawk:
@bbowman yes, I need to add a stub XDN for the xlang foundation types. Async and Collections at a minimum. Personally, I also want to have Rx interfaces (IObverver<T>/IObservable<T>) included the foundation.

We need to talk about nullability and boxing before we settle on an approach to property values, which in turn will affect IPropertySet. But that will likely be in foundation as well

URI I'm not sure on yet. It's *NOT* a part of the WinRT foundation contract (though it is in the foundation namespace). On the one hand, most languages have some concept of URI so it would be nice customize the projection of xlang uri. On the other, we don't have any types in the foundation that use URI. I'm wondering if we (or someone) might eventually ship a network xlang component. If so, should URI live there? TO be debated.

design_notes/XDN03 - xlang Type System.md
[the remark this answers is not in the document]
devhawk: @bbowman can you clarify what you mean by annotation support at the metadata level?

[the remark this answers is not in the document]
bbowman: I think its important that there is a communication channel to both xlang component authors and xlang component consumers for information that cannot be perfectly encoded in the API signature either due to deficiencies of the type system (i.e. Uint64 can't be perfectly projected to some languages) or due to some authoring decision (i.e. this parameter can't be null or only values 1-10 will be accepted). 

Obviously comments are one form pf this communication, however, due to the complexities of spoken languages and the fact that comments are traditionally "hands off", a more structured approach in metadata might be useful to "mark" certain attributes on the api such that the projection can inherit them to inform the component consumer and the tooling can add its own to help inform the component author. To my knowledge the WinRT tooling already allows for some "attributes" being added to APIs. I'm just suggesting we have a design note that formalizes what this looks like, how authors can expand it, etc.

Inspiration comes from Java: https://developer.android.com/reference/android/support/annotation/IntRange#annotations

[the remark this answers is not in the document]
bbowman: (Also a Design Note on API comments and tooling is probably worthwhile. That seems like an interesting problem in its own)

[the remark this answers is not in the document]
devhawk: yes, it does sound like an interesting problem space - can you file and issue @bbowman? Or you can start work on an XDN - anyone in the community can [author design notes](https://github.com/Microsoft/xlang/blob/master/design_notes/XDN00%20-%20Design%20Note%20Process.md#2-the-process)

zooba:
URIs have a canonical representation as a string, and everything else is just spans within the string/interpretations of substrings. Similarly for file system paths.

So I'd love to see a rich, transferable representation for both URIs and files/directories, as that moves more well-known validation into the ABI layer. But the underlying transfer type is really just a string, so I really just want a "project as string with extra validation if you know how" type, rather than being forced into matching the same properties between languages (for example, Python's `urlsplit` and `WinHttpCrackUrl` produce different results, plus Python doesn't have a URL "type", so passing through a string that can become a `System.Url` in .NET but be passed back as an `.AbsoluteUrl` string for other languages is fine).

design_notes/XDN03 - xlang Type System.md
```diff
@@ -15,3 +15,252 @@ status: draft
...
+When inheriting from a parameterized interface, the type parameters of the base interface must
+be specified, though they may be specified in terms of the derived interface's type parameters.
+
+A concrete interface is either a non-parameterized interface or a parameterized interface where all
+of the type parameters have been fully specified. Non-parameterized interfaces may inherit from
+concrete interfaces, but they may not inherit from a parameterized interface.
+
+#### Parameterized Interface Examples
+
+For example, xlang's foundation types includes a set of parameterized interfaces to represent
```
BenJKuhn: Since we haven't dug into the collection space and how similar / different xlang is from WinRT, should this have a comment that the collection itnerfaces described here are preliminary / placeholders?

devhawk: I will clarify to indicate this document is not specifying the collection design. I recently added a new [stub XDN](https://github.com/Microsoft/xlang/blob/master/design_notes/XDN07%20-%20xlang%20Foundation%20Types.md#collections) to capture foundation type design (including collections)

design_notes/XDN03 - xlang Type System.md
```diff
@@ -15,3 +15,255 @@ status: draft
...
+- C++ N/A
+- [Java Object.toString](https://docs.oracle.com/javase/7/docs/api/java/lang/Object.html#toString())
+- [JavaScript Object.prototype.toString](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/toString)
+- [.NET Object.ToString](https://docs.microsoft.com/en-us/dotnet/api/system.object.tostring?view=netframework-4.7.2)
+- [Objective-C NSObject description](https://developer.apple.com/documentation/objectivec/1418956-nsobject/1418746-description)
+- [Python __str__ magic method](https://docs.python.org/3/library/stdtypes.html#str)
+
+#### GetHashCode
+
+Returns an Int32 hash code that can be used to insert and identify an object in a hash-based collection.
```
kennykerr: Today language projections produce the hash and can thus produce pointer-sized hash values for efficiency. This seems like a limitation.

design_notes/XDN03 - xlang Type System.md
```diff
@@ -206,12 +206,19 @@ Pre-existing members may not be removed or modified in any way.
...
 
 #### ToString
 
+Returns a string that represents the current object. Typically, this defaults to the object's fully
+namespace-qualified type name, but can be customized. For example, an xlang type that represents a
+[JSON value](http://json.org/) would return the text representation of the JSON value from ToString.
+
 Projection language examples:
 
 - C++ N/A
```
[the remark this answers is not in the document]
devhawk: Also need Python and C#. 

I don't think we need Kotlin and Swift due to how those interop w/ Java and ObjC respectively. 

TS vs. JS is interesting since a TS projection of xlang is probably more "natural and familiar" than a JS one. However, I don't think we can make TS a requirement, so it seems reasonable to add code examples in both TS and JS

design_notes/XDN03 - xlang Type System.md
devhawk: Language projections can't provide value semantics for hash codes or equals - hence the reason for providing types the ability to customize those values. 

I was thinking we should provide a way to indicate (via custom attribute) "this type has reference semantics" as a hint to the language projection NOT to bother calling gethashcode or equals.

kennykerr: As an aside, we really need to support IntPtr. So many WinRT APIs have hacky workarounds for this limitation.

design_notes/XDN03 - xlang Type System.md
```diff
@@ -15,3 +15,255 @@ status: draft
...
+
+#### Parameters
+
+##### Array Parameters
+
+### Properties
+
+### Events
+
+### Intrinsic Operations
```
bbowman: I'm still wondering if we need an intrinsic operation for "associated objects" or some other way that the source object can actually know something about its projection(s). In Project Rome there is a current dilemma where Java Event Handlers require weak references because at the JNI layer a global reference is taken to store the event handler as a native reference in the source object. The java garbage collector can't see through this global reference and no clean up can happen until unsubscribe is called. I believe Cpp\WinRT takes the alternate approach and always attaches the Event Handler as a weak reference but that required us manually take strong references to objects since the projection object we used was somewhat ephemeral anyway. Basically we need to start to think about how we are going to cache and reuse projections and if any of that is best served on the source object itself such that it can cooperate with the projection.

kennykerr: WinRT delegates don't support weak references today.

design_notes/XDN03 - xlang Type System.md
devhawk: Good idea. Can you provide some examples from WinRT so I can review?

design_notes/XDN03 - xlang Type System.md
bbowman: Am I understanding this to further support the need of cooperative ref counting / reference holding between the projection and its source?

design_notes/XDN03 - xlang Type System.md
kennykerr: Event tokens for one. :) This is also the main reason WinRT APIs include "interop" headers that are manually written and shipped in the SDK, for example `windows.graphics.capture.interop.h`. There are some others where API authors smuggle "handles" as UINT64 values. There are legitimate reasons for this limitation though when it comes to marshaling across bitness and nothing inherently wrong with saying "just use UINT64". What you can't say is "just use UINT32".

devhawk: Good point about multiple language projections @BrentRe

devhawk: @zooba I am 100% *against* enabling pointer arithmetic for whatever we end up callign IntPtr

kennykerr: I'm not really arguing one way or the other, but this is how it works today in WinRT and there have been no complaints that I'm aware of, but I'm curious how exactly you would "pass a hash table from one language to another"? In WinRT today we pass *implementations* of hash tables around and the implementation controls the hashing algorithm. This is unaffected by other languages that may happen to be interacting with the data structure via some interface like `IMap<K,V>`.

BrentRe: Take, for example, a map of Point to String. One language projection of Point adds the Point to the map. That language projection calculates the hash code of Point by shifting X left 8 bits and adding it to Y.

Now you pass the collection to a different language projection. Code in that language attempts to retrieve a Map entry. It creates its own projection of a Point with the same X and Y values. But this language projection calculates the hash code differently - crypto hash, or shifting Y instead of X, or whatever. 

Now that language projection will fail to locate the Point in the Map.

kennykerr: Yes, I know that's what you meant but I'm saying that's not how it works. When I call `IMap<K, V>::Find` I don't provide a hash code. I simply provide the key. The implementation (that knows about the implementation) performs the hash operation as needed on the key it receives. The caller doesn't even know whether a hash is needed. The IMap could be implemented using an ordered map algorithm (that doesn't require a hash value).

BrentRe: How does a hypothetical implementation of Map, one that needs a hash code, get a hash code for an opaque object used as a key?

Won't it need to get it from the opaque object?

If so, I'm simply observing that if we decide that a projection provide the hash code, rather than the opaque object itself, then containers holding things originating from different projections will have problems. We can avoid these problems by requiring the opaque object itself to supply the hash code.

In .NET, for example, the Map doesn't compute the hash code for the key being added to the Map. Rather the Map asks the key object for its hash code.

DefaultRyan: An implementation of IMap can do whatever it wants. It can calculate the hash code from the raw ABI pointer for all I care.

I'm having more and more doubts about introducing something fundamental to the type system like hashability, equality, and sort ordering, because the "one size fits all" approach is bound to hit limitations. I think these should be implementation details, specific to the scenario that the container implementer wants to optimize for.

design_notes/XDN03 - xlang Type System.md
devhawk: added link to XDN 07 here

design_notes/XDN03 - xlang Type System.md
devhawk: THis issue is now being tracked by #144

design_notes/XDN03 - xlang Type System.md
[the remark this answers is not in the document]
devhawk: THis is now tracked by issue #130

design_notes/XDN03 - xlang Type System.md
devhawk: FYI, I've added "open issues" sections to several parts of this doc. I've added expanding the types usable as enums. Assuming these issues are still open when I push this PR to master, I'll open issues to track them (and add their IDs to the doc)

design_notes/XDN03 - xlang Type System.md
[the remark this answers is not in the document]
devhawk: I've added a philosophy section with links to all these languages

design_notes/XDN03 - xlang Type System.md
devhawk: It's not really a question of what we think is needed so much as whan the languages we project into will support. Java and .NET's intrinsic hash code type is int32. ObjectiveC and Python use a pointer sized signed integer (i.e. IntPtr). C++ uses std::size_t, which is unsigned and implementation dependent (but typically also pointer sized as I understand).

So the real question is if we want to limit the size to 32 bit because of Java and .NET or make it 64 bit and force those projections do adapt the 64 bit hash to 32 bit. 

I will add an open issue to capture this

design_notes/XDN05 - xlang Binary Interface.md
```diff
@@ -194,15 +200,17 @@ of this document. We need a design note on this.
...
-Most mainstream languages including
-[JavaScript](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/toString),
-[Python](https://docs.python.org/3/library/stdtypes.html#str),
-[.NET](https://docs.microsoft.com/en-us/dotnet/api/system.object.tostring?view=netframework-4.7.2),
-[Java](https://docs.oracle.com/javase/7/docs/api/java/lang/Object.html#toString()) and
-[Objective-C](https://developer.apple.com/documentation/objectivec/1418956-nsobject/1418746-description)
-have a mechanism to retrieve the string representation of an object.
+This information category should only be supported for types that wish to return a custom value for
+a language projection's equivalent of GetHashCode. Language projections may choose to fall back to
+a hash code derived from the object's reference identification (such as a C++ pointer) if HashCode
```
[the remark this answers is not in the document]
devhawk: If we build components in langauges other than C++, it may not be a pointer. Python objects have an id, which happens to be the object's address in memory on the CPython implementation, but that's an implementation detail.

devhawk:
@zooba Agreed on URI and file system path representation. However, this isn't the doc for that. I've started XDN07 for foundational types. It's just stubs right now. There's already a stub for URI, but not file paths. Feel free to add a stub section for file paths to that doc or open an issue

design_notes/XDN05 - xlang Binary Interface.md
```diff
@@ -225,7 +233,10 @@ The property type for ObjectSize is an unsigned 32-bit integer.
 
 ### Object Equality
```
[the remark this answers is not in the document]
devhawk: I think those rules belong in the type system doc, not the ABI doc. Please add comments to that doc to capture what you think is missing.

design_notes/XDN03 - xlang Type System.md
```diff
@@ -15,3 +15,444 @@ status: draft
...
+Int64      | 64 bit signed integer
+UInt8      | 8 bit unsigned integer
+UInt16     | 16 bit unsigned integer
+UInt32     | 32 bit unsigned integer
+UInt64     | 64 bit unsigned integer
+Single     | 32-bit IEEE 754 floating point number
+Double     | 64-bit IEEE 754 floating point number
+Char16     | 16-bit non-numeric value representing a UTF-16 code unit
+Boolean    | 8-bit Boolean value
+String     | immutable sequence of Char16s used to represent text
```
bbowman: Many projections choose to make the String object a full fledged object, which introduces the possibility to have things like strings in heterogeneous "object maps" etc. Would Xlang consider making its string type an xlangobject?

devhawk: I think strings need to be an fundamental type, but I could see a specific language projection customizing how strings are projected so they work better with that language's type system

devhawk:
FYI, all the feedback in this PR has been addressed. I want to start on a new revision of this doc, so I'm going to go ahead and accept this PR. New PR w/ new changes coming soon

PR #121, opened 2019-01-11
Add response file support for command line options

timr1126:
Previous cppwinrt solution used CommandLineToArgvW to parse the response file, however this uses wide characters and is part of a Windows API, hence not cross-platform. Instead, the parsing mimics what crt does to get the command line to argv and argc. The rest of the logic is very similar to the previous implementation.


src/library/cmd_reader.h
```diff
@@ -2,6 +2,8 @@
```
kennykerr: Did you base/reference you implementation on the CRT implementation?

timr1126: Yeah, Ben sent me the original code.

timr1126: Yeah, Ben sent me the original code.

src/library/cmd_reader.h
```diff
@@ -2,6 +2,8 @@
 
 #include "impl/base.h"
 
+using namespace std::experimental::filesystem;
```
kennykerr: Using directives in headers are not good design because it affects everyone including your header and can introduce namespace collisions.

src/library/cmd_reader.h
```diff
@@ -184,5 +165,166 @@ namespace xlang::cmd
         }
 
         std::map<std::string_view, std::vector<std::string>> m_options;
+
+		template<typename O, typename L>
```
kennykerr: #1 coding guideline: be consistent! 😉 

In this case, don't mix tabs and spaces.

src/library/cmd_reader.h
```diff
@@ -184,5 +165,166 @@ namespace xlang::cmd
...
+				auto last{ options.end() };
+				for (auto i = 0; i < argc; i++)
+				{
+					extract_option(argv[i], options, last);
+				}
+			}
+		}
+
+		template <typename Character>
+		static void parse_command_line(
```
kennykerr: Where did you get this implementation?

src/library/cmd_reader.h
```diff
@@ -184,5 +165,166 @@ namespace xlang::cmd
...
+				{
+					throw_invalid("Option '-", svarg, "' is not supported");
+				}
+
+				m_options.try_emplace(last->name);
+			}
+			else if (svarg[0] == '@')
+			{
+				svarg.remove_prefix(1);
+				extract_respose_file(svarg, options);
```
Scottj1s: typo

src/library/cmd_reader.h
```diff
@@ -184,5 +165,166 @@ namespace xlang::cmd
         }
 
         std::map<std::string_view, std::vector<std::string>> m_options;
+
+		template<typename O, typename L>
+		void extract_option(std::string const& sarg, O const& options, L &last)
```
Scottj1s: consistency: L& last

src/library/cmd_reader.h
```diff
@@ -184,5 +165,166 @@ namespace xlang::cmd
         }
 
         std::map<std::string_view, std::vector<std::string>> m_options;
+
+		template<typename O, typename L>
+		void extract_option(std::string const& sarg, O const& options, L &last) 
+		{
+			std::string_view svarg{ sarg };
```
Scottj1s: do you need svarg (or just use sarg directly)?

src/library/cmd_reader.h
kennykerr: Just make the parameter a string_view and the copies.

src/library/cmd_reader.h
kennykerr: This is unnecessary.

src/library/cmd_reader.h
```diff
@@ -184,5 +165,166 @@ namespace xlang::cmd
...
+				throw_invalid("Value '", svarg, "' is not supported");
+			}
+			else
+			{
+				m_options[last->name].push_back(std::string{ svarg });
+			}
+		}
+
+		template<typename O>
+		void extract_respose_file(std::string_view const& arg, O const& options)
```
Scottj1s: typo

Scottj1s:
Tim, can we get some standalone unit test cases added under xlang/src/test/cpp/test/?  Don't need to drive the exe, just the cmd_reader with expected exceptions for failure cases, etc.

src/library/cmd_reader.h
```diff
@@ -184,5 +165,166 @@ namespace xlang::cmd
...
+		{
+			path response_path{ std::string{ arg } };
+			std::wstring extension = response_path.extension();
```
kennykerr: extension returns a path, not a wstring. This won't compile on Clang.

src/library/cmd_reader.h
timr1126: It's parse_command_line from argv_parsing.cpp, slightly modified. Sounds good.

src/library/cmd_reader.h
```diff
@@ -35,31 +35,10 @@ namespace xlang::cmd
...
-                else if (last == options.end())
-                {
-                    throw_invalid("Value '", arg, "' is not supported");
-                }
-                else
-                {
-                    m_options[last->name].push_back(std::string{ arg });
-                }
-            }
+			for (C i = 1; i < argc; ++i)
```
kennykerr: The code still has white space issues all over the place. Find the "show white space" option in your editor and turn it on. Then change your editor settings to use a tab size of 4 and to insert tabs as spaces. Then cut and paste the entire source file and it should reset the white space consistently. Confirm visually.

src/library/cmd_reader.h
```diff
@@ -35,31 +35,10 @@ namespace xlang::cmd
...
+			{
+				extract_option(std::string_view{ std::string{ argv[i] } }, options, last);
```
kennykerr: All of these conversions should be unnecessary as the `extract_option` function takes a `string_view`. You should be able to simply pass the `argv[i]` as the first argument.

src/library/cmd_reader.h
```diff
@@ -184,5 +163,160 @@ namespace xlang::cmd
         }
 
         std::map<std::string_view, std::vector<std::string>> m_options;
+
+		template<typename O, typename L>
+		void extract_option(std::string_view arg, O const& options, L& last)
```
kennykerr: Consistency - use `const&` for `string_view` as that done elsewhere in this file.

src/library/cmd_reader.h
```diff
@@ -184,5 +163,160 @@ namespace xlang::cmd
...
+			{
+				size_t argc = 0;
+				std::vector<std::string> argv;
+
+				parse_command_line(line_buf.data(), argv, &argc);
+
+				auto last{ options.end() };
+				for (auto i = 0; i < argc; i++)
+				{
+					extract_option(std::string_view{ argv[i] }, options, last);
```
kennykerr: This conversion should also be unnecessary.

src/library/cmd_reader.h
```diff
@@ -184,5 +163,160 @@ namespace xlang::cmd
...
+			}
+		}
+
+		template <typename Character>
+		static void parse_command_line(Character* cmdstart, std::vector<std::string>& argv, size_t* argument_count)
+		{
+			*argument_count  = 0;
+
+			Character c;
+			std::string arg("");
```
kennykerr: This constructor argument is unnecessary.

src/library/cmd_reader.h
```diff
@@ -184,5 +163,160 @@ namespace xlang::cmd
...
+			int copy_character; // 1 = copy char to arg
+			unsigned numslash; // num of backslashes seen
```
kennykerr: This comment is redundant. Perhaps just call the variable `backslash_count`.

src/library/cmd_reader.h
```diff
@@ -184,5 +163,160 @@ namespace xlang::cmd
...
+				}
+			}
+		}
+
+		template <typename Character>
+		static void parse_command_line(Character* cmdstart, std::vector<std::string>& argv, size_t* argument_count)
+		{
+			*argument_count  = 0;
+
+			Character c;
```
kennykerr: All the locals in this function should be initialized during declaration.

src/library/cmd_reader.h
```diff
@@ -184,5 +163,160 @@ namespace xlang::cmd
...
+			int copy_character; // 1 = copy char to arg
+			unsigned numslash; // num of backslashes seen
+			bool in_quotes;
+			bool first_arg;
+
+			Character* p = cmdstart;
+			in_quotes = false;
+			first_arg = true;
+
+			// Loop on each argument
```
kennykerr: Please remove comments that don't substantially make the code more understandable.

src/library/cmd_reader.h
```diff
@@ -184,5 +163,160 @@ namespace xlang::cmd
...
+					{
+						arg.push_back(*p);
+					}
+
+					++p;
+				}
+
+				first_arg = false;
+			}
+		}
```
kennykerr: This really needs some unit tests. 😉

src/library/cmd_reader.h
timr1126: remove_prefix doesn't work with const& set. Should I change other methods to not be const& for consistency?

src/test/cpp/test/CmdReader.cpp
```diff
@@ -0,0 +1,137 @@
+#include "pch.h"
+#include "winrt/Windows.Foundation.h"
+#include "cmd_reader.h"
+#include <fstream>
+
+using namespace xlang::cmd;
+
+TEST_CASE("CmdReader")
+{
+    const std::vector<option> options
```
Scottj1s: static

src/library/cmd_reader.h
```diff
@@ -184,5 +163,153 @@ namespace xlang::cmd
...
+            }
+        }
+
+        template<typename O>
+        void extract_response_file(std::string_view const& arg, O const& options)
+        {
+            std::experimental::filesystem::path response_path{ std::string{ arg } };
+            std::string extension = response_path.extension().generic_string();
+            std::transform(extension.begin(), extension.end(), extension.begin(),
+                [](unsigned char c) { return static_cast<unsigned char>(::tolower(c)); });
```
Scottj1s: can make c 'auto'

src/test/cpp/test/CmdReader.cpp
```diff
@@ -0,0 +1,137 @@
...
+        REQUIRE_FALSE(args.exists("filter"));
+        REQUIRE_FALSE(args.exists("name"));
+        REQUIRE_FALSE(args.exists("verbose"));
+    }
+
+    // response file #1: filename no quotes
+    {
+        const char* argv[] = { "progname", "@respfile.txt" };
+        const size_t argc = 2;
+        std::ofstream resp_file;
```
Scottj1s: consider making an raii type here to consolidate common logic and delete file in presence of an exception

src/test/cpp/test/CmdReader.cpp
```diff
@@ -0,0 +1,137 @@
...
+        { "component", 0, 1 },
+        { "filter", 0 },
+        { "name", 0, 1 },
+        { "verbose", 0, 0 },
+    };
+
+    // input and output
+    {
+        const char* argv[] = { "progname", "-in", "example_file.in", "-out", "example_file.out" };
+        const size_t argc = 5;
```
Scottj1s: all argc values can be = _countof(argv)

src/test/cpp/test/CmdReader.cpp
kennykerr: Might be simpler and more maintainable just to have an actual file in the test project containing the test arguments and not bother with generating and then deleting a file.

src/test/cpp/test/CmdReader.cpp
```diff
@@ -0,0 +1,137 @@
...
+    }
+
+    // response file #3: filename with quote within name
+    {
+        const char* argv[] = { "progname", "@respfile.txt" };
+        const size_t argc = 2;
+        std::ofstream resp_file;
+
+        resp_file.open("respfile.txt");
+        resp_file << "-in example\\\"file.in -out example\\\"file.out";
```
Scottj1s: consider raw string literals

src/test/cpp/test/CmdReader.cpp
```diff
@@ -0,0 +1,137 @@
+#include "pch.h"
+#include "winrt/Windows.Foundation.h"
+#include "cmd_reader.h"
+#include <fstream>
+
+using namespace xlang::cmd;
+
+TEST_CASE("CmdReader")
```
devhawk: @kennykerr why are test names PascalCased instead of snake_cased?

src/library/cmd_reader.h
```diff
@@ -184,5 +163,153 @@ namespace xlang::cmd
...
+            if (is_directory(response_path) || extension == ".winmd")
```
Scottj1s: please add comment that this line is intended to help diagnose misuse of the '@' prefix (e.g., when supplying metadata files)

src/test/cpp/test/CmdReader.cpp
Scottj1s: I actually like the generation code - everything in one place

src/test/cpp/test/CmdReader.cpp
kennykerr: Originally so that I could easily distinguish between library types and test code, but that's not a compelling reason. I can switch it to make it more consistent overall.

devhawk: I actually think that's a compelling reason and switching it seems like a bunch of busy work when we have more important things to do

kennykerr:
Thanks Tim!

PR #122, opened 2019-01-11
C++/WinRT restore SDK support

kennykerr:
This update restores support for referring a specific SDK rather than just a set of winmd files. All of the following now works as before:

`-in sdk`
`-in sdk+`
`-in 10.0.18314.0`
`-in 10.0.18314.0+`

This update also includes a few bug fixes and tweaks to bring the xlang version on par with the production version of the cppwinrt compiler. I've also ported a few more unit tests.


devhawk:
@kennykerr can you delete the comments that are not relevant to this PR?

src/library/cmd_reader.h
```diff
@@ -157,19 +166,305 @@ namespace xlang::cmd
...
+            {
+                if (handle)
+                {
+                    RegCloseKey(handle);
+                }
+            }
+        };
+
+        template <typename T>
+        struct com_ptr
```
Scottj1s: any reason we can use base_com_ptr.h?

src/library/cmd_reader.h
```diff
@@ -157,19 +166,305 @@ namespace xlang::cmd
...
         }
 
+        auto files(std::string_view const& name) const
+        {
+            return files(name, [](auto&&) {return true; });
+        }
+
     private:
 
+#if XLANG_PLATFORM_WINDOWS
```
Scottj1s: wondering if we should factor out windows support into a separate header...

src/library/cmd_reader.h
```diff
@@ -157,19 +166,305 @@ namespace xlang::cmd
...
+            if (result < 0)
+            {
+                throw_invalid("Could not read the Windows SDK's Platform.xml");
+            }
+        }
+
+        static void add_files_from_xml(
+            std::set<std::string>& files,
+            std::string const& sdk_version,
+            std::experimental::filesystem::path const& xml_path,
```
Scottj1s: should we be using std::filesystem::path now?

kennykerr: Yes, I'd love to. I thought Harry mentioned some reason why that didn't work on the Ubuntu build at present. I can check whether this has been fixed.

src/library/cmd_reader.h
```diff
@@ -157,19 +166,305 @@ namespace xlang::cmd
...
+            HKEY key;
+
+            if (0 != RegOpenKeyEx(
+                HKEY_LOCAL_MACHINE,
+                "SOFTWARE\\Microsoft\\Windows Kits\\Installed Roots",
+                0,
+                KEY_READ,
+                &key))
+            {
+                throw_invalid("Could not find the Windows SDK");
```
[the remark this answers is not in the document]
kennykerr: 👍

src/library/impl/meta_reader/database.h
```diff
@@ -30,6 +30,71 @@ namespace xlang::meta::reader
         database(database&&) = delete;
         database& operator=(database&&) = delete;
 
+        static bool is_database(std::string_view const& path)
```
Scottj1s: consider factoring this and the duplicate ctor logic into a helper

src/library/cmd_reader.h
devhawk: Clang 6.0's libc++ doesn't include filesystem in std namespace. https://github.com/llvm-mirror/libcxx/tree/release_60/include

kennykerr: Clang 7 has been out for almost 6 months. 😉

src/tool/cpp/cppwinrt/main.cpp
```diff
@@ -47,8 +47,10 @@ namespace xlang
 
         settings.verbose = args.exists("verbose");
         settings.root = args.value("root", "winrt");
-        settings.input = args.files("input");
-        settings.reference = args.files("reference");
+
+        settings.input = args.files("input", database::is_database);
```
Scottj1s: curious why a deep inspection of the PE header is necessary, versus just an extension check?

src/library/cmd_reader.h
```diff
@@ -157,19 +166,305 @@ namespace xlang::cmd
...
+
+            std::wstring root((path_size / sizeof(wchar_t)) - 1, L'?');
+
+            RegQueryValueExW(
+                key.handle,
+                L"KitsRoot10",
+                nullptr,
+                nullptr,
+                reinterpret_cast<BYTE*>(root.data()),
+                &path_size);
```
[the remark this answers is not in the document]
Scottj1s: fair enough

src/tool/cpp/cppwinrt/main.cpp
kennykerr: Previously, we just checked the extension but I was concerned about this being used with xlang (non-WinRT) projects where there has been talk of using a different file extension.

PR #123, opened 2019-01-14
Fix C++/WinRT override class code gen

kennykerr:
Previously, C++/WinRT would not generate the appropriate using declarations to disambiguate overloads of required interfaces. This is now fixed and the resulting user override can now call those base members. 

Beyond that, this update includes some minor code gen formatting tweaks as well as a workaround for another MIDLRT regression related to attributes.


kennykerr:
Completing this PR as it resolves a build break (caused by the MIDL regression) that's preventing me from completing the other approved PR... 🙄

PR #126, opened 2019-01-15
C++/WinRT workaround for OS compiler

kennykerr:
The Windows OS is built with a fork of the Microsoft C++ compiler. The current version of this compiler is roughly comparable to Visual C++ 15.8, but includes a critical constexpr bug that has since been fixed in commercial releases. This workaround allows C++/WinRT to be ingested into the OS repo and build successfully with the this compiler.


kennykerr:
Closing to collect a few more workarounds for the OS build.

PR #150, opened 2019-01-28
C++/WinRT help diagnose stack allocations

kennykerr:
This update contains a few minor tweaks to aid component developers. Some code gen and formatting fixes as well as a debug-only assertion to help detect stack allocations. 

Since the projected and implementation class names are (by default) the same and only differ by namespace, it is possible to mistake the one for the other and accidentally create an implementation on the stack rather than using the `make<T>` helpers. WRL avoids this by playing tricks with `new` and `delete`, which is basically guaranteed to cause other problems and is why C++/WinRT intentionally does not mess with new and delete.

Still, this can be hard to diagnose in some cases as the object may be destroyed while outstanding references are still in flight. An assertion will now pick this up for debug builds. The `is_stack_object` function was kindly donated by @jamesmcnellis.

### DIFF
--- a/src/tool/cpp/cppwinrt/strings/base_implements.h
+++ b/src/tool/cpp/cppwinrt/strings/base_implements.h
@@ -1226,7 +1226,7 @@ namespace winrt
 #ifdef _DEBUG
         implements() noexcept
         {
-            WINRT_ASSERT(!is_stack_address(this));
+            WINRT_ASSERT(!is_stack_object());
         }
 #endif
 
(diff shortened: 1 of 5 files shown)


src/tool/cpp/cppwinrt/strings/base_implements.h
```diff
@@ -1223,6 +1223,13 @@ namespace winrt
         using implements_type = implements;
         using IInspectable = Windows::Foundation::IInspectable;
 
+#ifdef _DEBUG
+        implements() noexcept
+        {
+            WINRT_ASSERT(!is_stack_address(this));
```
JamesMcNellis: My one comment is that this check is expensive (three syscalls--one for each VirtualQuery).  I don't know whether it is _too_ expensive--that is for you to say.

src/tool/cpp/cppwinrt/strings/base_implements.h
```diff
@@ -1284,6 +1291,44 @@ namespace winrt
...
+                size_t RegionSize;
+                uint32_t State;
+                uint32_t Protect;
+                uint32_t Type;
+            };
+
+            static constexpr uint32_t page_guard{ 0x100 };
+
+            memory_basic_information info{};
+            WINRT_VirtualQuery(address, &info, sizeof(info));
```
JamesMcNellis: One thing I omitted from my proof of concept:  You should check the return value of VirtualQuery and make sure that it is equal to sizeof(info) before using the info structure.

src/tool/cpp/cppwinrt/strings/base_implements.h
```diff
@@ -1284,6 +1291,44 @@ namespace winrt
...
+            }
+
+            WINRT_VirtualQuery(info.AllocationBase, &info, sizeof(info));
+
+            if ((info.Protect& page_guard) == page_guard)
+            {
+                return true;
+            }
+
+            void const* const potential_page = static_cast<char const*>(info.AllocationBase) + info.RegionSize;
```
JamesMcNellis: `potential_guard_page`?  (It's definitely a page--no potentiality about that :-))

src/tool/cpp/cppwinrt/strings/base_implements.h
kennykerr: Yes, a little concerning but `winrt::implements` tends to be used for course grained allocations. Rarely thousands of small objects (delegates don't use winrt::implements). Plus, this only applies to debug builds. If there are complaints I can offer a macro to opt this out of debug builds.

src/tool/cpp/cppwinrt/strings/base_implements.h
kennykerr: 👍

src/tool/cpp/cppwinrt/strings/base_implements.h
kennykerr: A bit too much brevity there on my part. 😉

PR #153, opened 2019-02-01
Python/WinRT: Array Parameters

devhawk:
This update adds support for calling WinRT methods with array parameters from Python

> Note, there is currently a GC issue in the Python/WinRT projection that causes a crash in the GetMany tests in testjson.py. An `items.clear()` call has been added to those tests to work around this issue. I'm currently planning on not fixing this issue as part of this PR. Rather, I will tackle ref counting cleanup as a seperate PR. 

Generated code available: https://github.com/devhawk/pywinrt-output/pull/4


src/test/python/testjson.py
```diff
@@ -158,6 +158,34 @@ def test_invalid_param_count_static(self):
...
+
+        self.assertEqual(count, 3)
+        self.assertEqual(len(items), 3)
+
+        self.assertFalse(items[0].GetBoolean())
+        self.assertEqual(items[1].GetNumber(), 42)
+        self.assertEqual(items[2].ValueType, 0)
+
+        # TODO: remove clear call after resolving leak issue
+        items.clear()
```
devhawk: As per the initial PR comment, this call is here to work around a crashing bug. I'm planning on reworking my ref counting code by moving to a smart pointer (a la https://github.com/zooba/pyptr) in a seperate PR

PR #154, opened 2019-02-02
Move C++/WinRT coroutine helpers to dedicated header and add support for DispatcherQueue to resume_foreground

kennykerr:
In previous commits I have steadily reduced the size of `winrt/base.h` by moving a lot of the previously hand-rolled base types (like all of the generics) into their respective namespace headers now that the winmd parser is able to handle that level of type information and we can simply generate those types. This update takes the next step by moving all of the coroutine helpers into a dedicated header. 

Apart from being more maintainable and supportable, it finally means that coroutine helpers like `resume_foreground` no longer have to be tacked onto the end of a specific namespace header and can instead more naturally include their dependencies. This further allows `resume_foreground` to support not only resuming on a given `Windows.UI.Core.CoreDispatcher` but it can now also support resuming on a given `Windows.System.DispatcherQueue`. Previously, only one could be supported but not both since the definition could only reside in one namespace.

The `resume_foreground` helpers are now also decorated with `[[nodiscard]]` thus improving their usability as I had previously done for the thread pool helpers.

The rest of the churn in this PR is mostly just file movement/renames to make it all a bit more consistent and approachable.

It's also become increasingly impractical to test C++/WinRT without access to the broader Windows API. We may need to take a dependency on the Windows SDK just for testing but for now I've just updated the test project to build with local metadata while we resolve that issue separately.

Here's an example of the DispatcherQueue support:

```C++
fire_and_forget Async(DispatcherQueueController controller)
{
    bool queued = co_await resume_foreground(controller.DispatcherQueue());
    assert(queued);

    co_await controller.ShutdownQueueAsync();

    queued = co_await resume_foreground(controller.DispatcherQueue());
    assert(!queued);
}
```


src/tool/cpp/cppwinrt/strings/base_coroutine.h
```diff
@@ -0,0 +1,414 @@
...
+            return true;
+        }
+
+        void await_suspend(std::experimental::coroutine_handle<>) const noexcept
+        {
+        }
+
+        cancellation_token<Promise> await_resume() const noexcept
+        {
+            return*this;
```
DefaultRyan: Nit: space

kennykerr:
@devhawk Hey Harry, do you know why the Windows build might be failing? From the logs it appears as if it cannot find the WinMetadata folder. Any ideas?

DefaultRyan:
I think I saw something about the local WinMetadata living in a different folder now. Can't remember where though.

Get Outlook for Android<https://aka.ms/ghei36>

________________________________

kennykerr:
@devhawk I tried a few different things to get the remote build going, but no luck.

devhawk:
@kennykerr I won't have time to look until monday, but based on the error log it look like the checked in copy of Windows.Foundation.winmd is corrupt. I've seen it happen where git will try and do the auto CR/LF transformation on binary files. Our [.gitattributes](https://www.git-scm.com/docs/gitattributes) only excludes .png and .onnx from this transformation. 

2019-02-02T22:45:44.1852130Z midlrt : error MIDL1001 : [msg]cannot open input file [context]Error opening or processing WinMD c:\__w\1\s\src\test\cpp\metadata\windows.foundation.winmd. HRESULT 0x80131124.
2019-02-02T22:45:44.1879699Z C:\__w\1\s\src\test\cpp\component.idl(1) : error MIDL5157 : [msg]An invalid winmd file was specified for import [context]: C:\__w\1\s\src\test\cpp\metadata\Windows.Foundation.winmd

devhawk:
@DefaultRyan I can't believe the WinMetadata file moved. Besides, the build container image is on windows server core 1803.

kennykerr:
The git corruption thing is plausible. I'll give that a try - thanks.

Still doesn't explain why it failed to find WinMetadata - the change to that folder happened in 19H1 so it won't affect the build server.

kennykerr:
Victory! Thanks Harry.

PR #164, opened 2019-02-08
Rounded out command line switches and usage for compat with old cppwinrt.exe

Scottj1s:


src/tool/cpp/cppwinrt/code_writers.h
```diff
@@ -2,20 +2,28 @@
 
 namespace xlang
 {
-    static void write_license(writer& w)
+    static void write_prolog(writer& w)
```
kennykerr: Maybe "preamble" rather than "prolog", not to be confused with https://en.wikipedia.org/wiki/Prolog 😉

Scottj1s: yes good point

src/tool/cpp/cppwinrt/main.cpp
```diff
@@ -13,37 +14,105 @@ namespace xlang
...
+        { "opt", 0, 0, {}, "Generate component projection with unified construction support" },
+        { "help", 0, cmd::option::no_max, {}, "Show detailed help with examples" },
+        { "root", 0, 1 }, // "Root folder name for projection headers (defaults to winrt)
+        { "filter" }, // One or more prefixes to include in input (same as -include)
+        { "lib", 0, 1 }, // Specify library prefix (defaults to winrt)
+        { "license", 0, 0 }, // Generate license comment
+        { "brackets", 0, 0 }, // Use angle brackets for #includes (defaults to quotes)
+    };
+
+    static void printUsage(writer& w, bool details)
```
kennykerr: The time you're spending with Java is rubbing off on you. :wink:

`print_usage`

Scottj1s: yes, it is!

src/tool/cpp/cppwinrt/main.cpp
```diff
@@ -13,37 +14,105 @@ namespace xlang
...
+        time_t t{ time(nullptr) };
+        tm tm{};
+        localtime_s(&tm, &t);
+
+        w.write(format, 
+            XLANG_VERSION_STRING, 
+            1900 + tm.tm_year, 
+            bind_each(printOption, options));
+
+        if (details)
```
kennykerr: I don't think any of this needs to be in the tool. We can point to the docs page. Beyond that it seems impractical to try to make the tool self-documenting without substantially more effort. I think a list of options is sufficient and perhaps a link to the docs. Then we can focus on having quality detailed usage information in one place (online).

src/tool/cpp/cppwinrt/main.cpp
```diff
@@ -13,37 +14,98 @@ namespace xlang
...
+Options:
+
+%  ^@<path>             Response file containing command line options
+
+Where <spec> is one or more of:
+
+  path                Path to winmd file or recursively scanned folder
+  local               Local ^%WinDir^%\System32\WinMetadata folder
+  sdk[+]              Current version of Windows SDK [with extensions]
+  10.0.12345.0[+]     Specific version of Windows SDK [with extensions]
```
kennykerr: I like 10.0.12345.0 rather than a specific version. 👍

src/tool/cpp/cppwinrt/main.cpp
```diff
@@ -13,37 +14,98 @@ namespace xlang
...
+        { "prefix", 0, cmd::option::no_max, {}, "Use dotted namespace convention for component files (defaults to folders)" },
+        { "pch", 0, 1, "<name>", "Specify name of precompiled header file (defaults to pch.h)" },
+        { "include", 0, cmd::option::no_max, "<prefix>", "One or more prefixes to include in input" },
+        { "exclude", 0, cmd::option::no_max, "<prefix>", "One or more prefixes to exclude from input" },
+        { "base", 0, 0, {}, "Generate base.h unconditionally" },
+        { "opt", 0, 0, {}, "Generate component projection with unified construction support" },
+        { "help", 0, cmd::option::no_max, {}, "Show detailed help with examples" },
+        { "root", 0, 1 }, // "Root folder name for projection headers (defaults to winrt)
+        { "filter" }, // One or more prefixes to include in input (same as -include)
+        { "lib", 0, 1 }, // Specify library prefix (defaults to winrt)
```
kennykerr: Remove "lib" - that was not in the internal version and doesn't work.

src/tool/cpp/cppwinrt/main.cpp
```diff
@@ -13,37 +14,98 @@ namespace xlang
...
+        { "verbose", 0, 0, {}, "Show detailed progress information" },
+        { "overwrite", 0, 0, {}, "Overwrite generated component files" },
+        { "prefix", 0, cmd::option::no_max, {}, "Use dotted namespace convention for component files (defaults to folders)" },
+        { "pch", 0, 1, "<name>", "Specify name of precompiled header file (defaults to pch.h)" },
+        { "include", 0, cmd::option::no_max, "<prefix>", "One or more prefixes to include in input" },
+        { "exclude", 0, cmd::option::no_max, "<prefix>", "One or more prefixes to exclude from input" },
+        { "base", 0, 0, {}, "Generate base.h unconditionally" },
+        { "opt", 0, 0, {}, "Generate component projection with unified construction support" },
+        { "help", 0, cmd::option::no_max, {}, "Show detailed help with examples" },
+        { "root", 0, 1 }, // "Root folder name for projection headers (defaults to winrt)
```
kennykerr: Remove "root" - that was not in the internal version and doesn't work.

src/tool/cpp/cppwinrt/main.cpp
Scottj1s: let's rip out the settings.component_lib stuff in a separate PR.  for now it's undocumented.

src/tool/cpp/cppwinrt/main.cpp
Scottj1s: also undocumented, so we can remove at leisure

src/tool/cpp/cppwinrt/main.cpp
```diff
@@ -15,33 +16,70 @@ namespace xlang
 
     struct usage_exception {};
 
-    static void process_args(int const argc, char** argv)
+    static std::vector<cmd::option> options
```
DefaultRyan: I've always wondered, why is this a static std::vector, instead of a constexpr array?

Scottj1s: great question

kennykerr: Because for some weird reason this doesn't work:

```
std::array<cmd::option, 18> options
{
    { "input", 0 },
    { "reference", 0 },
    { "output", 0, 1 },
    { "component", 0, 1 },
````

Instead, when I switch to using an array I need to do this:

```
std::array<cmd::option, 18> options
{
    cmd::option{ "input", 0 },
    cmd::option{ "reference", 0 },
    cmd::option{ "output", 0, 1 },
    cmd::option{ "component", 0, 1 },
```

Not a high priority so haven't dug into why exactly. Uniform initialization and all that... Feel free to fix whatever's wrong.

Comment (DefaultRyan) [reply src/tool/cpp/cppwinrt/main.cpp]: Instead of std::array, the terse syntax should work with plain arrays:
```c++
constexpr cmd::option[] options
{
    { "input", 0},
    { "reference", 0},
```

Scottj1s:
No idea what happened with the PR - committing and will address any build breaks later.

src/tool/cpp/cppwinrt/main.cpp
kennykerr: Pity.

PR #165, opened 2019-02-08
made options a constexpr array, removed unused lib & root

Scottj1s:


src/tool/cpp/cppwinrt/component_writers.h
```diff
@@ -138,7 +138,7 @@ namespace xlang
     {
         w.write_root_include("base");
         auto format = R"(%
-bool WINRT_CALL %_can_unload_now() noexcept
+bool WINRT_CALL winrt_can_unload_now() noexcept
```
kennykerr: No, this is wrong. It's very important that this preamble to these functions is still configurable.

src/tool/cpp/cppwinrt/component_writers.h
```diff
@@ -149,7 +149,7 @@ bool WINRT_CALL %_can_unload_now() noexcept
     return true;
 }
 
-void* WINRT_CALL %_get_activation_factory(std::wstring_view const& name)
+void* WINRT_CALL winrt_get_activation_factory(std::wstring_view const& name)
```
kennykerr: Same here

src/tool/cpp/cppwinrt/component_writers.h
```diff
@@ -168,7 +168,7 @@ int32_t WINRT_CALL WINRT_CanUnloadNow() noexcept
     }
 #endif
 
-    return %_can_unload_now() ? 0 : 1;
+    return winrt_can_unload_now() ? 0 : 1;
```
kennykerr: And here

src/tool/cpp/cppwinrt/component_writers.h
```diff
@@ -179,7 +179,7 @@ int32_t WINRT_CALL WINRT_GetActivationFactory(void* classId, void** factory) noe
         wchar_t const* const buffer = WINRT_WindowsGetStringRawBuffer(classId, &length);
         std::wstring_view const name{ buffer, length };
 
-        *factory = %_get_activation_factory(name);
+        *factory = winrt_get_activation_factory(name);
```
kennykerr: And here

src/tool/cpp/cppwinrt/main.cpp
```diff
@@ -32,9 +32,7 @@ namespace xlang
         { "base", 0, 0, {}, "Generate base.h unconditionally" },
         { "opt", 0, 0, {}, "Generate component projection with unified construction support" },
         { "help", 0, cmd::option::no_max, {}, "Show detailed help with examples" },
-        { "root", 0, 1 }, // "Root folder name for projection headers (defaults to winrt)
         { "filter" }, // One or more prefixes to include in input (same as -include)
-        { "lib", 0, 1 }, // Specify library prefix (defaults to winrt)
```
kennykerr: I'll fix this up. I forgot the lib prefix applies to those module exports. It's critical that they can be changed to allow multiple libs to be combined into a single DLL.

PR #166, opened 2019-02-09
Restore -lib option I mistakenly told Scott to remove :)

kennykerr:
The -lib option is critical to supporting larger projects (like the OS) where multiple implementation projects are frequently combined as individual libs into a single DLL. This option makes it far easier to combine their exports. 

This update also includes a few other minor code gen improvements I had queued up in my branch.


src/tool/cpp/cppwinrt/component_writers.h
```diff
@@ -198,7 +198,11 @@ int32_t WINRT_CALL WINRT_GetActivationFactory(void* classId, void** factory) noe
 
         w.write(format,
             bind_each<write_component_include>(classes),
-            bind_each<write_component_activation>(classes));
+            settings.component_lib,
```
jonwis: Out of curiosity, why is this line repeated? Or am I just not seeing the difference?

kennykerr: You mean the `settings.component_lib` argument? Because the same value is needed multiple times by the format string.

Issue #170, opened 2019-02-13
Provide mechanism to pass build pipeline version down to cmake


devhawk:
Currently, src\tool\cpp\cppwinrt\version.h is updated by our internal build system. We should be generating this file as part of the cmake compile process, as demonstrated in the [CMake tutorial](https://cmake.org/cmake-tutorial/#s1)

Scottj1s:
Rather than using version.h as the version source for other projects, the version number should be explicitly passed to cmake and propagated down.

devhawk:
So we would pass XLANG_VERSION_STRING (or it's equivalent) into cl as a command line parameter for cpp/cppwinrt project?

Scottj1s:
We still need to keep the cpp\cppwinrt\version.h around (it's included in the projection).  But other projects (natvis, nuget, vsix) could do their own versioning thing and not depend on cppwinrt.

devhawk:
It sounds like we can seperate this into to parts.

1. Use a cmake variable to track cppwinrt version number. We'd generate version.h with that and use it for package projects that need the version information
2. Enable passing in the cppwinrt version number from the build pipeline (with a resonable default for local build)

Scottj1s:
Fixed by #172

PR #172, opened 2019-02-14
Added build version property, to be passed from build definition

Scottj1s:
(defaulting to 1.0.0 otherwise)

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -3,6 +3,10 @@ cmake_minimum_required(VERSION 3.9)
 set(XLANG_SCRIPTS_PATH "${CMAKE_CURRENT_SOURCE_DIR}/scripts")
 set(XLANG_LIBRARY_PATH "${CMAKE_CURRENT_SOURCE_DIR}/library")
 
+if(NOT DEFINED XLANG_BUILD_VERSION)
+set(XLANG_BUILD_VERSION "1.0.0")
+endif()
+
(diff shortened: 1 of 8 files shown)


src/CMakeLists.txt
```diff
@@ -3,6 +3,10 @@ cmake_minimum_required(VERSION 3.9)
 set(XLANG_SCRIPTS_PATH "${CMAKE_CURRENT_SOURCE_DIR}/scripts")
 set(XLANG_LIBRARY_PATH "${CMAKE_CURRENT_SOURCE_DIR}/library")
 
+if(NOT DEFINED XLANG_BUILD_VERSION)
```
Scottj1s: keeping this distinct from the other issue about xlang vs cppwinrt version

src/tool/cpp/cppwinrt/CMakeLists.txt
```diff
@@ -15,3 +15,5 @@ else()
     target_link_libraries(cppwinrt c++ c++abi c++experimental)
     target_link_libraries(cppwinrt -lpthread)
 endif()
+
+string(APPEND CMAKE_CXX_FLAGS " -D \"XLANG_VERSION_STRING=\\\"${XLANG_BUILD_VERSION}\\\"\"")
```
Scottj1s: sadly, all the escapes appear necessary for string macro definitions

Scottj1s:
This addresses #170

PR #174, opened 2019-02-14
Defaultryan/string convert

DefaultRyan:
Now we have a platform-independent UTF-8 to UTF-16 converter. It passes all the unit tests, but may still have some room for optimization.

### DIFF
--- a/src/platform/common_string_convert.cpp
+++ b/src/platform/common_string_convert.cpp
@@ -6,7 +6,7 @@
 namespace xlang::impl::convert
 {
     using utf8_worker_t = std::conditional_t<std::is_signed_v<xlang_char8>, uint8_t, xlang_char8>;
-    inline auto to_worker(xlang_char8 const* arg)
+    inline auto to_worker(xlang_char8 const* arg) noexcept
     {
         if constexpr (std::is_same_v<xlang_char8, utf8_worker_t>)
         {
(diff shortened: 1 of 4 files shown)


src/platform/common_string_convert.cpp
```diff
@@ -1,17 +1,280 @@
 #include "pal_internal.h"
 #include "string_convert.h"
 #include "pal_error.h"
+#include "string_traits.h"
+
+namespace xlang::impl::convert
+{
+    using utf8_worker_t = std::conditional_t<std::is_signed_v<xlang_char8>, uint8_t, xlang_char8>;
+    inline auto to_worker(xlang_char8 const* arg)
```
kennykerr: noexcept all the things...

src/platform/string_traits.h
```diff
@@ -2,24 +2,20 @@
...
 
-template <>
-struct alternate_type<char16_t>
-{
-    using result_type = xlang_char8;
-};
+    template <>
+    struct alternate_type<xlang_char8>
+    {
+        using result_type = char16_t;
```
devhawk: nit, isn't standard C++ naming convention to simply call this "type" not "result_type"?

devhawk:
FYI, I have ubuntu test runs [ready to go](https://dev.azure.com/msft-xlang/public/_build/results?buildId=595&view=ms.vss-test-web.build-test-results-tab) in AzDO as soon as this PR is checked in.

Issue #180, opened 2019-02-19
xlang::impl::convert_string returns 0 on non windows


devhawk:
Windows verison of convert_string returns result of `WideCharToMultiByte`/`MultiByteToWideChar`, which is described as "number of bytes/characters written to the buffer". Common version of convert_string calls the templated do_conversion function, that ends with the following:

``` cpp
        XLANG_ASSERT(output_cursor == output_end);
        return static_cast<uint32_t>(output_cursor - output_end);
```

Since we assert output_cursor == output_end, the return value is always zero.

PR #181, opened 2019-02-19
xplat activation

devhawk:
This PR enables activation on non-windows platforms. 

In order to test activation, this PR adds a new test component that is built w/o cppwinrt, so it can be compiled on non-windows. 

Both the test_platform app and the pal have RPATH set to $ORIGIN so it can load libraries from the current path. 

Note, there's an issue in the new common string conversion code #180 which common activation depends on. I've worked around it for now.


src/test/platform/activation.cpp
```diff
@@ -1,29 +1,29 @@
...
     xlang_string_header str_header{};
     xlang_string str{};
 
     result = xlang_create_string_reference_utf16(class_name.data(), static_cast<uint32_t>(class_name.size()), &str_header, &str);
     REQUIRE(result == nullptr);
 
-    winrt::Windows::Foundation::IUnknown factory{};
-    xlang_guid iid = to_guid(winrt::guid_of<winrt::Windows::Foundation::IActivationFactory>());
-    result = xlang_get_activation_factory(str, iid, winrt::put_abi(factory));
+    xlang_unknown* factory{};
```
Scottj1s: >xlang_unknown [](start = 4, length = 13)

com_ptr (or something like it, which shouldn't require XLANG_PLATFORM_WINDOWS)

src/CMakeLists.txt
```diff
@@ -71,6 +71,19 @@ function(CONSUME_PAL target)
     target_link_libraries(${target} pal)
 endfunction()
 
+# set test_platform linker flags to look for .so files in the current directory 
+# adapted from https://serverfault.com/a/926072
+function(RPATH_ORIGIN target)
+    if (NOT WIN32)
```
Scottj1s: this would pickup MacOS?  do we have a more specific platform filter?

src/test/abi_component/widget.cpp
```diff
@@ -0,0 +1,145 @@
...
+#include <pal_error.h>
+#include <string_view>
+
+struct widget : iwidget
+{
+    int32_t XLANG_CALL QueryInterface(xlang_guid const& id, void** object) noexcept final
+    {
+        if (id == xlang_unknown_guid)
+        {
+            static_assert(std::is_base_of_v<xlang_unknown, iwidget>, "Can only combine these two cases is this is true.");
```
Scottj1s: typo (is this)

src/test/abi_component/widget.cpp
```diff
@@ -0,0 +1,145 @@
...
+        xlang::impl::atomic_ref_count m_count;
+};
+
+struct widget_factory : iwidget_factory
+{
+    int32_t XLANG_CALL QueryInterface(xlang_guid const& id, void** object) noexcept final
+    {
+        if (id == xlang_unknown_guid)
+        {
+            static_assert(std::is_base_of_v<xlang_unknown, iwidget_factory>, "Can only combine these two cases is this is true.");
```
Scottj1s: typo again

src/CMakeLists.txt
devhawk: I'm not sure what we'll need for MacOS, but I figure we will cross that bridge when we get there.

src/test/platform/activation.cpp
```diff
@@ -1,29 +1,29 @@
...
+    xlang_guid iid = { 0x00000000,0x0000,0x0000,{ 0xC0,0x00,0x00,0x00,0x00,0x00,0x00,0x46 } };
```
Scottj1s: why not use xlang_unknown_guid from pal.h?

src/test/platform/activation.cpp
devhawk: Agree 100%...when we have a xplat implementation of com_ptr.

src/test/abi_component/widget.cpp
```diff
@@ -0,0 +1,145 @@
+#include "iwidget.h"
+#include <pal_error.h>
+#include <string_view>
+
+struct widget : iwidget
+{
+    int32_t XLANG_CALL QueryInterface(xlang_guid const& id, void** object) noexcept final
```
Scottj1s: why is there a mix of noexcept and XLANG_NOEXCEPT?

src/test/platform/activation.cpp
devhawk: added #185 to track

src/test/abi_component/widget.cpp
```diff
@@ -0,0 +1,145 @@
...
+        else if (id == iwidget_guid)
+        {
+            *object = static_cast<iwidget*>(this);
+        }
+        else
+        {
+            *object = nullptr;
+            return xlang_error_no_interface;
+        }
+        AddRef();
```
Scottj1s: should this be conditioned on object != nullptr?

src/test/abi_component/widget.cpp
devhawk: copy/paste error from pal_error.h. Fixed there too

src/test/platform/activation.cpp
devhawk: Because originally I was using IWidget instead of IUnknown. Will change to use from PAL

src/test/abi_component/widget.cpp
devhawk: again copy/paste from existing PAL files. I'll let @DefaultRyan comment, but I suspect these should all be noexcept. XLANG_NOEXCEPT appears to be there to enable pal.h to be compiled by a C compiler.

src/platform/common_activation.cpp
```diff
@@ -2,21 +2,45 @@
 #include "platform_activation.h"
 #include "string_convert.h"
 #include "pal_error.h"
+#include <string>
+#include <dlfcn.h>
 
 namespace xlang::impl
 {
     xlang_pfn_lib_get_activation_factory try_get_activation_func(
```
kennykerr: `func`? Prefer to be readable.

src/test/abi_component/widget.cpp
```diff
@@ -0,0 +1,145 @@
...
+        if (result == 0)
+        {
+            delete this;
+        }
+        return result;
+    }
+
+    xlang_error_info* activate_widget(iwidget** instance) noexcept override
+    {
+        if (!instance)
```
kennykerr: No need to validate this. AV is far better for invalid input.

src/test/abi_component/widget.cpp
```diff
@@ -0,0 +1,145 @@
...
+        {
+            return xlang::originate_error(xlang_error_pointer);
+        }
+
+        *instance = new widget{};
```
kennykerr: nothrow?

src/platform/common_activation.cpp
devhawk: I don't understand this comment.

src/test/abi_component/widget.cpp
devhawk: This is the opposite of what @Scottj1s suggested

src/test/abi_component/widget.cpp
devhawk: will do

src/test/abi_component/widget.cpp
devhawk: but I agree with it, so I'll remove these unless @Scottj1s raises a stink

src/test/abi_component/widget.cpp
```diff
@@ -0,0 +1,145 @@
...
+
+extern "C" xlang_error_info* XLANG_CALL xlang_lib_get_activation_factory(xlang_string class_name, xlang_guid const& iid, void** factory)  XLANG_NOEXCEPT
+{
+    char16_t const* buffer_ref{};
+    uint32_t length_ref{};
+    xlang_error_info* result{};
+    result = xlang_get_string_raw_buffer_utf16(class_name, &buffer_ref, &length_ref);
+    if (result != nullptr)
+    {
+        return result;
```
kennykerr: You fail to clear the out param

src/platform/common_activation.cpp
kennykerr: `try_get_activation_func` - what does `func` mean? Why not use a complete word?

devhawk: That's what @DefaultRyan named it

src/test/abi_component/widget.cpp
devhawk: will fix

src/platform/common_activation.cpp
```diff
@@ -2,21 +2,45 @@
...
+        module = dlopen(module_name.c_str(), RTLD_LAZY);
+
+        if (module)
+        {
+            return reinterpret_cast<xlang_pfn_lib_get_activation_factory>(dlsym(module, activation_fn_name.data()));
+        }
+        else
+        {
+            auto error = dlerror();
+            printf("%s", error);
```
DefaultRyan: Wut?

Seriously, though. If try_get_activation_func needs to return richer error information, let's update the signature to do so.

devhawk: Uhh, woops. THis was debugging code. Will remove

PR #187, opened 2019-02-20
Final preparations for VS 16 update

Scottj1s:
This work is focused mainly on updating the visualizer to support Worker Process Remoting in VS 16. This in turn required both arches to be built and packaged in the VSIX.  It also required updating the visualizer to using the new VS SDK NuGet packages, which in turn required a move back to msbuild files from cmake.


src/scripts/windows/build.cmd
```diff
@@ -23,7 +24,8 @@ goto :init
     set "OPT_VERBOSE="
     set "OPT_FORCE_CMAKE="
     set "BUILD_TYPE=Debug"
-    set "BUILD_VERSION="
+    set "BUILD_ARCH=%VSCMD_ARG_TGT_ARCH%"
```
Scottj1s: may want to build x64 from x86 or vice versa

src/scripts/windows/build.cmd
```diff
@@ -58,14 +62,14 @@ goto :init
 :main
 
     set SRC_PATH=%REPO_ROOT_PATH%\src
-    set BUILD_PATH=%REPO_ROOT_PATH%\_build\Windows\%VSCMD_ARG_TGT_ARCH%\%BUILD_TYPE%
+    set BUILD_PATH=%SRC_PATH%\_build\Windows\%BUILD_ARCH%\%BUILD_TYPE%
```
Scottj1s: this ensures that VS builds use the same cmake cache as command line

src/CMakeLists.txt
```diff
@@ -126,20 +126,6 @@ if (WIN32)
     add_definitions(-DNOMINMAX)
 endif()
 
-# determine if VS SDK workload is installed via vswhere.exe
```
Scottj1s: now that VS SDK is NuGet-ized, don't have to worry about environment (and don't want to)

src/tool/natvis/cppwinrtvisualizer.vsdconfigxml
```diff
@@ -5,7 +5,7 @@
     <DefineGuid Name="guidCppWinRTVisualizerSourceId" Value="7168763a-71e1-4447-b5d1-a44a52986527" />
 
     <NativeComponent ComponentId="1498d919-25ee-4913-910d-6f68b224b6e7" ComponentLevel="9991500" ModuleName="CppWinRTVisualizer.DLL">
-        <Class Name="CppWinRTVisualizer" ClassId="2ea3526b-93a1-4854-ae18-9028288f6f64">
+        <Class Name="CppWinrtVisualizer" ClassId="2ea3526b-93a1-4854-ae18-9028288f6f64" WorkerProcessSupported="true">
```
Scottj1s: temporarily forking the vsdconfigxml so we can build on both VS15 and VS16

src/tool/natvis/pch.h
```diff
@@ -1,6 +1,7 @@
 #pragma once
 
 #define WIN32_LEAN_AND_MEAN             // Exclude rarely-used stuff from Windows headers
+#define NOMINMAX
```
devhawk: Isn't this specified globally in the root cmake file?

src/tool/natvis/packages.version.props
```diff
@@ -0,0 +1,7 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="4.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <PropertyGroup>
+    <!-- NOTE: This needs to stay in sync with packages.config -->
```
devhawk: Should we be generating this file + packages.config with cmake to ensure these are in sync?

src/tool/natvis/pch.h
kennykerr: The msbuild project doesn't pick up the cmake defaults.

src/scripts/windows/build.cmd
devhawk: I don't understand. Before this change, both VS and command line would build into <xlang_root>\_build. I don't see a corresponding change to CMakeSettings.json to target <xlang_root>\src\_build

src/scripts/windows/build.cmd
```diff
@@ -48,7 +50,9 @@ goto :init
     if /i "%~1"=="-b"               set "BUILD_TYPE=%~2"   & shift & shift & goto :parse
     if /i "%~1"=="--build-type"     set "BUILD_TYPE=%~2"   & shift & shift & goto :parse
 
-    if /i "%~1"=="--build-version"     set "BUILD_VERSION=%~2"   & shift & shift & goto :parse
+    if /i "%~1"=="--build-arch"     set "BUILD_ARCH=%~2"   & shift & shift & goto :parse
```
devhawk: Can you describe how this is intended to be used? This script is (was?) intnded to be used from a VsDevCmd command prompt. You specify the target and host architecture when you start the command prompt via the -arch and -host_arch command line parameters. How does --build-arch interact with the VsDevCmd settings

src/package/vsix/CMakeLists.txt
```diff
@@ -1,18 +1,20 @@
...
+string(REGEX REPLACE "x86" "x64" cppwinrt_natvis_dir_x64 ${cppwinrt_natvis_dir_x86})
 
-get_target_property(cppwinrt_natvis_dir cppwinrtvisualizer "cppwinrt_natvis_dir")
 get_target_property(cppwinrt_nupkg_dir make_cppwinrt_nupkg "cppwinrt_nupkg_dir")
 
 file(TO_NATIVE_PATH ${CMAKE_CURRENT_BINARY_DIR}/Microsoft.Windows.CppWinRT.vsix cppwinrt_vsix)
 file(TO_NATIVE_PATH "${cppwinrt_nupkg_dir}/Microsoft.Windows.CppWinRT.${XLANG_BUILD_VERSION}.nupkg" cppwinrt_nupkg)
 
-set(build_vsix msbuild ${CMAKE_CURRENT_SOURCE_DIR}/vsix.csproj /nologo /m /p:Configuration=${CMAKE_BUILD_TYPE},Platform=$ENV{VSCMD_ARG_TGT_ARCH})
+set(build_vsix msbuild ${CMAKE_CURRENT_SOURCE_DIR}/vsix.csproj /nologo /m /p:Configuration=${CMAKE_BUILD_TYPE},Platform=${XLANG_BUILD_ARCH})
```
devhawk: Are we using msbuild from cmake here for convienence (i.e. we don't need to compile this xplat) or is there something specific for vsix scenarios that cmake doesn't support?

src/scripts/windows/build.cmd
Scottj1s: Not my experience.  Command line would build into <xlang_root>_build and VS into <xlang_root>\src\_build

src/tool/natvis/pch.h
Scottj1s: correct

src/tool/natvis/packages.version.props
Scottj1s: in due course - patterned this after Gregg Miskelly's VS SDK samples

src/scripts/windows/build.cmd
Scottj1s: This allows you to override the default behavior so automated builds can explicitly invoke x86 / x64 regardless of the dev command prompt flavor.

src/package/vsix/CMakeLists.txt
Scottj1s: We're using it for both vsix and now natvis, due to lots of magic behavior in VS project file support for both.  the alternative, reinventing all this in cmake for windows-specific builds, makes little sense and introduces friction.

src/scripts/windows/build.cmd
Scottj1s: (I even tried experimenting with current dir and build.cmd always build into <xlang_root>_build)

src/package/vsix/CMakeLists.txt
Scottj1s: (Natvis builds now require a VS SDK NuGet, as of VS 16)

src/tool/natvis/packages.version.props
Scottj1s: However, I'd likely generate this from the vcxproj, to maintain standalone behavior.  I'd like to keep the cmakelists.txt as a pure msbuild invoker.

src/scripts/windows/build.cmd
Scottj1s: as it happens, I ended up invoking both arches in the natvis cmakelists, so this change is largely superfluous

src/scripts/windows/build.cmd
devhawk: Apparently, it depends on where you open VS from. I always open it from the xlang root and so it hangs _build off there. I just tried opening VS from /src directory and it created the _build directory there. <sigh>

Since you're already mucking about in here, can we figure out how to configure CmakeSettings.json to ALWAYS build it in the same place (/src/_build seems like the logical place)

src/CMakeLists.txt
devhawk: 👍

src/tool/natvis/cppwinrtvisualizer.vsdconfigxml
devhawk: Is there a deprication plan for the VS15 version?

src/tool/natvis/pch.h
devhawk: 👍

src/tool/natvis/packages.version.props
devhawk: Makes sense. File an issue to track please

src/scripts/windows/build.cmd
devhawk: So lets back this out

src/package/vsix/CMakeLists.txt
devhawk: 👍

src/scripts/windows/build.cmd
Scottj1s: If we ever wanted to build x64 from a build slave, how would we do that?

src/scripts/windows/build.cmd
Scottj1s: I'll take a look

src/tool/natvis/cppwinrtvisualizer.vsdconfigxml
Scottj1s: Ben would like us to maintain it indefinitely, until/unless there's a forced reason not to.

src/scripts/windows/build.cmd
Scottj1s: ok, I can launch devenv root, with current directory pointing to root cmake cache (not under src) to enable Intellisense in VS - will back this out.

src/tool/natvis/packages.version.props
Scottj1s: turns out the props file wasn't even needed

src/scripts/windows/build.cmd
Scottj1s: Since it's purely an override, I'll go ahead and leave it in

devhawk:
File #189 to track moving /src/tool/natvis to live along side its siblings /src/pacakge/nuget and /src/package/vsix

src/CMakeLists.txt
devhawk: I was unable to build make_cppwinrt_vsix on my devbox. I don't have the VisualStudioExtension workload installed. THis code was checking that so we would bypass building natvis on machines that don't have the correct config. I think that's broken now

Scottj1s: thanks!  I've migrated the project to using the Microsoft.VSSDK.BuildTools NuGet (although to load it in VS still requires the VS SDK MSI, but builds are at least independent)

devhawk: You mean loading the vsix.sln and/or cppwinrtvisualizer.sln (or assoicated *proj files) in VS requires the VS SDK MSI, right? If I load the cmake project in VS, do I need VS SDK MSI?

src/scripts/windows/build.cmd
devhawk: I don't think you can do that in cmake. For command line, you have to be in a VsDevCmd prompt so you have to choose your host and target archtiectures ahead of time. For VS, you have to select a specific configuration that includes target archtiecture. See docs on [CMakeSettings inheritEnvironments](https://docs.microsoft.com/en-us/cpp/ide/cmake-tools-for-visual-cpp?view=vs-2017#inherited-environments-visual-studio-2017-version-155)

Scottj1s: can you elaborate?  this change is simply treating the VSCMD_ARG_TGT_ARCH env var as a default

devhawk: My point is that you can't override the target archtiecture of the cmake build with this argument. It will still pick up the target arch from whatever settings the developer chose when they launched the VsDevCmd prompt. So if you're in an x86 arch prompt, this argument makes it look like you can build for x64, but you can't.

Also, it looks like XLANG_BUILD_ARCH is only used in two places
* src/package/natvis/CMakeLists.txt adds it to the cppwinrt_natvis_dir path. But doesn't that cmake file ALWAYS build x86 and x64? Why does cppwinrt_natvis_dir need to include the chosen XLANG_BUILD_ARCH in the path if it's going to build both anyway?
* src/package/vsix/CMakeLists.txt uses it for the MSBuild /p:Platform value, but that entire build script is predicated on XLANG_BUILD_ARCH being x86. I'm pretty sure you could use [CMAKE_LIBRARY_ARCHITECTURE](https://cmake.org/cmake/help/v3.9/variable/CMAKE_LIBRARY_ARCHITECTURE.html) instead.

src/CMakeLists.txt
Scottj1s: correct - loading the cmake folder does not load the csproj

src/scripts/windows/build.cmd
Scottj1s: ok, I'll back this out

devhawk: 👍

PR #188, opened 2019-02-20
C++/WinRT now correctly handles out params

kennykerr:
Out params are pesky things and WinRT arrays are peskier still. Together it's a perfect storm of COM trivia. With this update, C++/WinRT is considerably more robust and resilient to abuse when it comes to out param and arrays, whether those parameters arrive via a language projection or from some cranky COM developer using the raw ABI and refusing to initialize variables consistently. Either way, C++/WinRT will do the right thing when it comes to handing off projected types to the ABI (by remembering to release any resources) and when it comes to zeroing out or clearing out parameters that arrive across the ABI.

I have [very](https://github.com/kennykerr/xlang_vs/blob/master/tests/test_cpp/out_params.cpp) [comprehensive](https://github.com/kennykerr/xlang_vs/blob/master/tests/test_cpp/out_params_abi.cpp) testing that I'm going to bring in with a separate PR once Scott's cmake/msbuild  heretical changes are introduced. Those tests cover both ABI and projection consumption to ensure all code paths are validated (even those that cannot possibly happen with a well-behaved language projection).


src/tool/cpp/cppwinrt/code_writers.h
```diff
@@ -587,7 +587,7 @@ namespace xlang
                 }
                 else
                 {
-                    format = "%.size(), get_abi(%)";
+                    format = "%.size(), put_abi(%)";
```
kennykerr: Finally "FillArray" out params actually uses put_abi (sensible) rather than get_abi (nonsensical). 😉

PR #190, opened 2019-02-22
PyWinRT Enum support

devhawk:
This PR generates Python enums from WinRT enums. It also fixes up a few things that were busted with array support

Note, when manually testing this change, I discovered a unrelated test failure (i.e. it's happening in master) that only surfaces on x86. Filed #191 to track.

Example generated code available here: https://github.com/devhawk/pywinrt-output/tree/enum


src/tool/python/code_writers.h
```diff
@@ -12,6 +12,41 @@ namespace xlang
         w.write("@ = __ns__.@\n", type.TypeName(), type.TypeName());
     }
 
+    void write_python_enum_name(writer& w, std::string_view const& name)
+    {
+        // None is a python keyword, so project it as NONE
+        if (name == "None")
+        {
+            for(auto&& c : name)
```
Scottj1s: why not just write("NONE")

devhawk: in case I need to add other keywords down the road

PR #193, opened 2019-02-25
Final C++/WinRT project structure for OS ingestion

kennykerr:
This update includes the final C++/WinRT project structure in preparation for OS ingestion for an upcoming Windows SDK (and the OS build itself).

This update also includes many many additional unit tests as well as the inclusion of the formerly internal legacy tests that are still quite substantial. 

This update also includes a much simpler VS-based dev environment to dramatically improve developer productivity.


src/test/component/CMakeLists.txt
```diff
@@ -1,6 +1,6 @@
 cmake_minimum_required(VERSION 3.9)
 
-file(TO_NATIVE_PATH ${CMAKE_BINARY_DIR}/tool/cpp/cppwinrt/cppwinrt.exe cpp_exe)
+file(TO_NATIVE_PATH ${CMAKE_CURRENT_SOURCE_DIR}/../../tool/cppwinrt/build/cppwinrt.exe cpp_exe)
```
DefaultRyan: Seems like a bad idea. The CMake binary directory is set independently of the src directory structure.

kennykerr: This is intentional. The cppwinrt binaries all build to a separate folder optimized for the developer inner loop.

devhawk: Why can't this separate folder live under CMAKE_BINARY_DIR?

.gitignore
```diff
@@ -341,4 +341,8 @@ src/test/python/output/
 src/test/python/__pycache__/
 
 # Ignore copied license
-src/package/vsix/LICENSE
\ No newline at end of file
+src/package/vsix/LICENSE
+
+build
+*.winmd
+*_h.h
```
devhawk: Are we adding this ignore because you're generating output into the source directories? This really isn't a good idea

src/test/CMakeLists.txt
```diff
@@ -7,7 +7,6 @@ add_subdirectory(abi_component)
 
 if (WIN32)
     add_subdirectory(component)
-    add_subdirectory(cpp)
```
devhawk: We don't have C++ tests anymore?

.gitignore
kennykerr: The old test projects rely on these files being created in the respective project dirs. This is the typical way that these msbuild-based projects work. I didn't want to mess with them until I have more time to clean then up.

src/test/CMakeLists.txt
kennykerr: Not universal tests no. We have cppwinrt-specific tests only. Once Ryan has the xlang version spun up there will likely be xlang-specific C++ tests for that.

src/test/component/CMakeLists.txt
kennykerr: The cmake binary dir is very inconvenient for dev inner loop.

devhawk:
cpp_test step failed: https://dev.azure.com/msft-xlang/public/_build/results?buildId=723

src/test/component/CMakeLists.txt
kennykerr: I could of course make the cmake build default to the CMAKE_BINARY_DIR while the VS build uses a more intuitive location for local builds.

kennykerr:
Closing the PR due to objections over the use of VS project support.

.gitignore
Scottj1s: you could either pass the cmake_binary_dir in to your project and use that to configure build output outside the src tree, or take a deferral and clean this up later (open a work item)

kennykerr: Done: https://github.com/Microsoft/xlang/issues/195

src/tool/cppwinrt/CMakeLists.txt
```diff
@@ -0,0 +1,21 @@
+cmake_minimum_required(VERSION 3.9)
+
+set(build msbuild /nologo /m /p:Configuration=${CMAKE_BUILD_TYPE},Platform=$ENV{VSCMD_ARG_TGT_ARCH} /p:CppWinRTVersion=${XLANG_BUILD_VERSION} ${CMAKE_CURRENT_SOURCE_DIR}/cppwinrt.sln)
```
Scottj1s: this is where you could pass something like:
/p:OutDir=${CMAKE_CURRENT_BINARY_DIR}\build
/p:IntDir=${CMAKE_CURRENT_BINARY_DIR}\build\temp\prebuild

Scottj1s: (and remove/default these in all vcxproj files)

kennykerr: Ah, perfect.

.gitignore
kennykerr: Actually I'll get this fixed right away.

src/tool/cppwinrt/CMakeLists.txt
Scottj1s: you could also have a vcxproj default that happens to align with what cmake passes in (or simply use that default all the time).  See here the CmakeBinaryDir property here for an example: https://github.com/Microsoft/xlang/blob/scottj1s/javax/src/tool/java/javart/vs/javart.vcxproj

kennykerr: done

src/test/component/CMakeLists.txt
kennykerr: done

.gitignore
kennykerr: fixed

src/tool/cppwinrt/cppwinrt.props
```diff
@@ -0,0 +1,62 @@
...
+      <PreprocessorDefinitions>XLANG_VERSION_STRING="$(CppWinRTVersion)";NOMINMAX;%(PreprocessorDefinitions)</PreprocessorDefinitions>
+      <MultiProcessorCompilation>true</MultiProcessorCompilation>
+    </ClCompile>
+  </ItemDefinitionGroup>
+
+  <PropertyGroup>
+    <CmakePlatform>x64</CmakePlatform>
+    <CmakePlatform Condition="'$(Platform)'=='Win32'">x86</CmakePlatform>
+    <CmakeOutDir>$(SolutionDir)..\..\..\_build\Windows\$(CmakePlatform)\$(Configuration)\tool\cppwinrt\</CmakeOutDir>
+    <OutDir>$(CmakeOutDir)</OutDir>
```
Scottj1s: did you also need to set IntDir?

src/tool/cppwinrt/cppwinrt.props
```diff
@@ -0,0 +1,62 @@
...
+
+This file sets PlatformToolset/WindowsTargetPlatformVersion to allow this project to compile with multiple versions
+of Visual Studio.
+-->
+
+<PropertyGroup>
+    <PlatformToolset>v140</PlatformToolset>
+    <PlatformToolset Condition="'$(VisualStudioVersion)' == '15.0'">v141</PlatformToolset>
+    <PlatformToolset Condition="'$(VisualStudioVersion)' == '16.0'">v142</PlatformToolset>
+    <CppWinRTVersion Condition="'$(CppWinRTVersion)'==''">2.3.4.5</CppWinRTVersion>
```
Scottj1s: seems arbitrary - maybe 2.0.0.0 ?

src/tool/cppwinrt/cppwinrt/cppwinrt.vcxproj
```diff
@@ -0,0 +1,273 @@
...
+  </ImportGroup>
+  <ImportGroup Label="PropertySheets" Condition="'$(Configuration)|$(Platform)'=='Debug|x64'">
+    <Import Project="$(UserRootDir)\Microsoft.Cpp.$(Platform).user.props" Condition="exists('$(UserRootDir)\Microsoft.Cpp.$(Platform).user.props')" Label="LocalAppDataPlatform" />
+  </ImportGroup>
+  <ImportGroup Label="PropertySheets" Condition="'$(Configuration)|$(Platform)'=='Release|x64'">
+    <Import Project="$(UserRootDir)\Microsoft.Cpp.$(Platform).user.props" Condition="exists('$(UserRootDir)\Microsoft.Cpp.$(Platform).user.props')" Label="LocalAppDataPlatform" />
+  </ImportGroup>
+  <PropertyGroup Label="UserMacros" />
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|Win32'">
+    <IntDir>$(OutDir)temp\$(ProjectName)\</IntDir>
```
Scottj1s: oh, I see what you're doing with IntDir

kennykerr: For some reason, it didn't work when set in the props file.

src/tool/cppwinrt/cppwinrt.props
kennykerr: It was intentional to test the generation of the rc file.

src/tool/cppwinrt/cppwinrt/cppwinrt.vcxproj
kennykerr: I think because $(ProjectName) was not set yet.

Scottj1s:
>     writer strings_h;

This is what the 4th time we've done this?  Powershell next time!

---
Refers to: src/tool/cppwinrt/prebuild/main.cpp:16 in 89831f2. [](commit_id = 89831f230cada941324ea73fab3d722f8eb76a78, deletion_comment = False)

Issue #104, opened 2018-11-30
clean up python projection error handling and ref counting


devhawk:
Need RAII wrappers to ensure correct lifetime management and error checking for Python object pointers 

Example https://github.com/zooba/pyptr

PR #196, opened 2019-03-01
Python/WinRT Assorted Fixes

devhawk:
* Manage PyObject* lifetimes with RAII wrappers
* add dealloc functions for structs and winrt_base
* rework native __await__ function for IAsync* types to better match original Python implementation
* fix array view construction for pass_array parameters
* rework how root module name is accessed in the projection generation code 
* add module name to projected types
* Update python tests to include generated folder in search path
* Remove explicit init apartment calls from python tests
* remove explicit debug flags (/Zi and /Od) from setup.py

Example generated code available: https://github.com/devhawk/pywinrt-output/pull/6/files


src/tool/python/strings/pybase.h
```diff
@@ -750,36 +750,52 @@ namespace py
         }
     };
 
+    struct pyobj_ptr_traits
+    {
+        using type = PyObject*;
```
Scottj1s: >PyObject* [](start = 21, length = 9)

consider typedef-ing a PyObjectPtr and using it throughout this header - everywhere you pass and return raw PyObject* pointers, to make things exception safe.

devhawk:
fixes #191

devhawk:
fixes #104

src/tool/python/strings/pybase.h
devhawk: Done. Also cleaned up the way WinRT delegates and async complete callback capture and clean up python state

devhawk:
https://github.com/devhawk/pywinrt-output/pull/6/files updated w/ latest projected code

PR #206, opened 2019-03-07
Remove fixed size buffer when reading lines from the response file.

jlaanstra:
This PR removes a hardcoded maximum line length and instead, dynamically allocates a string to read lines. 

Currently cppwinrt.exe silently fails. @kennykerr thoughts on outputting some kind of error when the response file contains invalid data?

### DIFF
--- a/src/library/cmd_reader.h
+++ b/src/library/cmd_reader.h
@@ -271,7 +271,7 @@ namespace xlang::cmd
             }
             std::string line_buf;
             std::ifstream response_file(absolute(response_path));
-            while (getline(response_file, line_buf)) // getline calls line_buf.erase()
+            while (getline(response_file, line_buf))
             {
                 size_t argc = 0;
                 std::vector<std::string> argv;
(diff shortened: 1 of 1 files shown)


kennykerr:
You can report errors using `throw_invalid()`.

jlaanstra:
Looks like that might be handled already as it hits the same code path as if it was passed in without a response file.

src/library/cmd_reader.h
```diff
@@ -269,9 +269,9 @@ namespace xlang::cmd
             {
                 throw_invalid("'@' is reserved for response files");
             }
-            std::array<char, 8192> line_buf;
+            std::string line_buf;
             std::ifstream response_file(absolute(response_path));
-            while (response_file.getline(line_buf.data(), line_buf.size()))
+            while (getline(response_file, line_buf)) // getline calls line_buf.erase()
```
Scottj1s: comment unnecessary

jlaanstra:
```suggestion
            while (getline(response_file, line_buf))
```

kennykerr:
Thanks Johan!

PR #202, opened 2019-03-06
Python/WinRT: Update ML sample for latest version python projection 

devhawk:

PR #217, opened 2019-03-11
Devhawk/pyoverhaul

devhawk:
Draft PR for folks who commented on #202 to make suggestions related to python projection


devhawk:
Original: https://github.com/Microsoft/xlang/pull/202#discussion_r263843840 from @Scottj1s 

winrt_type could be simplified in a couple ways:
1. inherit python_type via a base_winrt_type
2. pull get_python_type out into a noexcept function template - we use this pattern a lot in cppwinrt

devhawk:
Original: https://github.com/Microsoft/xlang/pull/202#discussion_r263844363 from @Scottj1s 

curious about the use of __has_include - is it only to light up related behavior? if so, shouldn't that be the developers decision (whether to incur the compile time, etc)?

devhawk:
Original https://github.com/Microsoft/xlang/pull/202#discussion_r263848356 from @Scottj1s 

this is a good sample to comment on about a few things, as we're trying to reduce image size:
* consider a py::try_create(create_func) to consolidate all the try/catch blocks, wrapping the incoming create_func lambda
* consider passing PyObject * args as a PyConvertibleObject * args, providing implicit conversion so that args[n] can be passed directly as a param instead of requiring explicit convert_to calls
* also starts to make you wonder whether direct ABI calls might be better - eliminate all the try/catch and go straight HRESULT -> PyErr

devhawk:
original https://github.com/Microsoft/xlang/pull/202#discussion_r263860447 by @Scottj1s 

nit: short names are ambiguous (cb: count of bytes?)

devhawk:
original https://github.com/Microsoft/xlang/pull/202#discussion_r263861539 by @Scottj1s 

minor thought: derive each integral converter from a common implementation:
``` cpp
converter<int8_t> : integral_converter<int8_t, INT8_MIN, INT8_MAX>
```
or, to pick up the 32 & 64 bit variants, move the range check to the subclass and CRTP it:

``` cpp
converter<int8_t> : integral_converter<converter<int8_t>>
{
  using type = int8_t;
  static void check_range(int32_t result)
  {
    if (result < INT8_MIN || result > INT8_MAX)
    {
        throw winrt::hresult_invalid_argument();
    }
  }
}

converter<int32_t> : integral_converter<converter<int32_t>>
{
  using type = int32_t;
  static void check_range(int32_t){}
}
```

Scottj1s:
yes regarding PyConvertibleObject, not much value beyond prettying the code

---
In reply to: [471405930](https://github.com/Microsoft/xlang/pull/217#issuecomment-471405930) [](ancestors = 471405930)

src/tool/python/helpers.h
```diff
@@ -486,6 +488,73 @@ namespace xlang
...
+                    push_func(method_info{ method, info.type_arguments });
+                }
+            }
+        }
+        else
+        {
+            throw_invalid("only classes and interfaces have methods");
+        }
+
+#ifdef XLANG_DEBUG
```
Scottj1s: cruft?

devhawk:
I will open a new non-draft PR for this work

PR #227, opened 2019-03-17
Fix indentation for idl output

jonitis:
I was looking at output of idl and noticed that custom attributes for methods had wrong indentation of 4 spaces when method itself had 8 space indentation. Like this:

```
    interface ILearningModelStatics
    {
    [Windows.Foundation.Metadata.RemoteAsyncAttribute]
    [Windows.Foundation.Metadata.OverloadAttribute("LoadFromStorageFileAsync")]
        Windows.Foundation.IAsyncOperation`1<LearningModel> LoadFromStorageFileAsync(Windows.Storage.IStorageFile modelFile); 
```
vs.
```
    interface ILearningModelStatics
    {
        [Windows.Foundation.Metadata.RemoteAsyncAttribute]
        [Windows.Foundation.Metadata.OverloadAttribute("LoadFromStorageFileAsync")]
        Windows.Foundation.IAsyncOperation`1<LearningModel> LoadFromStorageFileAsync(Windows.Storage.IStorageFile modelFile); 
```

Inspired by devhawk python tool approach I used his indent_guard class to automatically increase indentation by one level where needed. Do not use hardcoded indentations of 4 or 8 spaces. Instead always use placeholder at the start of line where write_indent() will write proper number of spaces, depending on the current indentation level.

- Now write(CustomAttribute const& attr) and write_uuid(CustomAttributeSig const& arg) are similar to other other functions that they do not have enforced newline and indent at the start of line. The callers of these functions ensure that these are prepended.
- All manual loops that wrote custom attributes are replaced with bind_each<write_custom_attribute>(attribute)

I verified that idl output for windows.winmd differs only by whitespace changes for fixed custom attributes.

### DIFF
--- a/src/tool/idl/main.cpp
+++ b/src/tool/idl/main.cpp
@@ -14,8 +14,30 @@ template <typename...T> overloaded(T...)->overloaded<T...>;
 struct writer : writer_base<writer>
 {
     using writer_base<writer>::write;
+
+    struct indentation {
+        int32_t value{ 0 };
+    };
+
+    struct indent_guard
+    {
+        explicit indent_guard(writer& w, int32_t offset = 1) noexcept : _writer(w), _offset(offset)
+        {
+            _writer.indent.value += _offset;
+        }
+
+        ~indent_guard()
+        {
+            _writer.indent.value -= _offset;
+        }
+
+    private:
+        writer& _writer;
+        int32_t _offset;
+    };
+
     std::string_view current;
-    int32_t indent{ 0 };
+    indentation indent;
 
     std::vector<std::pair<GenericParam, GenericParam>> generic_param_stack;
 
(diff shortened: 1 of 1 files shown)


jonitis:
Probably it is worth to make dedicated type for indentation and then override write() with it as parameter (like abi tool does). It would make write() calls less verbose by avoiding explicit mention of write_indent().

devhawk:
The IDL tool was a never intended to be a real tool - it was primarily used as a testbed while @kennykerr  and @DefaultRyan were originally building the metadata reader. As such, I'm not sure it's worth it to fix. I just filed #248 to move this code into the samples directory

src/tool/idl/main.cpp
```diff
@@ -467,8 +489,21 @@ struct writer : writer_base<writer>
             write(")]");
         }
     }
+
+    void write(indentation indent_level)
```
devhawk: The python projection writer already has an implementation for automatically injecting indententation after a newline. I think that's a better design than having to explicitly write indentation. Might be worth factoring that out to make it easier for other languag projections to consume.

jonitis: Thanks for clarification. When looking how library works I've initially looked at cppwinrt (because I knew what it is supposed to do) and then at idl, since it was the smallest and easiest to understand. I've seen how python tool was post-processing write buffer to indent multi-line output. I can take this approach for idl. Even if tool is used only as sample code.

devhawk: If you're going to work in IDL project, do you mind also moving it over to samples dir?

jonitis:
As @devhawk suggested I've changed code to not explicitly write indentation prefixes, but rather use python binding tool approach that automatically indents text output by specializing writer::write_impl() methods.

I am closing this pull request, because it was based on changes in master branch. Will create new pull request based on feature branch.

PR #229, opened 2019-03-20
Add feature staging support to C++/WinRT

kennykerr:
Windows APIs typically need to be flighted such that teams working on different OS branches can use hold-back support to develop experimental or staged features while disabling those features from shipping builds (or different branches). This update adds support for OS developers to make use of feature staging metadata to elide component support for selected APIs. 

There are a few points to consider:

* Feature attributes only appear in internal metadata that is never shipped publicly. So in practice, this support is only for OS builds and there will never be any elision for components authored with published metadata.

* Language projections always include all types in a given projection if those types appear in metadata, regardless of any Feature attributes that may be present. 

* The C++/WinRT component support for feature staging only holds back the implementation of feature-staged interfaces and runtime classes. The headers will still include types that are "always disabled", but a given component's generated scaffolding will not implement those types. This is intentional to avoid build instability and is consistent with how OS developers using WIL + WRL would manage this manually.


src/tool/cppwinrt/cppwinrt/helpers.h
```diff
@@ -157,6 +157,22 @@ namespace xlang
...
+        if (!feature)
+        {
+            return false;
+        }
+
+        // The first argument is a Windows.Foundation.Metadata.FeatureStage enum class.
+        auto stage = std::get<ElemSig::EnumValue>(std::get<ElemSig>(feature.Value().FixedArgs()[0].value).value);
+
+        // FeatureStage::AlwaysDisabled is zero.
+        return std::get<int32_t>(stage.value) == 0;
```
DefaultRyan: I *suppose* in this case, this is always true, but it's not exactly setting a great precedent for decoding enums from metadata without relying on hard-coded integer values. We've had some comments about usability of this library, and we already have nearly all the pieces in place to do this in a more reliable manner.

Something like this would do the trick:
```c++
// Added to ElemSig::EnumValue
bool equals_enumerator(std::string_view name) const
{
  // Perhaps these first few lines should be EnumDefinition::get_enumerator(std::string_view name)?
  auto fields = type.m_typedef.FieldList();
  auto field = std::find_if(begin(fields), end(fields), [name](Field const& f) { return f.Name() == name; });
  XLANG_ASSERT(field != end(fields));

  // EnumValue::value_type is a strict subset of Constant::constant_type - simple to convert.
  auto input_value = std::visit([](auto&& v) { return Constant::contant_type{ v }; }, value.value);
  return field.Constant().Value() == input_value;
}
```
This would better serve the library in the long term and immediately improve readability of these sorts of checks - cppwinrt's handling of ComposableAttribute would also benefit.

src/tool/cppwinrt/test_component/test_component.idl
```diff
@@ -244,4 +244,77 @@ namespace test_component
...
+    namespace Velocity
+    {
+        // Don't ever define these feature_name definitions yourself!
+        // This is only for testing.
+        // API authors must use the feature macros generated by the build in
+        // order to allow features to be disabled/enabled on given branches.
+        feature_name Feature_AlwaysDisabled = { AlwaysDisabled, TRUE };
+        feature_name Feature_AlwaysEnabled = { AlwaysEnabled, TRUE };
+
+        // This interface is AlwaysDisabled thus in the type system but not
```
DefaultRyan: Superfluous "thus" in this comment and the others.

PR #231, opened 2019-03-20
Harden C++/WinRT even further to protect against bad COM code in Xaml

kennykerr:


kennykerr:
@jonwis This PR covers all these that I can think of where faulty COM code would trip us up. The rest should be guarded and never escape the projection.

PR #234, opened 2019-03-21
More C++/WinRT hardening and now with tests!

kennykerr:
I've added some tests for bad implementations. 

Once things settle down I'll add some diagnostics to help pinpoint (blame) bad implementations.


Scottj1s:
Given the confusion over whether failed calls should initialize out params, here's the definitive statement (yes):
All out pointer parameters must explicitly be set to NULL. These are usually passed in a pointer-to-pointer parameter but can also be passed as members of a structure that the caller allocates and the called code fills. The most straightforward way to ensure this is (in part) to set these values to NULL on function entry. This rule is important because it promotes more robust application interoperability.
https://docs.microsoft.com/en-us/windows/desktop/com/memory-management-rules

PR #235, opened 2019-03-22
XDN11 - Xlang Error handling

manodasanW:
This is the intial draft of the xlang error handling design note.  Please leave any feedback.


design_notes/XDN11 - Error handling.md
```diff
@@ -0,0 +1,194 @@
...
+	IUnknown* languageInformation)
+```
+
+If the error is a propagation of a previous xlang error (i.e. the component called another xlang projected component which returned an error), then the associated error info should be retrieved along with the associated xlang error.  The retrieved xlang error should be used as the error instead of translating the language error to a new xlang error.  This will allow the identity of the error to be independent of the translations it may go through as it propagates through various languages and for it to be defined as what the initial projection which returned the error determined it was equivalent to.  Diagnostic information collected by the language should still be added to the error info as it can provide useful information in the case the error becomes a failure.  This can be done by a call to xlang_propagate_error which will add the provided information to the given error info and store it on the thread level storage.
+
+xlang_propagate_error will be defined in the PAL and will take the following parameters:
+- an IXlangErrorInfo* for the error which is being propagated
+- a xlang_string with the language specific error
```
jonwis: Nit: Consider adding \`foo\` (tics) around things that are supposed to be symbols.

design_notes/XDN11 - Error handling.md
```diff
@@ -0,0 +1,194 @@
...
+	xlang_string languageError,
+	xlang_string stack,
+	xlang_string projectionIdentifier,
+	IUnknown* languageInformation)
+```
+
+After calling xlang's origination / propagation API, the projection should return the error as the result so that the consuming projection can receive the error and propagate it to the language it projects.  In some cases, the projection may want to call other functions (i.e. cleanup) before returning the error.  If any of these functions error, the projection may treat them as best effort and ignore the error and still propagate the previous error after calling them.  But, those functions may call the error origination API upon error and override the error info stored on the thread local storage.  To avoid this, the projection should retrieve the error info from the thread local storage before doing the best effort work.  Then after doing that work, it should set the retrieved error info back on the thread local store and then return the error.  To help with this, xlang will provide the xlang_get_error_info and xlang_set_error_info API.
+
+```
+void xlang_get_error_info(IXLangErrorInfo** errorInfo)
```
jonwis: Is this at the ABI level? Are there standardized projections on top of this, like `static XLangErrorInfo XLangErrorInfo::Current()` and `static void XLangErrorInfo::Set(XLangErrorInfo const&)`

design_notes/XDN11 - Error handling.md
```diff
@@ -0,0 +1,194 @@
...
...age isn't affected by object slicing.  For each of the errors from xlang that the projection maps to, a new type inheriting the native exception type for that error can be created and be used to store the error info.  This new type would just be an implementation detail of the projection and would not be caught directly by components written in that language.  Instead components would use the native exception types which these types inherit from to catch exceptions.  Projections would use these custom types to determine whether it is a translated error and to get the associated error info.  For error code based projections, an approach would be to use a thread level storage to store the error info.  The consuming projection would store the error info there, while the producing projection would retrieve it from there when the same error propagates to the producing side.  If the error is handled in the consumer, the consumer should have an API provided by the projection to clear the stored error info to ensure proper cleanup of it.  The consumer should also have an API available to get/set the stored error info if the error gets passed across threads or to log it if a failure occurs.
```
jonwis: Consider breaking each of these approaches into its own \#\#\# section.

design_notes/XDN11 - Error handling.md
```diff
@@ -0,0 +1,194 @@
...
+```
+
+## Providing additional language specific information
+
+Most of the parameters for xlang_originate_error and xlang_propagate_error are self-explanatory from the descriptions. But one parameter to call out is the IUnknown parameter used to provide additional language specific information.
+
+A projection can use this parameter to provide additional diagnostic information about the error.  It can do this by passing a COM object implementing the IXlangErrorValue interface.  This interface has a GetValue function which, when implemented, will allow to provide the information as a string.  This design allows any projection that the error propagates through to make this information available without it needing to know the specifics of what additional information each projection provides.
+
+```
+interface IXlangErrorValue : IUnknown
```
jonwis: I'm confused by the relation of this to `IXlangErrorInfo` defined below. Maybe 'Value' is the wrong name for this, and it should be `IXlangErrorExtendedInfo` ?  "Extended info" vs "language info" indicates that it's just generic information and might be related to the runtime of the language vs. the language itself.

Is the intent that a projection could QI for its own `IPythonErrorExtendedInfo` which extends this and has a `GetPyDump` method?

design_notes/XDN11 - Error handling.md
```diff
@@ -0,0 +1,194 @@
...
+
+Consuming projections should also retrieve the error info associated with the xlang error and associate it in some way with the native error.  It should be associated in a way that allows the logging of the native error to also log the information in the error info.  For instance, in certain languages, exceptions have a ToString function which consumers of that language call upon failure to log the error for later analysis.  This ToString function should also return the information in xlang's error info to allow the developer to get a better picture of the failure.  In addition, it should also be associated in a way that allows a producing projection for the same language to determine if there is an error info associated with the native error and to be able to retrieve it if there is.  This will allow the producing projection to propagate the error with the previously collected diagnostic information if the error ends up being propagated by the component instead of being handled.
```
jonwis: Feels like we're defining a bunch of rules for a projection to be Good.

Maybe pull out the top-line items into a "Guidance for Projection Authors" section, that has things like:

* **DO** invoke `xlang_originate_error` when detecting an error during a call to the implementation
* **DO** invoke `xlang_propagate_error` when passing along an error
* **CONSIDER** providing a projection-specific extended error information that implements `IXLangErrorExtendedInfo`
* **DO NOT** allow language-specific exceptions to leak through the projection

Might help to show a typical call sequence of {C# > projection > projection > Python} and call out the responsibilities of each element in the chain.

design_notes/XDN11 - Error handling.md
```diff
@@ -0,0 +1,194 @@
...
+
+## Error free functions
+
+Some languages have a concept of error free functions (i.e. noexcept in C++).  These functions will not be projected as error free because errors can occur while translating the call from the other language and while translating the result back (i.e. during activation of the class for the function).  But if there was any error during the execution of the function itself, the language itself may take actions based on what it does for errors in error free functions (i.e. fail fast for noexcept).
```
jonwis: Feels like this is in the "guidance to projection authors" section.  Activation can fail, but are there other 'this function really really can't fail without it being end-of-days' markers? Is there a MIDL3 marker for 'this method cannot fail' that could be projected into noexcept or per-language equivalents?

design_notes/XDN11 - Error handling.md
```diff
@@ -0,0 +1,194 @@
...
+- Author: Manodasan Wignarajah (mawign@microsoft.com)
+- Status: Draft
+
+## Abstract
+
+This document describes the error handling story for xlang and what it will support.
+
+## Overview
+
+As described in XDN01, xlang is a system that allows a component written in any supported language to be callable from any other supported language.  To achieve this, xlang needs to support a way to translate errors from one supported language to another supported language.
```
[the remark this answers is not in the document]
devhawk: I try to keep line length for markdown files under 100

design_notes/XDN11 - Error handling.md
```diff
@@ -0,0 +1,194 @@
...
+
+These translated errors may end up being handled or end up as application failures which are debugged by a developer, logged by the application, or captured in a dump.  To help facilitate this in a way that the developer has information available to them to diagnose these translated errors, xlang needs to support a way to retrieve the error context captured by a language and store it with the translated error for another language.  Errors may also end up traveling through multiple components written in different languages before reaching the failure point.  Due to this, xlang needs to support a way to store any additional error context from any language that the error propagates through.
+
+## Translation of errors
+
+Xlang's goal is to allow a component to be written using natural and familiar code for the language it is implemented in and for it to be consumed using natural and familiar code for the language which the consumer is written in.  To achieve this, xlang needs to translate errors from the natural form of one language to a natural form of that error in another language.
```
devhawk: nit, we have been using all lower case for xlang, even when it starts a sentence.

design_notes/XDN11 - Error handling.md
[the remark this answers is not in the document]
manodasanW: Yea I looked at the other design notes yesterday and fixed it based on them.

design_notes/XDN11 - Error handling.md
```diff
@@ -0,0 +1,194 @@
...
+```
+void xlang_propagate_error(
+	IXlangErrorInfo* errorInfo,
+	xlang_string languageError,
+	xlang_string stack,
+	xlang_string projectionIdentifier,
+	IUnknown* languageInformation)
+```
+
+After calling xlang's origination / propagation API, the projection should return the error as the result so that the consuming projection can receive the error and propagate it to the language it projects.  In some cases, the projection may want to call other functions (i.e. cleanup) before returning the error.  If any of these functions error, the projection may treat them as best effort and ignore the error and still propagate the previous error after calling them.  But, those functions may call the error origination API upon error and override the error info stored on the thread local storage.  To avoid this, the projection should retrieve the error info from the thread local storage before doing the best effort work.  Then after doing that work, it should set the retrieved error info back on the thread local store and then return the error.  To help with this, xlang will provide the xlang_get_error_info and xlang_set_error_info API.
```
devhawk: I know we've discussed patterns beyond storing error information in thread local storage. We need to capture those other patterns as well as our reasoning for choosing this approach. DNs are intended to capture why we made design decisions in addition to the actual design decisions themselves.

design_notes/XDN11 - Error handling.md
```diff
@@ -0,0 +1,194 @@
...
+
+## Xlang errors
+
+Xlang will use an Int32, xlang_result, to represent its errors.  The representation of this 32-bit value will be based on the same model used in COM for HRESULTs to assist with sharing error codes, but will not necessarily be defined in this design note on a bit by bit basis as is in COM.  This is because xlang will not inherit all the errors used in COM and will instead define a set of common errors with their values defined.  This will allow xlang to avoid defining all the errors available in COM on each supported platform especially when most will not be used.  This will also allow for projections to avoid needing to maintain a large mapping of errors.
+
+The following errors will be defined in xlang:
+
+``` 
+xlang_fail (0x80004005) - To be used as a generic error when no equivalent or better error mapping is available
+xlang_invalid_arg (0x80000003) - To be used to report any invalid parameter values in a call
```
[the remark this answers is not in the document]
manodasanW: Converted to table and fixed up the descriptions.

design_notes/XDN11 - Error handling.md
```diff
@@ -0,0 +1,346 @@
...
+value will be based on the same model used in COM for HRESULTs to assist with sharing error codes,
+but will not necessarily be defined in this design note on a bit by bit basis as is in COM.
+This is because xlang will not inherit all the errors used in COM and will instead define a set of
+common errors with their values defined. This will allow xlang to avoid defining all the errors
+available in COM on each supported platform especially when most will not be used. This will also
+allow for projections to avoid needing to maintain a large mapping of errors.
+
+The following errors will be defined in xlang:
+
+xlang error                | Value      | Description
```
devhawk: As per Arch meeting, this list should be smaller and we should specifically call out that these error codes should be relevant to language projection authors. Components should NOT be defining domain-specific error codes to return at the ABI layer

design_notes/XDN11 - Error handling.md
```diff
@@ -0,0 +1,346 @@
...
+
+- Language specific error (i.e. error code, exception type)
+- Error message
+- Stack trace
+- Projection identifier
+- Any additional diagnostic information
+
+Some of this information can be represented by different languages in different ways. For instance,
+in some languages, stack traces can be an array of addresses while in others it can be a list of
+function names. Due to this, xlang will store the information generically as strings and it will
```
devhawk: should we specify conventions for these string representations to enable tooling?

design_notes/XDN11 - Error handling.md
```diff
@@ -0,0 +1,346 @@
...
+failure occurred. To assist with this, xlang will provide the ability for a projection to associate
+diagnostic information with the error. The following information can be associated:
+
+- Language specific error (i.e. error code, exception type)
+- Error message
+- Stack trace
+- Projection identifier
+- Any additional diagnostic information
+
+Some of this information can be represented by different languages in different ways. For instance,
```
devhawk: As per arch meeting, can we provide helper functions for language projections for things like generating stack strings for dynamic languages (native languages should get stack traces in correct format automatically in error origination function)

devhawk: Maybe have the language register a handler for stack trace parsing

design_notes/XDN11 - Error handling.md
```diff
@@ -0,0 +1,346 @@
...
+necessary information to decide what to do with a failure.
+
+Due to this, xlang encourages the use of errors only for diagnosing failures and not for controlling
+code flow. For controlling code flow, xlang encourages the use of the Try pattern in functions.
+The Try pattern can be used to pass back custom domain specific errors as parameters if necessary.
+The parameter type and its possible values are declared in the producing component and would be translated
+across languages as is. This means the consumer can rely on using it to make decisions without being
+concerned of the translation accuracy.
+
+## Diagnosing application failures
```
devhawk: Mike brought up the idea of stowed exceptions - when an exception gets ignored but then end the code ends up in a bad case down the line and raises a new exception. Is there a way to get access to the details of the ignored exception?

devhawk: Kenny called this "proximity debugging"

design_notes/XDN11 - Error handling.md
```diff
@@ -0,0 +1,346 @@
...
+	xlang_string projectionIdentifier,
+	IUnknown* languageInformation)
+```
+
+After calling xlang's origination / propagation API, the projection should return the error as the
+result so that the consuming projection can receive the error and propagate it to the language it projects.
+In some cases, the projection may want to call other functions (i.e. cleanup) before returning the error.
+If any of these functions error, the projection may treat them as best effort and ignore the error and
+still propagate the previous error after calling them. But, those functions may call the
+error origination API upon error and override the error info stored on the thread local storage.
```
devhawk: As per arch meeting, there was significant push back to using a TLS side channel instead of passing the error information as part of the ABI as either the return value or as an out param

devhawk: If we use return value for error info pointer, will that impact our ability to use location transparent marshaling in the future

devhawk: How do we handle error's that don't have object info (like OOM)

design_notes/XDN11 - Error handling.md
devhawk: As per arch meeting, should this interface inherit from IXlangObject instead of IUnknown

design_notes/XDN11 - Error handling.md
```diff
@@ -0,0 +1,194 @@
...
+
+## Alternative to errors
+
+Even though xlang does provide a set of errors to assist with the error translation from one language to another in an equivalent form, it is limited.  In addition, when errors are translated from one language to another, there is a chance that the accuracy of the error is lost due to an exact error mapping is not available in the other language.  This means the caller of an API may not always have the accurate and necessary information to decide what to do with a failure.
+
+Due to this, xlang encourages the use of errors only for diagnosing failures and not for controlling code flow.  For controlling code flow, xlang encourages the use of the Try pattern in functions.  The Try pattern can be used to pass back custom domain specific errors as parameters if necessary.  The parameter type and its possible values are declared in the producing component and would be translated across languages as is.  This means the consumer can rely on using it to make decisions without being concerned of the translation accuracy.
```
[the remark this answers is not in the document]
manodasanW: Opened #255 to track this.

design_notes/XDN11 - Error handling.md
```diff
@@ -0,0 +1,194 @@
...
+
+## Xlang errors
+
+Xlang will use an Int32, xlang_result, to represent its errors.  The representation of this 32-bit value will be based on the same model used in COM for HRESULTs to assist with sharing error codes, but will not necessarily be defined in this design note on a bit by bit basis as is in COM.  This is because xlang will not inherit all the errors used in COM and will instead define a set of common errors with their values defined.  This will allow xlang to avoid defining all the errors available in COM on each supported platform especially when most will not be used.  This will also allow for projections to avoid needing to maintain a large mapping of errors.
+
+The following errors will be defined in xlang:
+
+``` 
+xlang_fail (0x80004005) - To be used as a generic error when no equivalent or better error mapping is available
```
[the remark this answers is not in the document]
manodasanW: I understand what you are saying.  During our last arch meeting, we discussed limiting the errors defined in xlang to ones which are likely to occur when projecting / translating a call because we want to discourage projected components from returning errors and rather have them using the try pattern.  In my next commit, I will remove a bunch of errors and scope it to ones that are likely to occur in projections.  The errors left I believe will mostly be ones that have posix mappings.  There are going to be errors which are not in the list of errors defined in xlang and those should map to xlang_fail.  We will support storing the posix error as part of the diagnostic informaton.

I don't believe we want to have a design where on Windows we use Windows errors and on Linux we use posix errors as that would most likely provide an inconsistent experience to consumers of xlang components when the errors get projected.

design_notes/XDN11 - Error handling.md
manodasanW: Addressed.

design_notes/XDN11 - Error handling.md
```diff
@@ -0,0 +1,194 @@
...
+xlang_handle (0x80000006) - To be used to report invalid handles
+xlang_abort (0x80000007) - To be used to report cancellations
+xlang_file_not_found (0x80070002) - To be used to report a file was not found
+xlang_path_not_found (0x80070003) - To be used to report a directory was not found
+xlang_arithmetic (0x80070216) - To be used to report any errors during arithmetic operations
+xlang_io (0x80131620) - To be used to report any IO errors than do not have a better error mapping
+xlang_bounds (0x8000000b) - To be used to report invalid indexes such as for arrays
+xlang_timeout (0x80131505) - To be used to report timeouts
+xlang_thread_state (0x80131520) - To be used to report that an operation can not be done due to the thread state
+xlang_thread_interrupted (0x80131519) - To be used to report an interrupt on a thread
```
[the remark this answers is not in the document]
manodasanW: Error code has been removed as we decided to limit the errors to ones which projections would encounter.

design_notes/XDN11 - Error handling.md
```diff
@@ -0,0 +1,194 @@
...
+xlang_originate_error will be defined in the PAL and will take the following parameters:
+- a xlang_result with the translated xlang error for the language specific error that occurred
+- a xlang_string with the language specific error that occurred
+- a xlang_string with the error message
+- a xlang_string with the error stack if available
+- a xlang_string with an identifier for the projection
+- an IUnknown pointer with any additional language specific information (can be null)
+
+```
+void xlang_originate_error(
```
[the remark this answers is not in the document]
kennykerr: Yes. Today language projections always follow a call to RoOriginateLanguageException with a call to GetRestrictedErrorInfo.

design_notes/XDN11 - Error handling.md
```diff
@@ -0,0 +1,345 @@
...
+- an IUnknown pointer with any additional language specific information (can be null)
+
+```cpp
+void xlang_originate_error(
+	xlang_result error,
+	xlang_string languageError,
+	xlang_string message,
+	xlang_string stack,
+	xlang_string projectionIdentifier,
+	IUnknown* languageInformation)
```
kennykerr: Should indicate that most of these are optional (and thus order them accordingly). In many cases you'll only have an error code and an error message.

design_notes/XDN11 - Error handling.md
```diff
@@ -0,0 +1,345 @@
...
+
+xlang_propagate_error will be defined in the PAL and will take the following parameters:
+- an IXlangErrorInfo* for the error which is being propagated
+- a xlang_string with the language specific error
+- a xlang_string indicating the error stack if available
+- a xlang_string indicating an identifier for the language
+- an IUnknown pointer with any additional language specific information (can be null)
+
+```cpp
+void xlang_propagate_error(
```
kennykerr: All these origination APIs would be substantially simpler if the error object were simply returned by the ABI.

design_notes/XDN11 - Error handling.md
```diff
@@ -0,0 +1,345 @@
...
+This document describes the error handling story for xlang and what it will support.
+
+## Overview
+
+As described in XDN01, xlang is a system that allows a component written in any supported language
+to be callable from any other supported language. To achieve this, xlang needs to support a way to
+translate errors from one supported language to another supported language.
+
+These translated errors may end up being handled or end up as application failures which are
+debugged by a developer, logged by the application, or captured in a dump. To help facilitate
```
bbowman: Is it worth calling at high level the difference in expectations for "implementation debugging information" and "external facing pretty errors"? When designing similar copmponents for Project Rome we took a stance similar to WinRT's RoOriginateError which is intended for an audience external to the component vs exception messages and general logging that are usually for the benefit of the component developer. These lines obviously get blurred as a consumer reports issues to the developer but we felt it was an important audience distinction to call out and make sure a solution addressed the needs of both parties.

design_notes/XDN11 - Error handling.md
```diff
@@ -0,0 +1,345 @@
...
+
+Errors coming from components written in other languages can be one of the causes of application
+failures. Just like with any other failure, developers will want to diagnose where and why the
+failure occurred. To assist with this, xlang will provide the ability for a projection to associate
+diagnostic information with the error. The following information can be associated:
+
+- Language specific error (i.e. error code, exception type)
+- Error message
+- Stack trace
+- Projection identifier
```
bbowman: What value does the projection identifier give us?

design_notes/XDN11 - Error handling.md
bbowman: Agreed. In Rome we used a lightweight object model for this where it was a vector of Stack Frames and each frame has a few strings that can tailored to the desired purpose (file name, function, etc). Some structure to this would be helpful in converting I would think.

design_notes/XDN11 - Error handling.md
```diff
@@ -0,0 +1,345 @@
...
+the information collected from the language that initially returned the error. An example of this is
+the stack trace. A language will typically only be aware of the stack within its language boundary and
+thereby the diagnostic information collected at the initial error would only have a partial stack.
+The following additional information can be associated with the error as it travels through projections:
+
+- Stack trace
+- Projection identifier
+- Any additional diagnostic information
+
+The diagnostic information will be stored in a COM object (error info) implementing the
```
bbowman: Should we try to avoid mentioning COM except when comparing systems? This would be an XLangObject really right?

design_notes/XDN11 - Error handling.md
```diff
@@ -0,0 +1,345 @@
...
+
+Producing projections may encounter a language error when projecting an API call to a component.
+The error will either be from the translation of the call to the component's language or be from
+the component itself as a result of the call. Either way, upon encountering an error, the projection
+should determine whether the error is new or is a propagation of a previous xlang error. Strategies
+on doing this will be discussed in a later section.
+
+If the error is new, the projection should translate the error to an equivalent xlang error or to
+xlang_fail if there is no equivalent mapping. After translating the error, the projection should
+call xlang's error origination API, xlang_originate_error, to store the diagnostic information
```
bbowman: I feel like this should be called out earlier as a top level thing so that readers are familiar with all the pieces of the picture.

design_notes/XDN11 - Error handling.md
bbowman: I think this doc could benefit from a discussion of error flows in async calling patterns as well.

design_notes/XDN11 - Error handling.md
devhawk: As per arch meeting, @bbowman further expounded on this topic. In particular, he clarified the idea of different audiences for error info - the component owner vs. the component user (aka developer) vs the app user (non developer). Makes sense to call this out explicitly. Also, maybe worth calling out how component user can communicate to the component owner to distinguish between "user is using the component wrong" vs. "there's a bug in the component". Having this called out in the overview seems reasonable as we describe further parts of the error model later in the doc

design_notes/XDN11 - Error handling.md
```diff
@@ -0,0 +1,345 @@
...
+xlang_fail                 | 0x80004005 | Generic error when no equivalent or better error mapping is available
+xlang_handle               | 0x80000006 | Invalid handle
+xlang_invalid_arg          | 0x80000003 | Invalid argument in a call
+xlang_no_interface         | 0x80000004 | Class or interface not found
+xlang_not_impl             | 0x80000001 | Function / feature is not implemented
+xlang_out_of_memory        | 0x80000002 | Allocation issue resulting from not enough memory
+xlang_pointer              | 0x80000005 | Null pointer related errors
+xlang_type_load            | 0x80131522 | Syntax / type error during runtime
+
+Even though xlang only defines these errors, it will not explicitly prevent the use of other errors
```
devhawk: As per arch meeting, reserve 0 as success. 

Also, we had an interesting discussion about the need for minting further error codes. Generally we have a philosophy of not wanting to use exceptions for flow control. There are suggestions for separating what we called "domain errors" from more general runtime/language errors.

design_notes/XDN11 - Error handling.md
devhawk: As per arch meeting, we want to consider keeping a buffer of errors, but that such a feature is outside the scope of this DN

design_notes/XDN11 - Error handling.md
devhawk: As per arch meeting, projection identifier indicates the language the error occurred in. We discussed the format of this identifier. This is primarily intended human consumption so it should be a string. Would be useful to include information about projection name (C++/CX vs C++/WinRT) and maybe potentially version info

design_notes/XDN11 - Error handling.md
devhawk: WinRT has a mechanism for tying together logical call stacks - the feature is called async causality. It would probably be out of scope for first iteration of the error DN  but is a good suggestion for later updates

design_notes/XDN11 - Error handling.md
devhawk: As per arch meeting, @bbowman expanded on how Rome does this. Rome will convert the lightweight OM to something Java specific so that the Java developer can see the stack in a way that feels like a native Java exception.

design_notes/XDN11 - Error handling.md
devhawk: as per arch meeting, this should be using IXlangObject

design_notes/XDN11 - Error handling.md
devhawk: Maybe we should have a diagram that shows end to end in the overview that includes error origination mention in passing, with details in this section.

design_notes/XDN11 - Error handling.md
```diff
@@ -0,0 +1,194 @@
...
+
+If the error is new, the projection should translate the error to an equivalent xlang error or to xlang_fail if there is no equivalent mapping.  After translating the error, the projection should call xlang's error origination API, xlang_originate_error, to store the diagnostic information collected by the language for the error.  The API will create an error info with the provided diagnostic information and store it on a thread level storage.  The created error info will remain on the thread level storage until it is retrieved by the consuming projection.
+
+xlang_originate_error will be defined in the PAL and will take the following parameters:
+- a xlang_result with the translated xlang error for the language specific error that occurred
+- a xlang_string with the language specific error that occurred
+- a xlang_string with the error message
+- a xlang_string with the error stack if available
+- a xlang_string with an identifier for the projection
+- an IUnknown pointer with any additional language specific information (can be null)
```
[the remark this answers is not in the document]
devhawk: As per earlier comment, this should be IXlangObject

design_notes/XDN11 - Error handling.md
devhawk: As per arch meeting, we are in agreement on returning an xlang error object instead of an error code and using a side channel for communicating rich error info. null return indicates success, non-null is an error object pointer (pre-add-reffed).

design_notes/XDN11 - Error handling.md
```diff
@@ -0,0 +1,194 @@
...
+
+Due to this, xlang encourages the use of errors only for diagnosing failures and not for controlling code flow.  For controlling code flow, xlang encourages the use of the Try pattern in functions.  The Try pattern can be used to pass back custom domain specific errors as parameters if necessary.  The parameter type and its possible values are declared in the producing component and would be translated across languages as is.  This means the consumer can rely on using it to make decisions without being concerned of the translation accuracy.
+
+## Diagnosing application failures
+
+Errors coming from components written in other languages can be one of the causes of application failures.  Just like with any other failure, developers will want to diagnose where and why the failure occurred.  To assist with this, xlang will provide the ability for a projection to associate diagnostic information with the error.  The following information can be associated:
+
+- Language specific error (i.e. error code, exception type)
+- Error message
+- Stack trace
```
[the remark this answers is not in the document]
manodasanW: Addressed.

design_notes/XDN11 - Error handling.md
```diff
@@ -0,0 +1,194 @@
...
+
+## Handling language errors
+
+Producing projections may encounter a language error when projecting an API call to a component.  The error will either be from the translation of the call to the component's language or be from the component itself as a result of the call.  Either way, upon encountering an error, the projection should determine whether the error is new or is a propagation of a previous xlang error.  Strategies on doing this will be discussed in a later section.
+
+If the error is new, the projection should translate the error to an equivalent xlang error or to xlang_fail if there is no equivalent mapping.  After translating the error, the projection should call xlang's error origination API, xlang_originate_error, to store the diagnostic information collected by the language for the error.  The API will create an error info with the provided diagnostic information and store it on a thread level storage.  The created error info will remain on the thread level storage until it is retrieved by the consuming projection.
```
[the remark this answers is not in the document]
manodasanW: No longer an issue as we are no longer using the TLS to store the error info.

design_notes/XDN11 - Error handling.md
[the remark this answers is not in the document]
manodasanW: Thought about this a bit.  I think we want to provide the capability for a projection to pass a non-projected language object to this field without needing to worry about wrapping it in an IXlangObject or implement that interface on it.  An example of this would be when a projection wants to store information to re-construct their exception later.  That object shouldn't need to be a projected type as it is intended for only them to retrieve it later.  In addition, I don't expect a projected component / consumer to be looking at this parameter, but rather projections exposing it in the language exception if it implements IXlangExtendedErrorInfo or something they recognize.  Due to that, leaving it as IUknown which is the xlagn equivalent for non projected types.

design_notes/XDN11 - Error handling.md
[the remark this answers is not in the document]
manodasanW: Addressed.

design_notes/XDN11 - Error handling.md
manodasanW: Addressed.

design_notes/XDN11 - Error handling.md
manodasanW: Given the move to returning the error info rather than storing it on a TLS, these functions were removed.

design_notes/XDN11 - Error handling.md
manodasanW: Addressed.

design_notes/XDN11 - Error handling.md
manodasanW: Yes, that sounds like a good idea.  I added into a guidance section at the top.  Will also have an upcoming change with the diagram.

design_notes/XDN11 - Error handling.md
manodasanW: Addressed, I discussed the approaches we considered.

design_notes/XDN11 - Error handling.md
manodasanW: I addressed most of this.  I didn't allow for a option for a language to register a handler, but provided the flexibility for a language to change the stack captured back us.

design_notes/XDN11 - Error handling.md
manodasanW: Addressed returning the error info rather than storing on TLS.

design_notes/XDN11 - Error handling.md
manodasanW: We are now returning the error info.

design_notes/XDN11 - Error handling.md
manodasanW: Called this out in the overview.

design_notes/XDN11 - Error handling.md
manodasanW: Addressed.

design_notes/XDN11 - Error handling.md
manodasanW: This is now a IXlangObject.

design_notes/XDN11 - Error handling.md
manodasanW: Addressed by talking about this in the overview and in the guidance to projections section.  Will also include a diagram.

design_notes/XDN11 - Error handling.md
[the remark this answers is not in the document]
BrentRe: Leaving it as IUnknown is a mistake in my opinion. In practically every case in WinRT, when we decided to "optimize" by using IUnknown instead of IInspectable (e.g. delegates), it turns out to be a mistake.

Encoding policy decisions ("dont expect") into the type system is limiting and often wrong over time.

IMO this should be IXlangObject so it's something that can be represented / manipulated within the xlang type system.

design_notes/XDN11 - Error handling.md
manodasanW: Addressed.

design_notes/XDN11 - Error handling.md
manodasanW: Addressed by calling out only the error is required.  But, it is still encouraged for projections to provide the other fields as it helps a developer in diagnosing the issue.

design_notes/XDN11 - Error handling.md
```diff
@@ -0,0 +1,194 @@
...
+
+- Stack trace
+- Projection identifier
+- Any additional diagnostic information
+
+The diagnostic information will be stored in a COM object (error info) implementing the IXlangErrorInfo interface.  As with any COM object, it will be reference counted and will implement IUnknown.  This will allow it to be easily accessed by projections written in different languages while still being able to ensure the cleanup of it.
```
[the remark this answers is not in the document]
manodasanW: Yea, I have updated the error info object to be IXlangObject.

design_notes/XDN11 - Error handling.md
```diff
@@ -0,0 +1,542 @@
...
+and resulted in a failure.
+
+![Error origination diagram](XDN11_origination_diagram.png)
+Diagram highlighting error handling of a C# app calling a Python API which calls a Java API that results in an error.
+
+## Guidance for Projection Authors
+
+- **DO** translate language errors to equivalent xlang errors and xlang errors to equivalent natural
+language errors at the projection boundary.
+- **DO NOT** let language errors propagate past the projection boundary.
```
Scottj1s: be sure to emphasize that it's exceptions specifically that must not pass the boundary

design_notes/XDN11 - Error handling.md
```diff
@@ -0,0 +1,542 @@
...
+but will not necessarily be defined in this design note on a bit by bit basis as is in COM.
+This is because xlang will not inherit all the HRESULT errors and will instead define a set of
+errors with their values defined. This will allow xlang to avoid defining all the possible HRESULT
+errors on each supported platform especially when most will not be used as xlang will be scope its errors
+to ones that are likely to occur during translation of calls. This will also allow for projections to
+avoid needing to maintain a large mapping of errors.
+
+The following errors will be defined in xlang:
+
+xlang error          | Value      | Description
```
Scottj1s: per arch discussion today, agreed that a zero-based monotonically increasing enum has a few advantages:
1. it does not confuse cross plat work by injecting Windows-isms
2. it permits direct indexing into a table (e.g., of exception factory methods)
3. it enables a simple version check for projections to translate errors they know of (<xlang_error_count) and handling all others (>=xlang_error_count) 'generically'

design_notes/XDN11 - Error handling.md
```diff
@@ -0,0 +1,542 @@
...
+xlang_bounds         | 0x8000000b | Invalid index used such as when indexing arrays
+xlang_fail           | 0x80004005 | Generic error when no equivalent or better error mapping is available
+xlang_handle         | 0x80000006 | Invalid handle
+xlang_invalid_arg    | 0x80000003 | Invalid argument in a call
+xlang_invalid_state  | 0x8000000d | Call isn't allowed in the current state
+xlang_no_interface   | 0x80000004 | Class or interface not found
+xlang_not_impl       | 0x80000001 | Function / feature is not implemented
+xlang_out_of_memory  | 0x80000002 | Allocation issue resulting from not enough memory
+xlang_pointer        | 0x80000005 | Null pointer related errors
+xlang_success        | 0x00000000 | Reserved, not to be used
```
Scottj1s: move to top of list, call out at special "Operation succeeded (should not be originated)"

design_notes/XDN11 - Error handling.md
```diff
@@ -0,0 +1,542 @@
...
+- a `xlang_string` indicating an identifier for the language
+- a `xlang_string` with the language specific error
+- a `IXlangErrorExecutionTrace` with the execution trace if available
+- an `IUnknown` pointer with any additional language specific information (can be null)
+
+```cpp
+void PropagateError(
+    xlang_string projectionIdentifier,
+    xlang_string languageError,
+    IXlangErrorExecutionTrace* executionTrace,
```
Scottj1s: executionTrace should be IUnknown, to permit both:
1. simple IStringable implementation
2. versioning later (any interface would, but IUnknown makes it clearer)

design_notes/XDN11 - Error handling.md
```diff
@@ -0,0 +1,542 @@
...
+error origination and propagation APIs. This identifier will help developers know which language the
+error came from as it may not be obvious from the language error and the execution trace. It will
+also help tooling determine whether the diagnostic information is from a projection that it recognizes
+and can light up additional features such as converting the execution trace to something developer friendly.
+
+Given that projections may evolve over time and change what they store as part of the diagnostic
+information, the identifier provided should also have a version number. Due to this, xlang recommends
+the identifier be in the form of <unique_projection_identifer>_<version_number>.
+
+## Execution traces
```
Scottj1s: per discussion, this is great stuff, but best factored out into a separate design note.  The IUnknown executionTrace param above makes a neat firewall for discussing these topics separately, and this one was just introduced.

design_notes/XDN11 - Error handling.md
```diff
@@ -0,0 +1,542 @@
...
+wrapper around it.
+
+A projection can use this parameter to provide additional diagnostic information about the error.
+It can do this by passing an object implementing the `IXlangErrorExtendedInfo` interface. This interface
+has a `GetExtendedInfo` function which, when implemented, will allow to provide the information as a string.
+This design allows any projection that the error propagates through to make this information available
+without it needing to know the specifics of what additional information each projection provides.
+
+```cpp
+interface IXlangErrorExtendedInfo : IUnknown
```
Scottj1s: per discussion, there is no need for this interface - IStringable will do

design_notes/XDN11 - Error handling.md
```diff
@@ -0,0 +1,542 @@
...
+```cpp
+interface IXLangErrorInfo
+{
+    void GetError(xlang_result* error);
+    void GetMessage(xlang_string* message);
+    void GetLanguageError(xlang_string* error);
+    void GetExecutionTrace(IXlangErrorExecutionTrace* executionTrace);
+    void GetLanguageIdentifier(xlang_string* identifier);
+    void GetLanguageInformation(IUnknown* languageInformation);
+    void GetLanguageInformationAsString(xlang_string* languageInformation);
```
Scottj1s: again, this method can go - IStringable from GetLanguageInformation will do

design_notes/XDN11 - Error handling.md
```diff
@@ -0,0 +1,542 @@
...
+
+Some languages have a concept of error free functions (i.e. noexcept in C++). These functions
+will not be projected as error free because errors can occur while translating the call from the
+other language and while translating the result back (i.e. during activation of the class for the
+function). But if there was any error during the execution of the function itself, the language
+itself may take actions based on what it does for errors in error free functions (i.e. fail fast for
+noexcept).
+
+Even though none of the projected functions will be error free, the PAL can consist of error free
+functions. These functions can be distinguished by the return parameter being void instead of
```
Scottj1s: per discussion - the ABI must never change due to introduction (or removal) of the noexcept attribute.  only the projection may choose to optimize it's codegen based on noexcept - both to remove an if check and to remove catch block at the call site.  an ABI C header could choose to omit [[nodiscard]] on [noexcept] methods, however

design_notes/XDN11 - Error handling.md
manodasanW: Addressed.

design_notes/XDN11 - Error handling.md
manodasanW: Addressed.

design_notes/XDN11 - Error handling.md
manodasanW: Addressed.

design_notes/XDN11 - Error handling.md
manodasanW: Addressed.

design_notes/XDN11 - Error handling.md
manodasanW: Removed it from this design note for now and marked this section as in progress.  Will start a new PR with this section.

design_notes/XDN11 - Error handling.md
manodasanW: Addressed by using the GetObjectInfo string representation category which seems to be what we decided on to address this interface.

design_notes/XDN11 - Error handling.md
manodasanW: Addressed.

design_notes/XDN11 - Error handling.md
manodasanW: I eneded up removing part of this as we hadn't had a chance to review that section and I want to close on this PR.  I will have a future PR update with this addressed.

design_notes/XDN11 - Error handling.md
manodasanW: Will address in future PR.

design_notes/XDN11 - Error handling.md
manodasanW: We do have a MIDL3 marker for this, but apparently not all projections recognize it.  I updated this section such that projections that do recognize it can do so and gain potential optimizations.

design_notes/XDN11 - Error handling.md
manodasanW: This interface was meant as a way to provide custom diagnosic information in a way that allows other projections to expose it without needed to know what is stored.  But, I ended up deciding we don't need to define a interface for this because there is existing xlang interfaces that allow to accomplish this.  But. as noted, a projection can pass something implementing their own interface and query for it when they retrieve this object.

design_notes/XDN11 - Error handling.md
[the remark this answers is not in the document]
manodasanW: We have decided to not base these errors directly off of HRESULT values and instead have our own values for them as it should be easy enough for projections for Window's APIs to have a switch which converts from one to the other.

design_notes/XDN11 - Error handling.md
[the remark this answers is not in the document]
manodasanW: I understand what you are saying with the delegates in WinRT.  But, the current model proposed is an app calling a xlang component shouldn't directly be interacting with the error info object.  The projection for the app's language should take the info in the error info object and expose it as part of the language error for that projection.  This means that the members we store as part of the error info object don't need to be projected, but we need to allow a projection to take that info and put its details as part of its language error representation which we do and talk about in the design note.  If we do end up seeing a need in the future where the error info object needs to be projected to the app, we can end up adding another interface to allow for that.

Also a projection is free to stored a projected object in this parameter, but it would need to expose it as part of its own interface.

PR #242, opened 2019-03-23
cppxlang able to compile against clang/linux

DefaultRyan:
Still need to redirect some of the guts from Windows to the PAL, so you can't do anything interesting without hitting link errors on clang/Linux.

But at least it's compiling cleanly now, so this is a good checkpoint to get some of this work checked in while I continue with more of the migration.


src/test/platform/CMakeLists.txt
```diff
@@ -9,7 +9,8 @@ target_sources(test_platform
 target_include_directories(test_platform
     PUBLIC ${XLANG_LIBRARY_PATH} ${XLANG_TEST_INC_PATH} ${CMAKE_CURRENT_SOURCE_DIR}/../output/component/source)
 
-CONSUME_PAL(test_platform)
+# CONSUME_PAL(test_platform)
```
jonwis: Consider removing commented code before merging.

src/tool/cppxlang/strings/base_string.h
```diff
@@ -345,70 +322,70 @@ namespace winrt
     inline hstring to_hstring(uint8_t value)
     {
         wchar_t buffer[32];
-        swprintf_s(buffer, L"%hhu", value);
+        swprintf(buffer, 32, L"%hhu", value);
```
jonwis: Consider `countof()` or `ARRAYSIZE()` or similar instead of repeating the literal size.

src/tool/cppxlang/strings/base_string_operators.h
```diff
@@ -26,9 +26,9 @@ namespace winrt
         return left == std::wstring_view(right);
     }
 
-    bool operator==(hstring const& left, nullptr_t) = delete;
+    bool operator==(hstring const& left, std::nullptr_t) = delete;
```
kennykerr: Any particular reason?

src/tool/cppxlang/strings/base_string.h
kennykerr: `_countof` is VC-specific. Perhaps `std::array` to avoid repeating 32.

src/tool/cppxlang/strings/base_string_operators.h
DefaultRyan: Standards conformance, clang rejects it otherwise.

Issue #201, opened 2019-03-06
Project Python names with correct casing


zooba:
For comfortable use from Python, it would be ideal for names to be converted into the typical casing used by idiomatic Python code.

Kind | Casing
---|---
Namespace | lowercase
Class/Interface | PascalCase
Enum | PascalCase
Function/method | lower_snake_case
Constant | UPPER_SNAKE_CASE
Enum value | UPPER_SNAKE_CASE
??? | ???

(Happy to add more, but I need reminding what else is there.)

The rule for converting PascalCase to snake_case is to find all sequences of one or more uppercase characters and insert an underscore before the last one. So `MyClass` becomes `My_Class` while `IODevice` becomes `IO_Device` (`IOD`->`IO_D`). Then upper/lower the whole thing.

Since this conversion may cause collisions, ideally there'd remain a way to access names by their unmodified form, probably through a helper function (`obj.my_function` or `get_by_original_name(obj, "MyFunction")`).

devhawk:
Namespaces are already projected as lowercase (see https://github.com/devhawk/pywinrt-output/tree/master/generated/pyrt/windows for example).

WinRT doesn't have constants beyond enum values and WinRT uses PascalCasing generally for type and member names. So I think that means we are good on type names. Please confirm. 

I assume we use lower_snake_case for *all* type members except Enum values? in particular, what about struct field names? I'm assuming lower_snake_case there 

So that leaves 
- [ ] lower snake case non enum value type members (class/interface methods, properties and events, struct fields)
- [ ] upper snake case for enum value members

zooba:
Looks good to me. Feel free to update my original table if you only want those ones called out

zooba:
Parameter names should be lower snake case as well.

devhawk:
My [first draft on write_snake_case](https://github.com/Microsoft/xlang/blob/devhawk/pyname/src/tool/python/code_writers.h#L15) adds an underscore if the current character is upper and the next character is lower. Spot checking the namespaces I use for test looks pretty good...except for places where we use UInt. Those look wrong. I'm thinking I should special case those. 

@zooba, @BrentRe I'm interested in your thoughts here

``` csharp
// C# projection
public enum PropertyType
{
    Empty = 0,
    Int16 = 2,
    UInt16 = 3
    OtherType = 20,
    UInt8Array = 1025,
    Int16Array = 1026,
    // remaining values omitted
}

public enum DirectXPixelFormat
{
    Unknown = 0,
[...]

PR #253, opened 2019-03-28
Python/WinRT - snake_casing and other changes

devhawk:
* UPPER_SNAKE_CASE enum values
* lower_snake_case methods and properties of projected types
* fold try catch blocks into trycatch_invoker template
* generate type aliases in py::wrapper namespace for more readable code
* generate projected paremeterized types in py::proj and py::impl namespaces instead of name mangling
* better naming + return value lifetime management in generated delegate wrappers
* create_python_type / register_python_type functions to eliminate repeated code from module_exec functions
* remove /debug linker flag from setup.py
* lots of internal structure improvements

As always, reviewers can see the generated code changes in the the https://github.com/devhawk/pywinrt-output repo. The https://github.com/devhawk/pywinrt-output/tree/overhaul branch contains the changes assosiated with this PR


src/test/python/complete_winml_tutorial.py
```diff
@@ -80,9 +80,9 @@ def print_results(results, labels):
     topProbabilities = [0.0 for x in range(3)]
     topProbabilityLabelIndexes = [0 for x in range(3)]
 
-    for i in range(results.Size):
+    for i in range(results.size):
         for j in range(3):
-            result = results.GetAt(i)
+            result = results.get_at(i)
```
zooba: Can we use regular iteration here? (`for i, result in enumerate(results):` if we need to keep `i` around)

src/test/python/loader_native.py
```diff
@@ -0,0 +1,15 @@
+import sys
+sys.path.append("./generated")
```
zooba: Do we need this? If so, can we resolve the full path (`os.path.abspath("generated")`) here? Otherwise the import directories move around if you change working directory throughout the library.

src/test/python/testcollections.py
```diff
@@ -4,18 +4,18 @@
...
+        m.insert("hello", "world")
 
-        self.assertTrue(m.HasKey("hello"))
-        self.assertFalse(m.HasKey("world"))
-        self.assertEqual(m.Size, 1)
-        self.assertEqual(m.Lookup("hello"), "world")
+        self.assertTrue(m.has_key("hello"))
+        self.assertFalse(m.has_key("world"))
+        self.assertEqual(m.size, 1)
+        self.assertEqual(m.lookup("hello"), "world")
```
zooba: Any specific wrappers for these yet? `__in__`, `__len__`, `__getitem__`?

src/test/python/loader_native.py
devhawk: this is just for tests, and frankly I need to both clean up the tests + move the generated code used by these tests out of the source tree

src/test/python/testcollections.py
devhawk: Yeah, IMap/IMapView support mapping protocol. Will add tests

src/test/python/complete_winml_tutorial.py
devhawk: probably. Thanks for the suggestion

devhawk: yup. New code will be:

``` python
    for i, result in enumerate(results):
        for j in range(3):
            if result > topProbabilities[j]:
                topProbabilityLabelIndexes[j] = i
                topProbabilities[j] = result
                break
```

src/test/python/CMakeLists.txt
```diff
@@ -14,50 +14,60 @@ set(included_namespaces
     Windows.Foundation
     Windows.Graphics.DirectX
 
+
     # uncomment out the following to run complete_winml_tutorial.py
-    Windows.AI.MachineLearning
-    Windows.Graphics.Imaging
-    Windows.Media
-    Windows.Storage
+    # Windows.AI.MachineLearning
```
kennykerr: Why not just remove these lines? Just curious.

src/test/python/complete_winml_tutorial.py
```diff
@@ -29,42 +29,42 @@ def sync_wrapper(*args, **kwds):
 
     return async_wrapper if asyncio.iscoroutinefunction(fun) else sync_wrapper
 
-import pyrt.windows.ai.machinelearning as winml 
+import winrt.windows.ai.machinelearning as winml
```
kennykerr: Why is the namespace lower case? Don't you honor the casing of the API?

src/test/python/complete_winml_tutorial.py
```diff
@@ -29,42 +29,42 @@ def sync_wrapper(*args, **kwds):
...
     return async_wrapper if asyncio.iscoroutinefunction(fun) else sync_wrapper
 
-import pyrt.windows.ai.machinelearning as winml 
+import winrt.windows.ai.machinelearning as winml 
 import os
 
 @timed_op
 def load_model(model_path):
-    return winml.LearningModel.LoadFromFilePath(os.fspath(model_path))
+    return winml.LearningModel.load_from_file_path(os.fspath(model_path))
```
kennykerr: Here the class appears to have the right casing but the method again has a different casing. Seems rather confusing.

src/tool/python/code_writers.h
```diff
@@ -1,60 +1,92 @@
...
+        XLANG_ASSERT(name.size() > 0);
+
+        for (std::string_view::size_type i = 0; i < name.size(); i++)
         {
-            return;
+            if (isupper(name[i]) && i > 0)
+            {
+                //TODO: special case IPAddress so that it doesn't get emitted as i_p_address
+                auto sub = name.substr(i - 1, 4);
+                if (sub != "UInt" && sub[0] != '_')
```
kennykerr: Seems you're going to need a lot of special casing for edge cases.

src/tool/python/code_writers.h
```diff
@@ -72,638 +104,283 @@ namespace xlang
...
             }
 
-            void handle(GenericTypeInstSig const& type)
-            {
-                handle(get_ireference_type(type));
-            }
+            w.write(R"(}
+catch(...)
+{
+    return -1;
```
kennykerr: This looks suspicious. Why are we throwing away any trace of the error.

src/test/python/CMakeLists.txt
devhawk: so I can easily uncomment them to run the complete_winml_tutorial test

kennykerr:
This PR is just way too big to adequately review, especially since CodeFlow doesn't work with so many commits. The GitHub view is just *not* conducive to large PRs.

src/test/python/complete_winml_tutorial.py
devhawk: lowercase package names is python's convention. #201

src/test/python/complete_winml_tutorial.py
devhawk: Again, python convention for members is lower_snake_case as per #201

src/test/python/complete_winml_tutorial.py
```diff
@@ -29,42 +29,42 @@ def sync_wrapper(*args, **kwds):
...
+    file = await StorageFile.get_file_from_path_async(os.fspath(file_path))
+    stream = await file.open_async(FileAccessMode.READ) 
+    decoder = await BitmapDecoder.create_async(stream)
+    software_bitmap = await decoder.get_software_bitmap_async()
+    return VideoFrame.create_with_software_bitmap(software_bitmap)
 
 @timed_op
 def bind_model(model, image_frame):
-    device = winml.LearningModelDevice(winml.LearningModelDeviceKind.Default)
+    device = winml.LearningModelDevice(winml.LearningModelDeviceKind.DEFAULT)
```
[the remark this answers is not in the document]
devhawk: Yup. Enum members are UPPER_SNAKE_CASED

jonwis:
> This PR is just way too big to adequately review, especially since CodeFlow doesn't work with so many commits. The GitHub view is just _not_ conducive to large PRs.

Maybe create an intermediate branch between master and this one, then stage PRs into it, and then one final PR into master?

src/tool/python/code_writers.h
```diff
@@ -1,60 +1,92 @@
...
     {
-        if (is_exclusive_to(type))
+        XLANG_ASSERT(name.size() > 0);
+
+        for (std::string_view::size_type i = 0; i < name.size(); i++)
         {
-            return;
+            if (isupper(name[i]) && i > 0)
+            {
+                //TODO: special case IPAddress so that it doesn't get emitted as i_p_address
```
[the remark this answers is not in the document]
devhawk: THanks for catching this. I think I'll fix it as part of this PR rather than file a bug

src/test/python/complete_winml_tutorial.py
kennykerr: Sure, I get that Python has that convention just like C++ has the convention of using lower_case for types and members. But that does not mean that it is a requirement. It's going to get rather difficult to avoid having a growing list of exceptions as you deal with various APIs that you you need to special case. It is also lossy in the sense that you cannot readily roundtrip the name back to the original. C++/WinRT takes the approach that while C++ libraries tend to use lower_case, API authors choose what to name their APIs and how they are cased. This is far less problematic. Anyway, just an observation.

src/tool/python/code_writers.h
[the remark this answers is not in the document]
devhawk: fixed

src/tool/python/code_writers.h
devhawk: I don't think so, but I updated the code to support easily adding more exceptions

src/tool/python/code_writers.h
devhawk: yup. Adding py::to_PyErr(); statement

PR #256, opened 2019-03-29
xmeta design notes draft

timr1126:
This is an initial draft of the design notes. Will need to elaborate, but the basics of xmeta are there.

### DIFF
--- a/design_notes/XDN13 - xmeta Architecture and Purpose.md
+++ b/design_notes/XDN13 - xmeta Architecture and Purpose.md
@@ -1,0 +1,33 @@
+---
+id: XDN13
+title: xmeta Architecture and Purpose
+author: t-tiroma@microsoft.com
+status: draft
+---
+
+# Title: XDN13 - xmeta Architecture and Purpose
+
+- Author: Tim Romanski (t-tiroma@microsoft.com)
+- Status: Draft
+
+## Abstract
+
+This document will outline the purpose of xmeta and the design and architecture decisions made for the tool.
+
+## Purpose
+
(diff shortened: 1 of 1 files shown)


design_notes/XDN13 - xmeta Architecture and Purpose.md
```diff
@@ -0,0 +1,33 @@
...
+
+## Abstract
+
+This document will outline the purpose of xmeta and the design and architecture decisions made for the tool.
+
+## Purpose
+
+The xmeta tool will be a cross platform version of MIDLRT. As a brief sidenote, MIDLRT was a fork of MIDL with a very verbose syntax. A new syntax called MIDL3 was created, and it was added to MIDLRT rather than having its own tool built from scratch. Unfortunately, there is now a lot of hardcoded WinRT knowledge in MIDLRT that would be difficult to adapt for xlang.
+
+Functionally, xmeta will take in a MIDL3 [IDL](https://docs.microsoft.com/en-us/windows/desktop/com/idl-files) file and previously compiled `.winmd` files, and will compile them into new `.winmd` metadata files.
```
BenJKuhn: General question: are we keeping the same extension for xlang metadata? That seems potentially troublesome.

timr1126: We will be renaming the extension. Probably to `.xmeta`.

jkoritzinsky: Also, will there be a bit set in the metadata to differentiate between a WinMD and an XMeta file like how there's currently the `tdWindowsRuntime` bit to differentiate a WinRT class and a .NET class?

BenJKuhn: We've discussed this bit from time to time, but haven't made a decision yet. One key data point is that although we're leveraging the .NET metadata format to describe types, we expect these will likely NOT be read by the CLR directly the way winrt types are. The xlang C# projection will likely be independent of the .NET core code, which makes this questions a bit less critical.

jkoritzinsky: Thanks for the response! That's good to know!

PR #259, opened 2019-03-30
Defaultryan/cppx2

DefaultRyan:
At this point we have a (barely) functional xlang projection. Overhauling the strings and error handling was a big chunk of the work. I've renamed the include path prefix from "winrt" to "xlang" to prevent header collisions and be sure I'm actually testing the xlang headers and not the cppwinrt headers. The mass namespace rename will happen later in a mechanical commit - that much noise would completely obscure the functional changes in this PR.

At this stage, we largely have parity with cppwinrt's hstring, but the dual-encoding design has caused some complexity. Coroutines and async are pretty much absent for non-Windows - some investigation remains to be done to see if we need to add some PAL support for that sort of thing. We have a bare-bones error handler, as that design is still in flux, and I wanted to avoid boiling the ocean on that until we've landed a design. 

I'm continuing to port the hstring tests from cppwinrt first, as that's the code that saw the most churn. Eventually, most of the unit tests from cppwinrt will move over. I believe that the implements template and all the component authoring support is still almost identical to cppwinrt, but we don't yet have any metadata with which to generate an actual type projection, so I haven't tackled that test coverage yet.

One interesting side effect of the string overhaul was that I needed to choose a "default" encoding for the strings the compiler generates (type names, etc). I went with utf-8, and as a result, I was able to remove the chunk of pinterface calculation code responsible for utf-16 to utf-8 conversion.

Overall, there are still quite a few relics of the WinRT type system in place, but those aspects will evolve as we finalize the xlang type system. The top priority for this phase was to get a C++ projection and base library that could target the PAL instead of Windows-specific functions.


src/tool/cppxlang/strings/base_error.h
```diff
@@ -392,6 +273,20 @@ namespace winrt
         }
     }
 
+	inline void check_hresult(xlang_error_info* result)
+	{
+		if (result != nullptr)
+		{
+			throw_hresult(result->error_code());
```
kennykerr: tabs?

src/tool/cppxlang/strings/base_meta.h
```diff
@@ -1,8 +1,9 @@
 
 namespace winrt
 {
-    void check_hresult(hresult const result);
-    hresult to_hresult() noexcept;
+	void check_hresult(hresult const result);
```
kennykerr: tabs?

DefaultRyan:
> Weird tab/space inconsistency. Otherwise looks good.

Thanks for catching that. Not sure how it happened. Visual C++ resetting my tabs settings on me for some reason, I suspect. 😐

Issue #248, opened 2019-03-25
move src/tool/idl into samples


devhawk:
The IDL tool was used as a testbed during the original creation of the metadata reader. It's still has value as an example of how to use the metadata reader, so we shouldn't delete it.

jonitis:
This can be closed now. Thanks for accepting the patch

PR #260, opened 2019-03-31
Fix indentation for idl output (Take 2)

jonitis:
This is followup to PR https://github.com/Microsoft/xlang/pull/227. I've closed original mostly because it was based on changes in master branch (and not feature branch).

I was looking at output of idl and noticed that custom attributes for methods had wrong indentation of 4 spaces when method itself had 8 space indentation. Like this:

```
    interface ILearningModelStatics
    {
    [Windows.Foundation.Metadata.RemoteAsyncAttribute]
    [Windows.Foundation.Metadata.OverloadAttribute("LoadFromStorageFileAsync")]
        Windows.Foundation.IAsyncOperation`1<LearningModel> LoadFromStorageFileAsync(Windows.Storage.IStorageFile modelFile); 
```
vs.
```
    interface ILearningModelStatics
    {
        [Windows.Foundation.Metadata.RemoteAsyncAttribute]
        [Windows.Foundation.Metadata.OverloadAttribute("LoadFromStorageFileAsync")]
        Windows.Foundation.IAsyncOperation`1<LearningModel> LoadFromStorageFileAsync(Windows.Storage.IStorageFile modelFile); 
```

Use same approach as @devhawk python tool that uses indent_guard class to automatically increase indentation in scope and specialize writer::write_impl() methods to automatically indent output for each line that starts on new line.

- Now write(CustomAttribute const& attr) and write_uuid(CustomAttributeSig const& arg) are similar to other functions that they do not have enforced newline and indent at the start of line.
- All manual loops that wrote custom attributes are replaced with bind_each<write_custom_attribute>(attribute)
- Compared to original python tool code the writer::write_impl() methods check if current line to output is equal to newline. If it is then there is no sense to indent it. 

I verified that idl output for windows.winmd differs only by whitespace changes for fixed custom attributes. 

If this approach is acceptable then next step is to factor out:
- struct indent_guard
- int32_t indent{};
- void write_impl(std::string_view const& value)
- void write_impl(char const value)
- std::string write_temp(std::string_view const& value, Args const& ... args)

into indented_writer class so that python and idl tools can use shared code. I also have not moved source to samples directory, since there is separate ticket https://github.com/Microsoft/xlang/issues/248 for it.


jonitis:
Windows build failed to create Docker container :(

devhawk:
re-queued the build. we've seen some flakiness from pipelines docker support recently. Unrelated to this PR

devhawk:
@jonitis I asked on #227 if you'd be willing to move the entire IDL project into the samples directory. I'm thinking /samples/meta_reader. Can you do that as part of this PR?

jonitis:
Ok, I will try to move it and modify build files.

jonitis:
> 
> 
> @jonitis I asked on #227 if you'd be willing to move the entire IDL project into the samples directory. I'm thinking /samples/meta_reader. Can you do that as part of this PR?

I moved code to samples directory, but I am not sure what to do with CMakeLists.txt file. I've renamed idl->meta_reader, but should it include master Cmake file from src directory?

devhawk:
I think we should add a CMakeLists file in /samples that includes a `add_subdirectory(meta_reader)`. We should also exclude the IDL project from build all with a `set_target_properties(idl PROPERTIES EXCLUDE_FROM_ALL TRUE)`

samples/meta_reader/main.cpp
```diff
@@ -47,6 +66,75 @@ struct writer : writer_base<writer>
...
+    }
+
+    void write_impl(std::string_view const& value)
+    {
+        std::string_view::size_type current_pos{ 0 };
+        auto on_new_line = back() == '\n';
+
+        while (true)
+        {
+            auto pos = value.find('\n', current_pos);
```
ChrisGuzak: `const auto pos = …`

samples/meta_reader/main.cpp
```diff
@@ -47,6 +66,75 @@ struct writer : writer_base<writer>
...
+                    writer_base::write_impl(value.substr(current_pos));
+                }
+
+                return;
+            }
+
+            auto current_line = value.substr(current_pos, pos - current_pos + 1);
+            auto empty_line = current_line[0] == '\n';
+
+            if (on_new_line && !empty_line) {
```
ChrisGuzak: move opening `{` to its own line.

samples/meta_reader/main.cpp
```diff
@@ -47,6 +66,75 @@ struct writer : writer_base<writer>
...
+        if (back() == '\n' && value != '\n')
+        {
+            write_indent();
+        }
+
+        writer_base::write_impl(value);
+    }
+
+    template <typename... Args>
+    std::string write_temp(std::string_view const& value, Args const& ... args)
```
ChrisGuzak: pass `string_view` by value

samples/meta_reader/main.cpp
jonitis: fixed

samples/meta_reader/main.cpp
jonitis: fixed

samples/meta_reader/main.cpp
jonitis: Not fixed.
Base class struct writer_base at src/library/text_writer has string_view passed by const reference. I just copied implementation from src/tool/python/type_writers.h. And all other code in this repository always pass string_views as const references and not by copy/value. I prefer it too, since it is more efficient to pass 64-bit pointer in one register, rather than use two registers for pointer and size. Especially if same string_view is passed down multiple functions deep.

kennykerr: Yep, consistency trumps here. The xlang repo uses `const&` wherever possible. It is also more efficient as you mentioned.

jonitis:
I addressed review notes related to c++ code that implements indentation.
But I have problems with Cmake files and moving code to sample directory. I've added EXCLUDE_FROM_ALL for meta_reader target so that it is not built by default. But I can't get it build with --target samples. Cmake complains that it does not know TARGET_CONFIG_MSVC_PCH. So it does not inherit from src/CmakeLists.txt. Probably because it tries to reference parent directory with .. 
It would be better to add top level CmakeLists.txt that would handle both src and samples directories, but that would affect all build scripts for CI etc.
Someone better familiar with your build system should complete that migration to samples directory.

P.S.
BTW I would prefer if idl stayed at tools directory. It had self-explanatory name and it was real tool :)

devhawk:
I'll take a look at your cmake files and propose a fix. 

/src/tool/idl was never a real tool - it was only a testbed for the metadata reader. We are working on a tool for generting metadata from IDL (#262, #264, and #256 all relate to this tool)

src/CMakeLists.txt
```diff
@@ -130,3 +130,4 @@ add_subdirectory(tool)
 add_subdirectory(test)
 add_subdirectory(platform)
 add_subdirectory(package)
+add_subdirectory(../samples ${CMAKE_CURRENT_BINARY_DIR}/samples)
```
devhawk: Doh! I totally forgot that samples is a peer to src and doesn't share a root w/ a CMakeLists file in it. Let me circle w/ the other devs and get back to you with a new recommendation.

devhawk: I'm interested what @DefaultRyan, @kennykerr  and @Scottj1s think, but I'm thinking we should *NOT* include samples in the /src/cmakelists.txt file. We shouldn't have a root /samples/cmakelists.txt file either. Just whatever the sample needs to build local to the sample folder. For this sample, that does mean having to set the XLANG_LIBRARY_PATH + compiler settings directly in this file rather than inheriting them from root cmake file. But I think it's better for samples to be isolated. Better to have all those definitions self contained for people attempting to learn from the sample than having to dig thru other cmake files to find them.

kennykerr: Agreed.

jonitis: Made meta_reader cmake file self-contained.
```
cmake ./meta_reader -B_build -GNinja && ninja -C _build -v
```

devhawk:
Looks great! I made a slight change to the cmake file - mirroring #290 - to account for CMake 3.13 (which ships w/ VS2019).

PR #264, opened 2019-04-02
Symbol table models

timr1126:
Models to be used for the symbol tables. These include models for namespaces, classes, structs, enums, interfaces, and their members. Members include delegates, events, methods and properties.


src/tool/xmeta/cpp/models/base_model.h
```diff
@@ -0,0 +1,62 @@
...
+{
+    namespace xmeta
+    {
+        class base_model
+        {
+        public:
+            std::string id;
+            size_t decl_line;
+
+            base_model(const std::string &id, const size_t &decl_line) : id{ id }, decl_line{ decl_line } { }
```
ChrisGuzak: consider `std::string_view id`. no reason to pass size_t by `const &`

src/tool/xmeta/cpp/models/base_model.h
```diff
@@ -0,0 +1,62 @@
...
+
+        protected:
+            base_model() = delete;
+
+        };
+
+        enum value_t
```
ChrisGuzak: `enum class`

src/tool/xmeta/cpp/models/base_model.h
```diff
@@ -0,0 +1,62 @@
...
+        class base_model
+        {
+        public:
+            std::string id;
+            size_t decl_line;
+
+            base_model(const std::string &id, const size_t &decl_line) : id{ id }, decl_line{ decl_line } { }
+
+        protected:
+            base_model() = delete;
```
ChrisGuzak: why is this protected? seems being public and deleted is better (simpler)

src/tool/xmeta/cpp/models/class_model.h
```diff
@@ -0,0 +1,46 @@
...
+        };
+
+        class class_model final : public base_model
+        {
+        public:
+            class_model(const std::string &id, const size_t &decl_line) : base_model{ id, decl_line } { }
+
+            std::string class_base_id;
+            std::shared_ptr<class_model> class_base;
+            std::vector<std::string> interface_base_list;
```
ChrisGuzak: consider using plural forms instead of `_list` as a suffix, `interface_bases`, etc.

src/tool/xmeta/cpp/models/class_model.h
```diff
@@ -0,0 +1,46 @@
...
+            std::vector<std::string> type_parameter_list;
+            std::vector<formal_parameter_t> formal_parameter_list;
+
+            // Members
+            std::vector<std::shared_ptr<event_model>> events;
+            std::vector<std::shared_ptr<method_model>> methods;
+            std::vector<std::shared_ptr<property_model>> properties;
```
ChrisGuzak: PME (properties, methods, events), consider re-ordering based on that.

src/tool/xmeta/cpp/models/enum_model.h
```diff
@@ -0,0 +1,44 @@
...
+
+#include <string>
+
+#include "base_model.h"
+
+namespace xlang
+{
+    namespace xmeta
+    {
+        enum enum_type
```
ChrisGuzak: `enum class`

src/tool/xmeta/cpp/models/namespace_model.cpp
```diff
@@ -0,0 +1,59 @@
+#include "namespace_model.h"
+
+namespace xlang
+{
+    namespace xmeta
+    {
+#pragma region namespace_body_model
+        bool namespace_body_model::member_id_exists(const std::string &member_id)
```
ChrisGuzak: `std::string_view` for input string values.

src/tool/xmeta/cpp/models/base_model.h
```diff
@@ -0,0 +1,62 @@
+#pragma once
+
+#include <string>
+
+namespace xlang
+{
+    namespace xmeta
+    {
+        class base_model
```
kennykerr: Use `struct` rather than `class` for brevity.

src/tool/xmeta/cpp/models/base_model.h
```diff
@@ -0,0 +1,62 @@
+#pragma once
+
+#include <string>
+
+namespace xlang
```
kennykerr: Use nested namespaces. https://en.cppreference.com/w/cpp/language/namespace

src/tool/xmeta/cpp/models/class_model.h
```diff
@@ -0,0 +1,46 @@
...
+    {
+        // Note: All classes are implicitly public, and explicitly specifying this is not allowed.
+        enum class_modifier_t
+        {
+            none,
+            sealed_class,
+            static_class
+        };
+
+        class class_model final : public base_model
```
kennykerr: `struct` rather than `class` to simplify the accessibility. Also why `final`?

src/tool/xmeta/cpp/models/class_model.h
```diff
@@ -0,0 +1,46 @@
...
+            class_model(const std::string &id, const size_t &decl_line) : base_model{ id, decl_line } { }
+
+            std::string class_base_id;
+            std::shared_ptr<class_model> class_base;
+            std::vector<std::string> interface_base_list;
+            std::vector<std::string> type_parameter_list;
+            std::vector<formal_parameter_t> formal_parameter_list;
+
+            // Members
+            std::vector<std::shared_ptr<event_model>> events;
```
kennykerr: Why `shared_ptr`? Are you planning on sharing them?

src/tool/xmeta/cpp/models/class_model.h
```diff
@@ -0,0 +1,46 @@
...
+            std::vector<std::shared_ptr<method_model>> methods;
+            std::vector<std::shared_ptr<property_model>> properties;
+
+        protected:
+            class_model() = delete;
```
kennykerr: There's no value in making these protected.

src/tool/xmeta/cpp/models/base_model.h
```diff
@@ -0,0 +1,62 @@
...
+            guid_type,
+            single_type,
+            double_type,
+        };
+
+        struct xmeta_type
+        {
+            bool is_void;
+            bool is_array;
+            bool is_class; // if (!is_class) { return type is value type (string, int16, int32, etc...) }
```
DefaultRyan: What about structs and interfaces? Perhaps a better distinction would be more like "is_fundamental".

src/tool/xmeta/cpp/models/enum_model.h
```diff
@@ -0,0 +1,44 @@
...
+        {
```
DefaultRyan: enums can also use bool as their underlying type,

src/tool/xmeta/cpp/models/class_model.h
timr1126: Not those. Those shared_ptr's can be removed.

src/tool/xmeta/cpp/models/event_model.h
```diff
@@ -0,0 +1,37 @@
...
+#include "base_model.h"
+
+namespace xlang
+{
+    namespace xmeta
+    {
+        enum event_modifier_t
+        {
+            no_event_modifier,
+            proected_event,
```
kythant: Typo: proected_event

src/tool/xmeta/cpp/models/interface_model.h
```diff
@@ -0,0 +1,34 @@
...
+namespace xlang
+{
+    namespace xmeta
+    {
+        class interface_model final : public base_model
+        {
+        public:
+            interface_model(const std::string &id, const size_t &decl_line) : base_model{ id, decl_line } { }
+
+            std::vector<std::string> interface_base_list;
```
kythant: Don't forget to keep it consistant ***_list to ***s per Chris's recommendation.

src/tool/xmeta/cpp/models/namespace_model.cpp
```diff
@@ -0,0 +1,59 @@
...
+        {
+            if (classes.find(member_id) != classes.end())
+                return true;
```
kythant: Nit: The style convention from OSG is that you must have curly brackets and you cannot omit them here.

src/tool/xmeta/cpp/models/namespace_model.cpp
```diff
@@ -0,0 +1,59 @@
...
+            for (auto ns_body : ns_bodies)
+            {
+                if (ns_body->member_id_exists(member_id))
+                {
+                    return true;
+                }
+            }
+
+            if (child_namespaces.find(member_id) != child_namespaces.end())
+                return true;
```
kythant: nit: curly brackets

src/tool/xmeta/cpp/models/namespace_model.cpp
```diff
@@ -0,0 +1,59 @@
...
+
+            return false;
+        }
+
+        std::string namespace_model::get_full_namespace_name()
+        {
+            if (parent_namespace != nullptr)
+                return parent_namespace->get_full_namespace_name() + "." + id;
```
kythant: and here

src/tool/xmeta/cpp/models/namespace_model.h
```diff
@@ -0,0 +1,70 @@
...
+
+
+namespace xlang
+{
+    namespace xmeta
+    {
+        struct namespace_body_model;
+        class namespace_model;
+
+        typedef std::vector<std::shared_ptr<namespace_model>> namespace_vector;
```
kythant: Is typedef really necessary here? How is it different from the other structs?

ChrisGuzak: +1. type aliases hide the concepts (a vector of models), harming legibility.

src/tool/xmeta/cpp/models/enum_model.h
devhawk: This seems wrong. What is the value of an enum backed by a bool?

ChrisGuzak: strong typing: https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#Ri-typed

BrentRe: Depends on the language. ECMA 335 CLI spec says:

II.14.3 Enums 
The symbols of an enum are represented by an underlying integer type:  one of { bool, char, int8, unsigned int8, int16, unsigned int16, int32, unsigned int32, int64, unsigned int64, native int, unsigned native int }

However, ECMA 334 C# specification says:

9.3.10 Enumeration types An enumeration type is a distinct type with named constants. Every enumeration type has an underlying type, which shall be byte, sbyte, short, ushort, int, uint, long or ulong.

Personally, I don't see much value in allowing a enum to have a bool underlying base type. I could vaguely see a rationale for char, but then we'd have trouble projecting such an enum into C#.

I'd vote for restricting to the C# allowed underlying types.

src/tool/xmeta/cpp/models/namespace_model.cpp
```diff
@@ -1,59 +1,69 @@
 #include "namespace_model.h"
 
-namespace xlang
+namespace xlang::xmeta
 {
-    namespace xmeta
-    {
 #pragma region namespace_body_model
-        bool namespace_body_model::member_id_exists(const std::string &member_id)
+    bool namespace_body_model::member_id_exists(std::string_view const& member_id)
```
kythant: You could make this a one liner using ||

return classes.find(member_id) != classes.end() || structs.find(member_id) != structs.end() || interfaces.find(member_id) != interfaces.end()

src/tool/xmeta/cpp/models/namespace_model.cpp
```diff
@@ -1,59 +1,69 @@
...
-                }
-            }
-
-            if (child_namespaces.find(member_id) != child_namespaces.end())
                 return true;
+            }
+        }
 
-            return false;
+        if (child_namespaces.find(member_id) != child_namespaces.end())
```
kythant: nit: Boolean zen, just use a return statement instead of "if true return true else return false" 

return child_namespaces.find(member_id) != child_namespaces.end();

src/tool/xmeta/cpp/models/namespace_model.cpp
```diff
@@ -0,0 +1,51 @@
...
+    }
+
+    std::string namespace_body_model::get_full_namespace_name()
+    {
+        return containing_namespace->get_full_namespace_name();
+    }
+#pragma endregion
+
+
+#pragma region namespace_model
```
kennykerr: I think this pragma is VS specific and not appropriate for a portable library. It's also not very beneficial in such a small file.

src/tool/xmeta/cpp/models/enum_model.h
DefaultRyan: Like Chris was saying, bool-backed enums are incredibly valuable for improving usability, readability, and correctness. As an example from Win32: `auto event = CreateEvent(nullptr, FALSE, FALSE, "Name");`. Let's change this to an auto-reset event: `auto event = CreateEvent(nullptr, FALSE, TRUE, "Name");`.

See the bug?

I accidentally changed the initial state, not the reset behavior. With vague bool parameters, it's too easy to forget what those two bools mean, and too easy to accidentally change the wrong one. This interface is far more usable and robust: `auto event = CreateEvent(nullptr, reset_behavior::auto, initial_state::non_signalled, "name");`.

src/tool/xmeta/cpp/models/class_model.h
```diff
@@ -0,0 +1,40 @@
...
+
+    struct class_model : base_model
+    {
+        class_model(std::string_view const& id, size_t decl_line) : base_model{ id, decl_line } { }
+        class_model() = delete;
+
+        std::string_view class_base_id;
+        std::shared_ptr<class_model> class_base;
+        std::vector<std::string_view> interface_bases;
+        std::vector<std::string_view> type_parameters;
```
DefaultRyan: It's not immediately obvious to me how type_parameters, formal_parameters, and generic_types interact.

BrentRe:
You don’t need enums with a bool underlying type for this, however. You just need a 0/1 integral enum.

That enums with a bool underlying type cannot be projected to C# seems like a detriment.

src/tool/xmeta/cpp/models/class_model.h
timr1126: Ah, generic_types are the same as type_parameters and can be removed. formal_parameters are for methods and delegates, it shouldn't even be in this file. I'll remove the two (generic_types and formal_parameters).

PR #265, opened 2019-04-02
[Draft] .d.ts generation tool

Tomas-Morris:
The generation of .d.ts definitions from winmd would be (and has been) a very useful tool as Windows apps allow the projection of such types into Javascript. This is of special interest to my team, so I have spent some cycles on making it work.

This draft is to verify the approach I am taking and to get guidance and suggestions on creating a well-built comprehensive tool for this job. 

Thanks


devhawk:
I'd recommend *NOT* including all the base_*.h files in your tool. If you're just creating .d.ts files, you won't need all of that stuff that is very specific to the C++ projection. Take a look at the IDL sample (currently in /src/tool/idl but moving to samples directory as part of #260)

Scottj1s:
Aside from cleaning up the help strings and settings class (which still have cppwinrt references), it looks like an idiomatic projection tool.  Hard to tell without seeing the generated .d.ts files, but I assume they're to your satisfaction.

ChrisGuzak:
@Tomas-Morris be sure to check http://osgvsowi/19923012

kennykerr:
Hey @Tomas-Morris, what's the status of this PR?

Tomas-Morris:
This PR is now stale as I have been on paternity leave. I will take this back up to clean up and make a true PR. I would still like to have some way to verify that the tool is comprehensive, as I can only verify the .winmd from my and my related teams.

PR #278, opened 2019-04-03
Partial fast ABI support for C++/WinRT

kennykerr:
This is the first stage of the fast ABI support for C++/WinRT. It is not complete, but I want to get what's done checked in to avoid this PR getting hopelessly too big. The bulk of the changes here are actually additional test code. There are also a few small bug fixes. Notables:

* I've renamed a few of the projects under /cppwinrt to make it easier to manage a growing set of test projects. 

* I fixed a bug in the RC generation where the version number shows up incorrectly.

* I hardened the WinRT delegate code gen to deal with out params reliably. This is just proactive work. I have not run into any cases of this in the wild, but it simplifies the code gen not having special code to deal with delegates.

* The fast ABI work itself is the basic shell and works for consumption and production. Consumption of the fast ABI works for both fastabi-aware consumers and non-fastabi-aware consumers. Callers must opt in to the fast ABI using the `-fastabi` command line option. This is consistent with MIDLRT and makes sense since this is a breaking change (for consumers). 

The next steps are to add more comprehensive testing for the fast ABI and ensure that it lines up with Duncan's implementation. This will naturally lead to more tuning of the implementation. But what we have here is enough for the Composition team to start experimenting, which is critical to get them unblocked and begin their adoption process. I then need to work with Scott on integrating the fast forwarder support.


src/tool/cppwinrt/cppwinrt/main.cpp
```diff
@@ -35,6 +36,8 @@ namespace xlang
         { "filter" }, // One or more prefixes to include in input (same as -include)
         { "license", 0, 0 }, // Generate license comment
         { "brackets", 0, 0 }, // Use angle brackets for #includes (defaults to quotes)
+        { "fastabi", 0, 0 }, // Enable support for the Fast ABI
```
dunhor: Out of curiosity, is there any reason to have an explicit switch for this?

kennykerr: Yes. Imagine an app that wants to run everywhere from Windows 8 to Windows 10 and doesn't want to deal with conditionally supporting the fast ABI. It just wants to work everywhere with a single binary. That developer won't want to use `-fastabi`.

dunhor: Heh, I guess there's no best option no matter how you spin it. I was thinking about the producer scenario where it would be rather negative to not specify the switch.

kennykerr: Right, the produce would *always* want to specify `-fastabi`. The consumer may not.

src/tool/cppwinrt/cppwinrt/helpers.h
```diff
@@ -246,17 +268,78 @@ namespace xlang
...
+
+    static void insert_or_assign(get_interfaces_t& interfaces, std::string_view const& name, interface_info&& info)
+    {
+        if (auto existing = find(interfaces, name))
+        {
+            *existing = std::move(info);
+        }
+        else
+        {
+            interfaces.emplace_back(name, info);
```
dunhor: move?

kennykerr: Yes! 👍

Scottj1s:
>         { "opt", 0, 0, {}, "Generate component projection with unified construction support" },

(unrelated) consider spelling out "optimized" and "library" since the cmd reader already support substring matches 

---
Refers to: src/tool/cppwinrt/cppwinrt/main.cpp:33 in 923e0d6. [](commit_id = 923e0d6bcf78279bd90cbadea5510d09eb982647, deletion_comment = False)

src/tool/cppwinrt/cppwinrt/main.cpp
```diff
@@ -35,6 +36,8 @@ namespace xlang
...
+        { "tests", 0, 0 }, // Generate internal test files
         { "ignore_velocity", 0, 0 }, // Ignore feature staging metadata and always include implementations
```
Scottj1s: no way to isolate this from the production exe?

src/tool/cppwinrt/cppwinrt/main.cpp
```diff
@@ -198,6 +203,36 @@ Where <spec> is one or more of:
...
+            for (auto&& type : members.classes)
+            {
+                if (!has_fastabi(type))
+                {
+                    continue;
+                }
+
+                auto default_interface = get_default_interface(type);
+
+                if (default_interface.type() == TypeDefOrRef::TypeDef)
```
Scottj1s: >default_interface [](start = 20, length = 17)

is it possible for a default interface to be generic?

src/tool/cppwinrt/cppwinrt/settings.h
```diff
@@ -22,13 +24,14 @@ namespace xlang
...
-        bool verbose{};
-
         std::set<std::string> include;
         std::set<std::string> exclude;
 
         meta::reader::filter projection_filter;
         meta::reader::filter component_filter;
+
+        bool fastabi{};
+        std::map<meta::reader::TypeDef, meta::reader::TypeDef> fastabi_defaults;
```
Scottj1s: >fastabi_defaults [](start = 63, length = 16)

nit: kind of a funny name (sounds like "default settings for fastabi")

src/tool/cppwinrt/cppwinrt/main.cpp
kennykerr: Yes, I've been thinking about eliding this somehow.

src/tool/cppwinrt/cppwinrt/main.cpp
kennykerr: Yes, although then it would not be exclusive and thus not eligible for the fast ABI.

src/tool/cppwinrt/cppwinrt/settings.h
kennykerr: Yes, naming is hard. Maybe just `fastabi_cache` since the function that creates it is called `build_fastabi_cache`.

src/tool/cppwinrt/test_component_fast/test_component_fast.idl
```diff
@@ -0,0 +1,157 @@
+import "Windows.Foundation.idl";
+
+namespace Windows.Foundation.Metadata
+{
+    [attributeusage(target_method, target_property)]
+    [attributename("noexcept")]
+    attribute NoExceptionAttribute
+    {
+    }
```
Scottj1s: this isn't in midlrt yet?

kennykerr: Only in APT7's version of MIDLRT - the build server won't have it for a while.

src/tool/cppwinrt/cppwinrt/helpers.h
```diff
@@ -246,17 +268,78 @@ namespace xlang
...
         bool base{};
+        bool exclusive{};
+        bool fastabi{};
+        std::pair<uint16_t, uint16_t> version{};
         std::vector<std::vector<std::string>> generic_param_stack{};
     };
 
-    static void get_interfaces_impl(writer& w, std::map<std::string, interface_info>& result, bool defaulted, bool overridable, bool base, std::vector<std::vector<std::string>> const& generic_param_stack, std::pair<InterfaceImpl, InterfaceImpl>&& children)
+    using get_interfaces_t = std::vector<std::pair<std::string, interface_info>>;
+
```
Scottj1s: I *really* like this pattern.  consider using it more liberally (e.g., for get_version above)

kennykerr: 👍

src/tool/cppwinrt/test_slow/Composition.cpp
```diff
@@ -0,0 +1,42 @@
+#include "pch.h"
+#include "winrt/test_component_fast.Composition.h"
+
+using namespace winrt::test_component_fast::Composition;
+
+// All, yes all, of the QIs in the test below are avoided by the fastabi.
```
Scottj1s: impressive! :smile:

PR #282, opened 2019-04-04
Py/WinRT - changes to support C++/WinRT 2.0

devhawk:
* changes to enable compiling python projection with /permissive-
* update setup.py with VS2019 settings, /permissive- and cppwinrt include folder
* -input sdk support
* change namespace generation logic to skip namespaces using same logic as C++/WinRT
* removed string files that are no longer used

*NOTE* this change does NOT affect cmake build files for python tests. The build process will be changing significantly to enable publishing the python projection binary. I'll be making those build system changes as part of the publication work.


Scottj1s:
sorry I missed the PR windows on this!

PR #283, opened 2019-04-04
Update msbuild invocations to allow RelWithDebInfo/MinSizeRel

dunhor:
Right now it's not possible to build with just CMake+Ninja and get a release build with symbols (at least not without scoping the build down to specific target(s)) since none of the vcxproj/csproj files allow these as valid configurations. This updates msbuild invocations to translate RelWithDebInfo/MinSizeRel into 'Release' when invoking msbuild.

### DIFF
--- a/src/package/natvis/cppwinrtvisualizer.vcxproj
+++ b/src/package/natvis/cppwinrtvisualizer.vcxproj
@@ -56,6 +56,7 @@
   <Import Project="$(VCTargetsPath)\Microsoft.Cpp.props" />
   <PropertyGroup>
     <DIASDKInc>$(VSInstallDir)DIA SDK\include</DIASDKInc>
+    <CmakeConfiguration Condition="'$(CmakeConfiguration)'==''">$(Configuration)</CmakeConfiguration>
   </PropertyGroup>
   <ImportGroup Label="ExtensionSettings">
   </ImportGroup>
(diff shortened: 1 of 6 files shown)


src/package/vsix/CMakeLists.txt
```diff
@@ -14,7 +14,7 @@ file(TO_NATIVE_PATH "${cppwinrt_nupkg_dir}/Microsoft.Windows.CppWinRT.${XLANG_BU
 
 file(DOWNLOAD https://dist.nuget.org/win-x86-commandline/latest/nuget.exe ${nuget_exe})
 
-set(build_vsix msbuild ${CMAKE_CURRENT_SOURCE_DIR}/vsix.csproj /nologo /m /p:Configuration=${CMAKE_BUILD_TYPE},Platform=$ENV{VSCMD_ARG_TGT_ARCH})
+set(build_vsix msbuild ${CMAKE_CURRENT_SOURCE_DIR}/vsix.csproj /nologo /m /p:Configuration=${MSBUILD_CONFIGURATION},Platform=$ENV{VSCMD_ARG_TGT_ARCH})
```
dunhor: Blarg. Also need to pass in CmakeBuildType here as well since the pipeline uses build configuration for paths

src/package/natvis/cppwinrtvisualizer.vcxproj
```diff
@@ -76,23 +77,23 @@
   <PropertyGroup Label="UserMacros" />
   <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|Win32'">
     <LinkIncremental>true</LinkIncremental>
-    <OutDir>x86\$(Configuration)\</OutDir>
-    <IntDir>x86\$(Configuration)\</IntDir>
+    <OutDir>x86\$(CmakeConfiguration)\</OutDir>
```
Scottj1s: how does this help?  aren't the visualizer and vsix projects still building into RelWithDebInfo, which breaks the azdo pipeline?

kennykerr: I thought we'd just use the names `Debug` and `Release` throughout.

dunhor:
Per conversation, going to abandon this and pursue a solution to enable symbol generation by default for release builds

src/package/natvis/cppwinrtvisualizer.vcxproj
dunhor: In this case, `CmakeConfiguration` can be configured by CMake, which would be set to `RelWithDebInfo`, which would match what ADO is expecting. All moot now, but that was the idea

Scottj1s: oh moot now, got it

PR #285, opened 2019-04-04
Py/WinRT: Size Optimization

devhawk:
A set of changes to minimize the binary size of the python language projection.

* switched back to inline try/catch blocks instead of using template trycatch_invoker
* created a set of error handling functions (set_invalid_activation_error, set_invalid_arg_count_error, set_invalid_kwd_args_error) to reduce generated code size
* optimized code generation for methods that don't have out parameters
* add noexcept in lots of places

With this change, the compiled DLL size of the full WinRT Python projection (not including XAML or Composition as Python projection doesn't support runtime class inheritance yet) drops from 15.07mb to 12.99mb. 

Code generation differences from master branch can be viewed in the devhawk/pywinrt-output repo: https://github.com/devhawk/pywinrt-output/compare/sizeopt


src/tool/python/helpers.h
```diff
@@ -581,6 +581,22 @@ namespace xlang
         return count;
     }
 
+    int count_out_param(std::vector<method_signature::param_t> const& params)
```
kennykerr: `std::count_if` 😉

devhawk: I changed the implementation of count_in_param/count_out_param to use count_if, but I didn't change the call site. The syntax for calling these algorithms is kinda verbose. That's getting fixed with Ranges, right?

kennykerr: That's right.

PR #286, opened 2019-04-04
Add flags to always produce symbols for Release builds

dunhor:
Per earlier conversation, adding compiler/linker flags to produce symbols for release builds as opposed to trying to get RelWithDebInfo to work with ADO


src/CMakeLists.txt
```diff
@@ -101,19 +101,27 @@ if (MSVC)
         string(REPLACE "/MD" "/MT" CMAKE_CXX_FLAGS_${build_type} "${CMAKE_CXX_FLAGS_${build_type}}")
         string(APPEND CMAKE_CXX_FLAGS_${build_type} " /GL")
 
+        # /GL requires /LTCG
+        string(APPEND CMAKE_EXE_LINKER_FLAGS_${build_type} " /LTCG")
+
         # TODO: replacing /INCREMENTAL leaves ":NO" on the command line, which screws up the link
         #       Figure out the best way to make these changes to the build and linker flags
```
dunhor: This seems pretty trivial to fix, but left it alone

src/CMakeLists.txt
```diff
@@ -101,19 +101,27 @@ if (MSVC)
         string(REPLACE "/MD" "/MT" CMAKE_CXX_FLAGS_${build_type} "${CMAKE_CXX_FLAGS_${build_type}}")
         string(APPEND CMAKE_CXX_FLAGS_${build_type} " /GL")
 
+        # /GL requires /LTCG
+        string(APPEND CMAKE_EXE_LINKER_FLAGS_${build_type} " /LTCG")
```
dunhor: The warnings were annoying me

src/CMakeLists.txt
```diff
@@ -101,19 +101,27 @@ if (MSVC)
...
+
         # TODO: replacing /INCREMENTAL leaves ":NO" on the command line, which screws up the link
         #       Figure out the best way to make these changes to the build and linker flags
         # string(REPLACE "/INCREMENTAL" "" CMAKE_EXE_LINKER_FLAGS_${build_type} "${CMAKE_EXE_LINKER_FLAGS_${build_type}}")
         # string(APPEND CMAKE_EXE_LINKER_FLAGS_${build_type} " /LTCG:INCREMENTAL /OPT:REF")
     endforeach()
 
+    # Always generate symbols for release builds
+    string(APPEND CMAKE_CXX_FLAGS_RELEASE " /Zi")
+    string(APPEND CMAKE_SHARED_LINKER_FLAGS_RELEASE " /DEBUG /OPT:REF /OPT:ICF")
```
devhawk: can we add /MAP too?

dunhor: Do we want that by default? I imagine it's not enormously useful in the general case and I assume increases build times

devhawk: I've been using it alot lately. Doesn't seem to increase build time by a significant amount that I've noticed.

dunhor: So I tried adding `/MAP` to the linker flags and I don't see a `.map` file getting produced. Is there anything additional that I need to do?

dunhor: nvm. Tried it again and it worked this time...

PR #287, opened 2019-04-05
Fix type order stability in C++/WinRT headers

kennykerr:
Type declaration order can become unpredictable, leading to headers changing when they shouldn't. This can cause PCH files to recompile unnecessarily.

### DIFF
--- a/src/tool/cppwinrt/cppwinrt/type_writers.h
+++ b/src/tool/cppwinrt/cppwinrt/type_writers.h
@@ -71,12 +71,20 @@ namespace xlang
     {
         using writer_base<writer>::write;
 
+        struct depends_compare
+        {
+            bool operator()(TypeDef const& left, TypeDef const& right) const
+            {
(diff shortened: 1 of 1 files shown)


src/tool/cppwinrt/cppwinrt/type_writers.h
```diff
@@ -71,12 +71,20 @@ namespace xlang
     {
         using writer_base<writer>::write;
 
+        struct depends_compare
+        {
+            bool operator()(TypeDef const& left, TypeDef const& right) const
```
DefaultRyan: noexcept

kennykerr: The trouble is that the `TypeDef`'s `TypeName` and `TypeNamespace` members are not `noexcept` so any error retrieving those values would not propagate to a console error message but would instead crash the compiler.

PR #296, opened 2019-04-09
Parser test and changes for midl3 compatibility

kythant:
This PR changes the grammar so that it can parse and be compatible with Midl3 files.

Major changes include:
1) Allow enum and property to have its own production for the identifiers because certain keywords for enum and property are allowed such as: String, Object, Int32, etc

2) ApiContract declaration and Attribute declaration

3) Test that runs and parses through 800~ midl3 files (Formatted to not have import statements and velocity keys). Some of these files will come in a later commit. 

4) Some more test for the lexer

For more information on everything that was changed in the grammar. See midl3compatissues.md included in this commit. Or you can also just read the grammar changes in this PR :)


src/tool/xmeta/grammar/midl3compatissues.md
```diff
@@ -0,0 +1,170 @@
...
+
+3) Method declarations can have attributes
+
+
+4) Parameter modifier has REF
+```
+void ReadBytes(ref UInt8[] value);
+```
+
+5) Classes are SEALED by default, need UNSEALED Modifier.
```
timr1126: Isn't this fixed?

kythant: Yes

src/tool/xmeta/grammar/XlangLexer.g4
```diff
@@ -26,20 +26,20 @@ fragment Newline
...
 ATTRIBUTE: 'attribute';
 CLASS: 'runtimeclass';
 CONST: 'const';
-DEFAULT: 'default';
 DELEGATE: 'delegate';
 ENUM: 'enum';
 EVENT: 'event';
 INTERFACE: 'interface';
-INTERNAL: 'internal';
+// INTERNAL: 'internal';
```
timr1126: Remove comment

kythant: Keeping it in as bookkeeping for now

PR #299, opened 2019-04-10
Enable NuGet packaging of Py/WinRT

devhawk:
* Define XLANG_VERSION_STRING macro for pywinrt and cppxlang via target_compile_definitions
* emit XLANG_VERSION_STRING in generated python files
* remove reliance on MIDLRT naming convention to detect property get/set and event add/remove methods 
* add print_usage for command line help text
* move existing C++/WinRT package contents in /src/package to /src/package/cppwinrt
* add nuget package assets for Py/WinRT to /src/package/pywinrt
* change default build version to 0.0.0 (typically, build server explicitly passes in the build version to use)


src/package/pywinrt/Microsoft.Windows.PyWinRT.nuspec
```diff
@@ -0,0 +1,23 @@
+﻿<?xml version="1.0"?>
+<package xmlns="http://schemas.microsoft.com/packaging/2013/05/nuspec.xsd">
+  <metadata minClientVersion="2.5">
+    <id>Microsoft.Windows.PyWinRT</id>
+    <version>1.0.0.0</version>
```
BenJKuhn: 0.9?

BenJKuhn: offline conversation, this is overwritten by cmake file, so this is just a dummy value anyway.

PR #321, opened 2019-04-16
Allow C++/WinRT namespace additions to be injected in the top-level headers

kennykerr:
The ability to inject the "special" additions to each namespace into the top-level .h headers rather than the .2.h headers solves a number of dependency/layering problems. First, mysterious compiler/linker are avoided because popular helpers like "single_threaded_vector" are only defined if their dependent types are actually defined. Second, it means that things like the coroutine helpers can actually be appended to individual namespace headers rather than being lumped into a separate header. While this was easier for the implementation, it has caused a bit of friction during the 2.0 rollout as developers have had to remember to include an additional header. 

The ability to do all of this hinges on the removal of the use of std::is_base_of, which has for a long time caused an overly deep type dependency in C++/WinRT where every use of get_abi had to have the full type defintion of its argument. This meant that more headers needed to be included and things like the input param types could not live in the top-level .h headers. 

Thanks to the xlang metadata reader, C++/WinRT now properly categorizes parameter types and more acurately casts to the ABI, thus avoiding get_abi in all cases and put_abi in some cases. I plan to further clean this up by removing all uses of put_abi, but this is a good first step as we no longer use get_abi anywhere internally.


src/tool/cppwinrt/strings/base_string.h
```diff
@@ -333,6 +333,57 @@ namespace winrt::impl
...
+            if constexpr (std::is_same_v<T, hstring>)
+            {
+                object.clear();
+            }
+            else
+            {
+                object = nullptr;
+            }
+
+            return (void**)(&object);
```
DefaultRyan: Boo C-style cast.

src/tool/cppwinrt/cppwinrt/code_writers.h
```diff
@@ -591,35 +591,42 @@ namespace xlang
...
-                    {
-                        w.write("get_abi(%)", param_name);
-                    }
-                    else
-                    {
-                        w.write(param_name);
-                    }
+                case param_category::object_type:
+                case param_category::string_type:
+                    w.write("*(void**)(&%)", param_name);
```
DefaultRyan: Replace C style cast

src/tool/cppwinrt/strings/base_string.h
kennykerr: It's pretty concise. 😉 But yeah...

src/tool/cppwinrt/cppwinrt/code_writers.h
kennykerr: Going to leave this as-is for now. It's tricky because this is used for in params so this goes from:

`(void**)(&param)` 

To this:

`reinterpret_cast<void**>(const_cast<FULL_TYPE_NAME_HERE*>(&param))` 

This is obviously super verbose and I'd argue makes the code harder to debug through.

PR #326, opened 2019-04-17
pe_writer and allow reader::database to accept a buffer instead of a file path

DefaultRyan:
This change introduces `xlang::meta::writer::pe_writer`, which can write out a compliant PE format executable file (aka winmd).

To improve testability, I've also modified `xlang::meta::reader::database` to accept a `std::vector<uint8_t>` buffer instead of a file path. This way, we can consume and decode the writer's output without needing to touch the file system.

### DIFF
--- a/src/library/impl/meta_reader/database.h
+++ b/src/library/impl/meta_reader/database.h
@@ -25,7 +25,7 @@ namespace xlang::meta::reader
 {
     struct cache;
 
-    struct database : file_view
+    struct database
     {
         database(database&&) = delete;
         database& operator=(database&&) = delete;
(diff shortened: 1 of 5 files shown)


src/library/impl/meta_reader/database.h
```diff
@@ -95,24 +95,141 @@ namespace xlang::meta::reader
             return true;
         }
 
-        explicit database(std::string_view const& path, cache const* cache = nullptr) : file_view{ path }, m_path{ path }, m_cache{ cache }
+        explicit database(std::vector<uint8_t>&& buffer, cache const* cache = nullptr) : m_buffer{ std::move(buffer) }, m_view{ m_buffer.data(), m_buffer.data() + m_buffer.size() }, m_cache{ cache }
```
DefaultRyan: This diff isn't rendering well. All that really happened here is:
- Moved the code in the current constructor into a private `initialize` method
- Added a constructor that takes `std::vector<uint8_t>`

PR #327, opened 2019-04-18
Update ABI tool for new fast ABI design and other changes for integrating into the Windows build

dunhor:
Mostly just updates to match the new fast ABI design and ripping out the old one. Also a few updates to make integration in the Windows build easier/possible


src/tool/abi/abi_writer.cpp
```diff
@@ -112,7 +112,25 @@ void writer::pop_inline_namespace()
     }
 }
 
-void writer::push_contract_guard(contract_attributes& vers)
+bool writer::push_contract_guard(version ver)
```
DefaultRyan: This return value appears to be unused.

src/tool/abi/abi_writer.cpp
```diff
@@ -112,7 +112,25 @@ void writer::pop_inline_namespace()
...
+}
+
+void writer::push_contract_guard(contract_version ver)
+{
+    contract_history history;
+    history.current_contract = ver;
+    push_contract_guard(std::move(history));
+}
+
+void writer::push_contract_guard(contract_history vers)
```
DefaultRyan: Why is this pass-by-value. const ref seems more appropriate.

dunhor: It takes ownership

src/tool/abi/abi_writer.cpp
dunhor: It is. You'll just have to look for its use in ternary operators

dunhor: E.g. `w.push_contract_guard(version_from_attribute(fastAttr)) ? 1 : 0`

src/tool/abi/metadata_cache.cpp
```diff
@@ -276,163 +277,173 @@ void metadata_cache::process_class_dependencies(init_state& state, class_type& t
...
-            if (!initialContract)
+            // Take note if this is the default interface
+            if (is_default(ifaceImpl))
             {
-                xlang::throw_invalid("Interface '", iface->clr_full_name(), "' has no contract attribute, but is "
-                    "trying to be used as a fast ABI interface for class '", type.clr_full_name(), "' which does have "
-                    "a contract attribute");
+                XLANG_ASSERT(!fastInterface);
+                XLANG_ASSERT(!iface->fast_class);
+                fastInterface = const_cast<interface_type*>(iface);
```
Scottj1s: >iface [](start = 60, length = 5)

just declare iface non-const to begin with?

src/tool/abi/types.cpp
```diff
@@ -108,6 +149,12 @@ void enum_type::write_cpp_definition(writer& w) const
     {
         if (auto value = field.Constant())
         {
+            auto isExperimental = ::is_experimental(field);
+            if (isExperimental)
+            {
+                w.write("#if defined(ENABLE_WINRT_EXPERIMENTAL_TYPES)\n");
+            }
+
```
Scottj1s: might make a nice little experimental_guard raii

src/tool/abi/metadata_cache.cpp
dunhor: It's not that easy since `find_dependent_type` returns a `const` ref since some of the items _are_ `const` and should be (e.g. element types, etc.). I may be able to do a bit better here since `interface_type`s are a little easier to look up

dunhor: On second thought, just going to add a comment. The `const_cast` here actually _is_ the behavior we want. We're getting something that _might_ be `const`, but when we reach this point, we _know_ that it's not.

src/tool/abi/types.cpp
dunhor: Not a bad idea, but the write ordering isn't always such that destructor order is desired. Could add scopes to force it, but I'm not sure it's necessarily worth the hassle. And it would only be usable in four locations.

PR #328, opened 2019-04-19
Populate xmeta enum model

timr1126:
Extract enum information from the generated ANTLR AST. This populates all enum values based on implicit and explicit assignment and dependence semantics. More details are in the xmeta semantics design notes.

This includes a symbol_table class that will store the model instances, along with several helper functions in the enum model, and the ANTLR listeners to populate the symbol table. This also includes some basic tests, testing implicit dependence, explicit dependence, and assigning both hex and decimal constant values to enum members.


src/test/xmeta/test_symbol_table.cpp
```diff
@@ -0,0 +1,92 @@
...
+
+    ANTLRInputStream input(test_idl);
+    XlangLexer lexer(&input);
+    CommonTokenStream tokens(&lexer);
+    XlangParser parser(&tokens);
+
+    auto tree = parser.xlang();
+    st.reset("ns_members.idl");
+    ast_to_st_listener listener{ st };
+    tree::ParseTreeWalker::DEFAULT.walk(&listener, tree);
```
kythant: Why don't you use setup_and_run_parser here? You can disable error reporting here but setting the last param to true if you know there are gong to be errors. 

setup_and_run_parser(test_idl, listener, true)

src/test/xmeta/helpers.h
```diff
@@ -0,0 +1,25 @@
+#pragma once
+
+#include "antlr4-runtime.h"
+#include "XlangParser.h"
+#include "XlangLexer.h"
+
+using namespace antlr4;
```
DefaultRyan: Don't drag entire namespaces into global scope in a header.
http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rs-using-directive

src/test/xmeta/test_symbol_table.cpp
```diff
@@ -0,0 +1,92 @@
...
+    REQUIRE(enum_members[0].get_id() == "e_member_1");
+    REQUIRE(enum_members[1].get_id() == "e_member_2");
+    REQUIRE(enum_members[2].get_id() == "e_member_3");
+    REQUIRE(enum_members[3].get_id() == "e_member_4");
+    REQUIRE(enum_members[4].get_id() == "e_member_5");
+    auto const& val1 = enum_members[0].get_value();
+    auto const& val2 = enum_members[1].get_value();
+    auto const& val3 = enum_members[2].get_value();
+    auto const& val4 = enum_members[3].get_value();
+    auto const& val5 = enum_members[4].get_value();
```
kythant: nit: lets combine these lines, make this method shorter

src/tool/xmeta/cpp/ast_to_st_listener.cpp
```diff
@@ -0,0 +1,206 @@
+#include <algorithm>
+#include <stdexcept>
+#include "symbol_table.h"
+#include "models/xmeta_models.h"
+#include "ast_to_st_listener.h"
+
+using namespace xlang::xmeta;
+extern bool semantic_error_exists;
+
+std::map<std::string, enum_semantics> enum_semantics_map
```
kythant: make this static and const?

src/test/xmeta/test_symbol_table.cpp
```diff
@@ -0,0 +1,92 @@
...
+#include "symbol_table.h"
+#include "XlangParser.h"
+#include "XlangLexer.h"
+#include "ast_to_st_listener.h"
+#include "helpers.h"
+
+using namespace antlr4;
+using namespace xlang::xmeta;
+
+xmeta_symbol_table st{ "" };
```
DefaultRyan: Is this really required to be global? Prefer to simply initialize a new object for each test case.

src/tool/xmeta/cpp/ast_to_st_listener.cpp
```diff
@@ -0,0 +1,206 @@
...
+         {
+             // Only handles implicit dependence. Explicit dependence is checked and resolved in resolve_enum_val.
+             new_enum->add_member(enum_member{ enum_member_id, decl_line, str_val });
+             return true;
+         }
+     }
+     else
+     {
+         // No expression implies implicit dependence.
+         if (new_enum->get_members().size() == 0)
```
kythant: use empty()

src/test/xmeta/test_symbol_table.cpp
```diff
@@ -0,0 +1,92 @@
...
+    auto const& val2 = enum_members[1].get_value();
+    auto const& val3 = enum_members[2].get_value();
+    auto const& val4 = enum_members[3].get_value();
+    auto const& val5 = enum_members[4].get_value();
+    REQUIRE((val1.is_resolved() && val1.get_resolved_target<int32_t>() == 0));
+    REQUIRE((val2.is_resolved() && val2.get_resolved_target<int32_t>() == 3));
+    REQUIRE((val3.is_resolved() && val3.get_resolved_target<int32_t>() == 4));
+    REQUIRE((val4.is_resolved() && val4.get_resolved_target<int32_t>() == 33));
+    REQUIRE((val5.is_resolved() && val5.get_resolved_target<int32_t>() == 33));
+}
```
kythant: Add a test case for incorrect semantic usages? ie circular expressions. No need to test for cases where we don't assign a integer since it is not semantic and it is tested in testlexer.cpp

src/tool/xmeta/cpp/ast_to_st_listener.cpp
kythant: use vector::empty()

src/tool/xmeta/cpp/ast_to_st_listener.cpp
```diff
@@ -0,0 +1,206 @@
...
+}
+
+bool ast_to_st_listener::resolve_enum_val(enum_member& e_member, std::shared_ptr<enum_model> const& new_enum, std::set<std::string_view>& dependents)
+{
+    auto const& val = e_member.get_value();
+    if (val.is_resolved())
+    {
+        return true;
+    }
+    else
```
kythant: nit: don't need this else

src/test/xmeta/test_symbol_table.cpp
```diff
@@ -0,0 +1,92 @@
...
+    REQUIRE(it != namespaces.end());
+    auto ns = it->second;
+    REQUIRE(ns->get_namespace_bodies().size() == 2);
+    REQUIRE(ns->get_namespace_bodies()[0]->get_containing_namespace() == ns);
+    REQUIRE(ns->get_namespace_bodies()[1]->get_containing_namespace() == ns);
+}
+
+TEST_CASE("Enum test")
+{
+    std::string test_idl =
```
DefaultRyan: Nit: To avoid all those backslashes you can use a raw string literal:
```c++
std::string test_idl = R"abc(
namespace n {}
namespace n{}
)abc";
```

src/test/xmeta/test_symbol_table.cpp
```diff
@@ -0,0 +1,92 @@
...
+    REQUIRE(enum_members[4].get_id() == "e_member_5");
+    auto const& val1 = enum_members[0].get_value();
+    auto const& val2 = enum_members[1].get_value();
+    auto const& val3 = enum_members[2].get_value();
+    auto const& val4 = enum_members[3].get_value();
+    auto const& val5 = enum_members[4].get_value();
+    REQUIRE((val1.is_resolved() && val1.get_resolved_target<int32_t>() == 0));
+    REQUIRE((val2.is_resolved() && val2.get_resolved_target<int32_t>() == 3));
+    REQUIRE((val3.is_resolved() && val3.get_resolved_target<int32_t>() == 4));
+    REQUIRE((val4.is_resolved() && val4.get_resolved_target<int32_t>() == 33));
```
DefaultRyan: Nit: The test input was in hex. It's slightly easier on the eyes if the test check is also in hex. (`0x21` instead of `33`)

src/tool/xmeta/cpp/ast_to_st_listener.cpp
```diff
@@ -0,0 +1,206 @@
...
+        type = enum_semantics_map[ctx->enum_base()->enum_integral_type()->getText()];
+    }
+
+    std::shared_ptr<enum_model> new_enum = std::make_shared<enum_model>(id->getText(), decl_line, m_st.get_cur_assembly(), type);
+
+    for (auto field : ctx->enum_body()->enum_member_declaration())
+    {
+        if (!extract_enum_member(field, new_enum))
+        {
+            return;
```
kythant: if this fails, we should stop the listener in its tracks and try to output a exception or something to indicate it failed

src/tool/xmeta/cpp/models/enum_model.h
```diff
@@ -110,19 +151,39 @@ namespace xlang::xmeta
             m_type{ t }
```
kythant: this looks weird on github, is there an extra indentation

src/tool/xmeta/cpp/symbol_table.h
```diff
@@ -0,0 +1,71 @@
...
+        void write_error(size_t decl_line, std::string_view const& msg);
+        void write_enum_member_name_error(size_t decl_line, std::string_view const& invalid_name, std::string_view const& enum_name);
+        void write_enum_member_expr_ref_error(size_t decl_line, std::string_view const& invalid_name, std::string_view const& enum_name);
+        void write_enum_circular_dependency(size_t decl_line, std::string_view const& invalid_member_id, std::string_view const& enum_name);
+        void write_enum_const_expr_range_error(size_t decl_line, std::string_view const& invalid_expr, std::string_view const& enum_name);
+        void write_namespace_name_error(size_t decl_line, std::string_view const& invalid_name, std::string_view const& original_name);
+        void write_namespace_member_name_error(size_t decl_line, std::string_view const& invalid_name);
+
+    private:
+        std::map<std::string_view, std::shared_ptr<namespace_model>, std::less<>> m_namespaces;
```
kythant: Wishlist: a method that returns a vector of namesapce_models

src/tool/xmeta/cpp/ast_to_st_listener.h
```diff
@@ -0,0 +1,28 @@
...
+    ast_to_st_listener() = delete;
+    ast_to_st_listener(xmeta_symbol_table & symbol_table);
+
+    void exitEnum_declaration(XlangParser::Enum_declarationContext *ctx) override;
+
+    void enterNamespace_declaration(XlangParser::Namespace_declarationContext *ctx) override;
+    void exitNamespace_declaration(XlangParser::Namespace_declarationContext *ctx) override;
+
+private:
+    xmeta_symbol_table & m_st;
```
DefaultRyan: Possible lifetime concern here with this type holding a raw reference. On its own, it's not evil, but all the usage I'm seeing is creating these two objects in the raw on their own, so things could get a little sketchy, especially with how verbose all the usage is so far.

Is there a wrapper/helper that could improve things twofold by 1) making lifetime guarantees by ensuring that the listener is a local that will go out of scope (and so it won't outlive the symbol table) and 2) hiding details that aren't needed for the end user. In my dream world, I wouldn't have to deal with the lexer, parser, tree, or any of the other types when invoking this.

src/tool/xmeta/cpp/ast_to_st_listener.cpp
DefaultRyan: Even worse, why is this a global object. If we need to perform this mapping, hide the details behind a function.

src/tool/xmeta/cpp/ast_to_st_listener.cpp
```diff
@@ -0,0 +1,206 @@
...
+    { "UInt32", enum_semantics::Uint32 },
+    { "Int64", enum_semantics::Int64 },
+    { "UInt64", enum_semantics::Uint64 }
+};
+
+ast_to_st_listener::ast_to_st_listener(xmeta_symbol_table & symbol_table) :
+    m_st{ symbol_table }
+{ }
+
+bool ast_to_st_listener::extract_enum_member(XlangParser::Enum_member_declarationContext *ast_enum_member, std::shared_ptr<enum_model> const& new_enum)
```
DefaultRyan: The semantics of this return value are unclear - is an error indicated by `true` or `false`? If you really want to accomplish this via a return value instead of throwing, then that's where `bool` enums come in handy to make the return type and values communicate meaning.
http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Ri-typed

src/tool/xmeta/cpp/symbol_table.h
timr1126: We can iterate through the map, vector not required

src/test/xmeta/test_symbol_table.cpp
timr1126: True, I'll fix that up.

src/test/xmeta/test_symbol_table.cpp
kythant: The name of this file should be testsymboltable.cpp. For some reason that is how all the tests cpp files are named in the repo rn.

timr1126: We're working with the C++ core guidlines according to the [xlang design docs](https://github.com/Microsoft/xlang/blob/master/design_notes/XDN02%20-%20Engineering%20Guidelines.md), which outlines the [underscore naming standard](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#Rl-camel). It's a lot more readable this way, and we should rename our other tests.

timr1126: I'm testing value assignment to the models to make sure they're populated correctly, which isn't covered by testlexer.cpp. I'll add more tests to the PR to cover the enum semantics.

src/test/xmeta/testlexer.cpp
```diff
@@ -1,76 +1,55 @@
...
-    std::string test_idl =
-        "namespace test{}";
-        
     xlang_test_listener listener;
-    REQUIRE(setup_and_run_parser(test_idl, listener) == 0);
+    xmeta_idl_reader reader{ "" };
+    REQUIRE(reader.read(test_idl, listener) == 0);
 
-    std::set<std::string> &namespaces = listener.namespaces;
+    auto& namespaces = listener.namespaces;
```
kythant: I don't really see a benefit of setting auto here. I guess its fine but yea.

src/test/xmeta/testlexer.cpp
```diff
@@ -121,47 +104,51 @@ TEST_CASE("Spacing")
...
-        { \
-            [contract(Windows.Foundation.UniversalApiContract, 1)] \
-            [uuid(b7de5527-4c8f-42dd-84da-5ec493abdb9a)] \
-            delegate void WebAccountProviderCommandInvokedHandler(WebAccountProviderCommand command); \
-        }";
+    std::istringstream test_idl{ R"(
+        namespace Windows.UI.ApplicationSettings
+        {
+            [contract(Windows.Foundation.UniversalApiContract, 1)]
+            [uuid(b7de5527-4c8f-42dd-84da-5ec493abdb9a)]
```
kythant: Can we change a few characters on this UUID? I don't feel comfortable exposing a UUID that is in use now that I think about it.

kennykerr: `uuidgen | clip`

src/test/xmeta/testlexer.cpp
timr1126: Just consistency. Shouldn't be in this PR though, we should have a separate commit for cleaning up the tests.

src/test/xmeta/testlexer.cpp
timr1126: @kythant agreed. We can put it in a test-cleanup commit.

src/test/xmeta/testlexer.cpp
kythant: It's no big deal just sneak it in haha

PR #331, opened 2019-04-19
Building Python/WinRT projection in Azure Pipelines

devhawk:
This PR includes azure pipelies yaml files to enable building the full Python projection of WinRT in Azure Pipelines. An example Azure build of this branch is here: Example Build: https://dev.azure.com/msft-xlang/public/_build/results?buildId=1877. Additionally, the full signed proejction is available from test.pypi at https://test.pypi.org/project/winrt/.

Note, this PR ~~has~~ had two open issues:
* ~~For fast iteration, it generates a small percentrage of the overall WinRT projection. Before merging, I will change this to generate the enture WinRT projection except for Windows.UI.Xaml and Windows.UI.Composition (which are not supported by Py/WinRT yet).~~ 
  * steps-build-projection.yml reads the projection type from a queue time build variable. If the value of ProjectionType is "full", it generates the full WinRT projection (except Xaml/Comp). Otherwise it generates the small partial projection. The [public projection build pipeline](https://dev.azure.com/msft-xlang/public/_build?definitionId=19&_a=summary) defaults to partial projection. Internal projection build pipeline defaults to full projeciton. 
* ~~I am currently working on a src\package\pywinrt\projection\pipeline-internal.yml pipeline file that will includes a step that signs the compiled binary.~~
  * /src/package/pywinrt/projection contains two pipeline yml files - pipeline-internal and pipeline-public. The only difference between the two pipelines is the addition of a signing step for pipeline-internal. Both yml files use templates from /src/package/pywinrt/projection/AzurePipelinesTemplates to eliminate duplication.

Other Changes:
* For packaging purposes, the binary extension now lives inside the top level package rather than as a sibling of the python code package
* removed code from generated setup.py file for compiling the extension module.
* wrap functions in generated cpp files in a namespace to enable unity builds (i.e. combining multiple cpp files into a single compilation unit). Not currently using unity builds, but we experimented with using them. We kept the code gen change in order to enable maximum flexibility in build pipeline creation down the road
* Add precompiled header support to generated projection
* The full projection uses a custom setup.py file rather than the one generated by pywinrt


Scottj1s:
any chance you can squash so CodeFlow can open this?  (limit is 64 commits)

PR #332, opened 2019-04-19
Write signature blobs

DefaultRyan:
This adds the ability to write signature blobs using the existing `meta::reader` signature types.


Scottj1s:
> namespace xlang::meta::reader

#pragma once here too? (opportunistic)

---
Refers to: src/library/impl/meta_reader/table.h:2 in aee6a81. [](commit_id = aee6a812e2a074cc1d3827dc83cb8970ec4f8036, deletion_comment = False)

src/library/impl/meta_reader/signature.h
```diff
@@ -1,3 +1,4 @@
+#pragma once
 
 namespace xlang::meta::reader
 {
```
Scottj1s: probably too much of a breaking change to move these into a reader/writer-neutral namespace like xlang::meta::signature?

src/library/impl/meta_writer/signature_writer.h
```diff
@@ -0,0 +1,216 @@
+#pragma once
+
+#include "../meta_reader/signature.h"
+
```
Scottj1s: tiny bit awkward

src/library/impl/meta_reader/signature.h
kennykerr: Seems like the natural thing to do. A bit weird otherwise.

src/library/impl/meta_writer/signature_writer.h
kennykerr: The namespace collapse would fix this.

src/library/impl/meta_reader/signature.h
DefaultRyan: I took a crack at this, and it turned into waaaay too much churn to swallow right now. The signature helper types are used extensively across the project in headers that assume they are in `xlang::meta::reader`.

That might be a good plan for these helpers eventually, but this parser development branch is the wrong time and place for such a sweeping refactor.

PR #334, opened 2019-04-19
commas not needed in tags

Scottj1s:


Scottj1s:
> Before we publish we should remove winrt/coroutine.h from the project templates.

will do, thx

PR #337, opened 2019-04-20
Major cleanup of antlr-related build cruft.

DefaultRyan:
This replaces the hard-coded assumptions about checked-in antlr libs/dlls with cmake targets to download (or optionally specif a checked-in zip) of the antlr4cpp source and built it on demand with the project's cmake settings.

Additionally, this targets antlr's generated output into the binary build dir (instead of into sources), and splits the xmeta tool's generated parser code into a separate library. This greatly simplifies test code consuming the xmeta grammar antlr-generated code.

With this change, we won't need any hand-executed generation scripts, and won't need to depend on pre-build binaries, and the only dependency needed to build everything is the existence of the Java Runtime.


DefaultRyan:
It appears there is still some configuring needed for the build containers, but I think we're in a closer spot now.

Issue #339, opened 2019-04-22
xmeta stricmp only works for windows


timr1126:
Used for type name semantic checks. We can use strcasecmp from strings.h for Posix-based machines. For now it only works for windows.

PR #340, opened 2019-04-23
Add namespace_member_model intermediate class

timr1126:
The class stores a reference to its parent namespace_body_model to make getting fully-qualified names easier. Helpers are included.


timr1126:
I moved some namespace_model implementations to a cpp to avoid circular dependency header includes.

src/tool/xmeta/cpp/models/delegate_model.h
```diff
@@ -4,17 +4,19 @@
...
-#include "base_model.h"
-#include "model_types.h"
+#include "namespace_member_model.h"
 #include "formal_parameter_model.h"
+#include "model_types.h"
 
 namespace xlang::xmeta
 {
-    struct delegate_model : base_model
+    struct namespace_body_model;
```
BenJKuhn: since you have the header above, this forward decl is superfluous, right?

src/tool/xmeta/cpp/models/namespace_model.cpp
```diff
@@ -0,0 +1,93 @@
...
+    }
+
+    void namespace_body_model::add_interface(std::shared_ptr<interface_model> const& im)
+    {
+        m_interfaces[im->get_id()] = im;
+    }
+
+    void namespace_body_model::add_enum(std::shared_ptr<enum_model> const& em)
+    {
+        m_enums.emplace_back(em);
```
BenJKuhn: what are the growth characteristics for the data structure here?

src/tool/xmeta/cpp/models/namespace_model.cpp
```diff
@@ -0,0 +1,93 @@
...
+    {
+        for (auto ns_body : m_namespace_bodies)
+        {
+            if (ns_body->member_id_exists(member_id))
+            {
+                return true;
+            }
+        }
+
+        return m_child_namespaces.find(member_id) != m_child_namespaces.end();
```
BenJKuhn: is it ok to add a class and namespaces using the same name?

BrentRe: WinRT doesn't allow a type to have the same name as a namespace. Additionally, WinRT doesn't allow type or namespaces to differ only by case. I think xlang should have the same rules.

timr1126: A nested namespace is considered to be a namespace member. This PR introduces an intermediate class identifying namespace members, so we can aggregate classes, structs, interfaces, enums, delegates, and nested namespaces into one map.

According to [this design note](https://github.com/Microsoft/xlang/blob/user/kythant/parser/design_notes/XDN03%20-%20xlang%20Type%20System.md#namespaces), "xlang namespaces and type names are case preserving but insensitive". So, you cannot have types differing only by case. I'll need to update "member_id_exists" with a custom comparator. The single map should clean things up.

src/tool/xmeta/cpp/models/delegate_model.h
timr1126: We include the namespace_member_model, which generalizes a namespace member. Namespace bodies are defined in `namespace_model.h`. Namespace members contain reference to their containing body, however we don't need to know how it's defined hence the forward declare.

src/tool/xmeta/cpp/models/namespace_model.cpp
timr1126: Actually a single map complicates metadata emission. If we wanted to emit enums, we'd have to filter through the map. It's cleaner to keep it separate, and have a longer member_id_exists method.

src/tool/xmeta/cpp/models/namespace_member_model.h
```diff
@@ -0,0 +1,43 @@
...
+            base_model{ id, decl_line, assembly_name },
+            m_containing_namespace_body{ containing_ns_body }
+        { }
+
+        auto const& get_containing_namespace_body() const noexcept
+        {
+            return m_containing_namespace_body;
+        }
+
+        auto const& get_fully_qualified_name()
```
kythant: does these have to be auto? can it be std::string? makes it more clear.

src/tool/xmeta/cpp/models/namespace_model.h
```diff
@@ -70,60 +73,28 @@ namespace xlang::xmeta
...
+        std::vector<std::shared_ptr<using_namespace_directive_model>> m_using_namespace_directives;
 
         // Members
         std::map<std::string_view, std::shared_ptr<class_model>, std::less<>> m_classes;
         std::map<std::string_view, std::shared_ptr<struct_model>, std::less<>> m_structs;
         std::map<std::string_view, std::shared_ptr<interface_model>, std::less<>> m_interfaces;
-        std::vector<std::shared_ptr<enum_model>> m_enums;
-        std::vector<delegate_model> m_delegates;
+        std::map<std::string_view, std::shared_ptr<enum_model>, std::less<>> m_enums;
+        std::map<std::string_view, std::shared_ptr<delegate_model>, std::less<>> m_delegates;
```
kythant: goodies

src/tool/xmeta/cpp/models/namespace_member_model.h
```diff
@@ -0,0 +1,43 @@
...
+        {
+            auto const& ns_name = m_containing_namespace_body->get_containing_namespace()->get_full_name();
+            return (ns_name == "")
+                ? get_id()
+                : ns_name + "." + get_id();
+        }
+
+        auto const& get_fully_qualified_member_name(std::shared_ptr<base_model> const& member)
```
DefaultRyan: These methods are returning a local variable (the appended string) by const reference, aka, a dangling reference. These are going to have to return a string by value if you're appending the names on the fly.

I think what we want is a `mutable std::optional<std::string> m_full_name;` where we populate and cache the fully qualified name the first time it's requested. This provides two benefits:

1. We don't repeatedly request the same string.
2. We can continue to return the string by const reference, reducing the propensity for callers to accidentally create expensive copies.

src/tool/xmeta/cpp/models/namespace_member_model.h
```diff
@@ -0,0 +1,43 @@
...
+
+        auto const& get_containing_namespace_body() const noexcept
+        {
+            return m_containing_namespace_body;
+        }
+
+        auto const& get_fully_qualified_name()
+        {
+            auto const& ns_name = m_containing_namespace_body->get_containing_namespace()->get_full_name();
+            return (ns_name == "")
```
DefaultRyan: This is an invalid case to check for. Putting types into the root namespace is an error.

src/tool/xmeta/cpp/models/namespace_model.cpp
```diff
@@ -0,0 +1,130 @@
...
+    void namespace_body_model::add_using_namespace_directive(std::shared_ptr<using_namespace_directive_model> const& und)
+    {
+        m_using_namespace_directives.emplace_back(std::move(und));
+    }
+
+    std::string copy_to_lower(std::string_view sv)
+    {
+        std::string s{ sv };
+        std::locale loc;
+        // ***NOTE*** ::tolower is temporary. This only works on the ASCII charset.
```
DefaultRyan: There's an issue for this: #339 . Reference this in the comment.

src/tool/xmeta/cpp/models/namespace_model.cpp
```diff
@@ -0,0 +1,130 @@
...
+#include "enum_model.h"
+#include "interface_model.h"
+#include "property_model.h"
+#include "struct_model.h"
+
+namespace xlang::xmeta
+{
+    void namespace_body_model::add_class(std::shared_ptr<class_model> const& cm)
+    {
+        m_classes[cm->get_id()] = cm;
```
DefaultRyan: Use `insert` instead. This usage of the array subscript operator has two issues:
1. It will silently overwrite an existing entry. `insert`'s return value allows you to check if the insertion was blocked because of an existing element, which allows us to throw an error.
2. It's less performant for insertions. The array subscript operator constructs a `value_type` key-value pair, with the value half default constructed, and then immediately overwrites the value.

src/tool/xmeta/cpp/models/namespace_model.cpp
```diff
@@ -0,0 +1,130 @@
...
+    }
+
+    void namespace_body_model::add_using_alias_directive(std::shared_ptr<using_alias_directive_model> const& uad)
+    {
+        m_using_alias_directives[uad->get_id()] = uad;
+    }
+
+    void namespace_body_model::add_using_namespace_directive(std::shared_ptr<using_namespace_directive_model> const& und)
+    {
+        m_using_namespace_directives.emplace_back(std::move(und));
```
DefaultRyan: This `std::move` won't do what you think it does. The argument was passed by const ref so no rvalue optimization will happen. In this case, it's simpler to remove the move.

src/tool/xmeta/cpp/models/namespace_model.cpp
```diff
@@ -0,0 +1,130 @@
...
+            return static_cast<char>(::tolower(c));
+        });
+        return s;
+    }
+
+    // Checks if two member IDs are semantically the same.
+    // namespace member IDs are case preserving but insensitive. This means if 'classA' is defined,
+    // you cannot define another member 'cLaSsA'. You also cannot reference 'classA' using 'cLaSsA'
+    // since case is preserved.
+    template<typename T>
```
DefaultRyan: How often will this check happen? It's creating two new strings every time it's called.

src/tool/xmeta/cpp/models/namespace_model.cpp
timr1126: Interesting, makes sense.

src/tool/xmeta/cpp/models/namespace_member_model.h
timr1126: Sounds good

src/tool/xmeta/cpp/models/namespace_model.cpp
timr1126: Yeah, my bad.

src/tool/xmeta/cpp/models/namespace_model.cpp
timr1126: We can use stricmp from string.h, but it's only Windows. Posix-based machines have strcasecmp from strings.h. I'll update the issue, since we're not using ::tolower anymore.

src/test/xmeta/testlexer.cpp
```diff
@@ -33,29 +33,30 @@ TEST_CASE("Namespace Identifier")
...
-    auto& namespaces = listener.namespaces;
-
-    REQUIRE(namespaces.find("test1AÆĦǆＺ") != namespaces.end()); // LU
-    REQUIRE(namespaces.find("test2aăɶｚ") != namespaces.end()); // LL
-    REQUIRE(namespaces.find("test3ǅᾜῼ") != namespaces.end()); // LT
-    REQUIRE(namespaces.find("test4ʰˀﾟ") != namespaces.end()); // LM
-    REQUIRE(namespaces.find("test5ªကညￜ") != namespaces.end()); // LO
-    REQUIRE(namespaces.find("test6ᛮⅫⅯ") != namespaces.end()); // NL
-}
+// TODO: Fix lexer/parser errors generated by this test
```
kythant: Keep this file unchange in the PR. THe fix is in the branch already.

src/tool/xmeta/cpp/models/class_model.h
```diff
@@ -37,60 +37,14 @@ namespace xlang::xmeta
...
-
-        void add_member(std::shared_ptr<event_model> const& member)
-        {
-            m_events.emplace_back(member);
-        }
-
         void add_protected_interface_ref(std::shared_ptr<interface_model> const& protected_interface_ref)
         {
-            m_static_interfaces.emplace_back(protected_interface_ref);
+            m_protected_interfaces.emplace_back(protected_interface_ref);
```
DefaultRyan: 👍

src/tool/xmeta/cpp/models/namespace_member_model.h
```diff
@@ -0,0 +1,48 @@
...
+                return namespace_id.value();
+            }
+            namespace_id = m_containing_namespace_body->get_containing_namespace()->get_full_name();
+            assert(namespace_id != "");
+            return namespace_id.value();
+        }
+
+    private:
+        std::shared_ptr<namespace_body_model> m_containing_namespace_body;
+        mutable std::optional<std::string> namespace_id = std::nullopt;
```
DefaultRyan: Technically, `std::nullopt` isn't needed here. The default constructor for `std::optional` does the same thing.

src/tool/xmeta/cpp/models/namespace_member_model.h
```diff
@@ -0,0 +1,48 @@
...
+
+        auto const& get_fully_qualified_name()
+        {
+            auto const& ns_name = m_containing_namespace_body->get_containing_namespace()->get_full_name();
+            return (ns_name == "")
+                ? get_id()
+                : ns_name + "." + get_id();
+        }
+
+        std::string const& get_fully_qualified_member_name(std::shared_ptr<base_model> const& member) const
```
DefaultRyan: Why are there two methods here? And why does this method never look at the member param?

src/tool/xmeta/cpp/models/namespace_model.cpp
```diff
@@ -0,0 +1,120 @@
...
+#include "enum_model.h"
+#include "interface_model.h"
+#include "property_model.h"
+#include "struct_model.h"
+
+namespace xlang::xmeta
+{
+    void namespace_body_model::add_class(std::shared_ptr<class_model> const& cm)
+    {
+        m_classes.insert(std::pair(std::string_view{ cm->get_id() }, cm));
```
DefaultRyan: Seems like we should check the return value here so we can throw an error if a collision occurred.

src/tool/xmeta/cpp/models/namespace_model.cpp
```diff
@@ -0,0 +1,120 @@
...
+
+        return contains_id(m_child_namespaces, member_id);
+    }
+
+    bool namespace_model::child_namespace_exists(std::string_view const& member_id) const
+    {
+        return m_child_namespaces.find(member_id) != m_child_namespaces.end();
+    }
+
+    std::string namespace_model::get_full_name() const
```
DefaultRyan: It's almost guaranteed that this will be asked for. Seems like we should just store this in a member variable and initialize it in the constructor.

src/tool/xmeta/cpp/xmeta_idl_reader.cpp
```diff
@@ -43,52 +35,72 @@ namespace xlang::xmeta
         return parser.getNumberOfSyntaxErrors();
     }
 
+    struct invalid_ns_id
+    {
+        invalid_ns_id(std::string_view const& name) : new_name{ name } { }
+        bool operator()(std::pair<std::string_view, std::shared_ptr<namespace_model>> const& v) const
+        {
+            auto old_name = v.first;
+#if defined(_WIN32)
```
DefaultRyan: If this method only works correctly on Windows, and we don't have an alternate implementation for other platforms, then only build this binary for Windows (instead of building misbehaving components for other platforms). Update CMakelists.txt accordingly.

DefaultRyan: If this method only works correctly on Windows, and we don't have an alternate implementation for other platforms, then only build this binary for Windows (instead of building misbehaving components for other platforms). Update CMakelists.txt accordingly.

src/tool/xmeta/cpp/models/namespace_member_model.h
timr1126: Yeah, I'll need to re-write those.

PR #341, opened 2019-04-23
Async timeout support for C++/WinRT - part 1

kennykerr:
Today C++/WinRT provides support for blocking wait via `async.get()` and cooperative wait via `co_await async` but there is not ability to use a timeout. This has lead some teams needing timeout support to implement various workarounds that are not very efficient or robust. Part 1 introduces support for a timeout for the blocking scenario. Part 2 introduces support for cooperative timeout for coroutines.

The `get` function follows the pattern established by `std::future` so it makes sense to adopt future's `wait_for` function to support a timeout as well. Note that WinRT only allows waiting once. That means you cannot use `wait_for` in a loop, or any other scenario where multiple waits would occur.

You can wait for an `IAsyncAction` as follows:

```C++

IAsyncAction async = ...

if (async.wait_for(5s) == AsyncStatus::Completed)
{
    puts("done");
}
```

You can wait for an `IAsyncOperation` as follows:

```C++
IAsyncOperation<hstring> async = ...

if (async.wait_for(5s) == AsyncStatus::Completed)
{
    printf("result: %ls\n", async.get().c_str());
}
```

Or more comprehensively:

```C++
IAsyncOperation<hstring> async = ...

switch (async.wait_for(2s))
{
case AsyncStatus::Completed:
    printf("result: %ls\n", async.get().c_str());
    break;
case AsyncStatus::Started:
    puts("still running");
    break;
case AsyncStatus::Canceled:
    puts("canceled");
    break;
case AsyncStatus::Error:
    puts("failed");
    break;
}
```

### DIFF
--- a/src/tool/cppwinrt/cppwinrt/code_writers.h
+++ b/src/tool/cppwinrt/cppwinrt/code_writers.h
@@ -1339,7 +1339,7 @@ namespace xlang
         {
             wait_get(static_cast<Windows::Foundation::IAsyncAction const&>(static_cast<D const&>(*this)));
         }
-        auto wait_for(std::chrono::high_resolution_clock::duration const& timeout) const
+        auto wait_for(Windows::Foundation::TimeSpan const& timeout) const
         {
             return impl::wait_for(static_cast<Windows::Foundation::IAsyncAction const&>(static_cast<D const&>(*this)), timeout);
         }
@@ -1351,9 +1351,9 @@ namespace xlang
         {
             return wait_get(static_cast<Windows::Foundation::IAsyncOperation<TResult> const&>(static_cast<D const&>(*this)));
         }
-        auto wait_for(std::chrono::high_resolution_clock::duration const& timeout) const
+        auto wait_for(Windows::Foundation::TimeSpan const& timeout) const
         {
-            return impl::wait_for(static_cast<Windows::Foundation::IAsyncAction const&>(static_cast<D const&>(*this)), timeout);
+            return impl::wait_for(static_cast<Windows::Foundation::IAsyncOperation<TResult> const&>(static_cast<D const&>(*this)), timeout);
         }
 )");
         }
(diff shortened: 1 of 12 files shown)


src/tool/cppwinrt/cppwinrt/code_writers.h
```diff
@@ -1337,33 +1337,49 @@ namespace xlang
...
             w.write(R"(        void get() const
         {
-            blocking_suspend(static_cast<Windows::Foundation::IAsyncAction const&>(static_cast<D const&>(*this)));
-            GetResults();
-        })");
+            wait_get(static_cast<Windows::Foundation::IAsyncAction const&>(static_cast<D const&>(*this)));
+        }
+        auto wait_for(Windows::Foundation::TimeSpan const& timeout) const
+        {
+            return impl::wait_for(static_cast<Windows::Foundation::IAsyncAction const&>(static_cast<D const&>(*this)), timeout);
```
Scottj1s: >impl:: [](start = 19, length = 6)

needed for disambiguation with current wait_for?  (not used on wait_get)

kennykerr: Right. `wait_for` calls `impl::wait_for` but `get` calls `impl::wait_get` so no confusion. I didn't want to call it `impl::get` and rather had the handful of impl functions start with `wait_xxx` for clarity.

cScalpel:
Thank you for this, kennykerr!!! It is exactly what I was looking for when using SerialDevice! I do have a question regarding my implementation:
After I determine that the AsyncStatus is not completed (no data on serial port when using SerialDevice, DataReader.LoadAsync() and IAsyncOperation.wait_for()) I call the IAsyncOperation.Cancel() function. It seems to work but when I return to make a subsequent DataReader.LoadAsync() call, I get a take_ownership_from_abi exception thrown. How can I perform any additional Async functions on the DataReader object after I cancelled the previous operation?

kennykerr:
@cScalpel Glad you like it! You'll need to check the specific information in the thrown exception to figure out what might be wrong (`take_ownership_from_abi` is not an exception). Whether you can reuse the object that produced the async object following cancellation depends on the implementation of that API. You might need to create a new `DataReader`, but I'd first test by wrapping the failing request in a catch block and inspecting the error message. Something like this:

```C++
try { ... } catch (hresult_error const& e)
{
    printf("%ls\n", e.message().c_str());
}
``` 

In general, we don't have expertise to answer questions about specific APIs. For that, I'd suggest Stack Overflow. Tag the question with `c++-winrt` or `windows-runtime` and we'll do our best to respond if we are able.

cScalpel:
Will do, the exception was indeed "The operation identifier is not valid." I'll update my question on stackoverflow. Thanks!

PR #344, opened 2019-04-23
Fast abi forwarder tear-off support 

Scottj1s:
Initial support for creating fast abi forwarder:
msbuild props/targets to preprocess and assemble arch-specific thunks
fast_abi_forwarder helper to call into thunks
test collateral


src/tool/cppwinrt/strings/base_fast_fwd.h
```diff
@@ -0,0 +1,2123 @@
+
+namespace winrt::impl
+{
+// Thunk definitions in arch-specific assembly sources
+#define FAST_FWD_THUNK(index) extern "C" void __stdcall fast_fwd_thunk##index();
```
kennykerr: Can we call it `WINRT_XXX` for consistency with all the other cppwinrt macros?

src/tool/cppwinrt/test_fast_fwd/FastForwarderTests.cpp
```diff
@@ -0,0 +1,332 @@
...
+using namespace Windows::Foundation;
+using namespace FastForwarderTest;
+
+#define REQUIRE_STR_EQUAL(s1, s2) \
+    REQUIRE(std::wstring(s1).compare(std::wstring(s2)) == 0)
+
+template<typename TForwarder, typename TOwner>
+auto make_fast_abi_forwarder(TOwner const& owner, size_t offset)
+{
+	TForwarder forwarder;
```
kennykerr: Some tabs in this file...

src/tool/cppwinrt/test_fast_fwd/pch.h
```diff
@@ -0,0 +1,13 @@
+#pragma once
+
+#include "catch.hpp"
+#include "winrt/coroutine.h"
+#include "winrt/Windows.Foundation.Collections.h"
+#include <windows.h>
+//#include <winrt.h>
```
kennykerr: Remove

src/tool/cppwinrt/test_fast_fwd/pch.h
```diff
@@ -0,0 +1,13 @@
+#pragma once
+
+#include "catch.hpp"
+#include "winrt/coroutine.h"
```
kennykerr: Remove

src/tool/cppwinrt/test_fast_fwd/pch.h
kennykerr: Probably a bunch of includes here aren't needed.

src/tool/cppwinrt/test_fast_fwd/FastForwarderTest.idl
```diff
@@ -0,0 +1,114 @@
...
+            Double  arg13,
+            Single  arg14,
+            Double  arg15,
+            Single  arg16,
+            Double  arg17,
+            Single  arg18);
+    }
+
+    [contract(FastForwarderTest.InternalContract, 1)]
+    runtimeclass Class
```
kennykerr: This class must actually implement multiple interfaces to ensure the forward offsetting works.

kennykerr: And the class lacks the [fastabi] attribute so cppwinrt won't know to fold the various interfaces.

src/tool/cppwinrt/strings/base_fast_fwd.h
Scottj1s: yes, WINRT_FF_THUNK

src/tool/cppwinrt/test_fast_fwd/FastForwarderTest.idl
Scottj1s: right, this was testing the forwarding logic without true multiple interfaces.  will add that in when we add the compiler hook.

PR #346, opened 2019-04-23
Emitting metadata for enum declarations

kythant:
Emits valid enum metadata and the tests for it using IMetadata interface from windows. 

Includes: 
xlang_model_walker.h that walks through each class declaration 
xlang_model_listener.h interface to listen to the walker

xmeta_emit.h is WIP and has skeletal and commented out code to emit other forms of metadata. 

In this PR, you will be able to run the main method which takes in a path to a idl file and it will produce a portable executable xmd file that you can crack open using ILdsm or ILspy.


src/test/xmeta/CMakeLists.txt
```diff
@@ -5,17 +5,32 @@ project(test_xmeta)
 # using /MD flag for antlr4_runtime (for Visual C++ compilers only)
 set(ANTLR4_WITH_STATIC_CRT ON)
 
+include_directories(
```
DefaultRyan: Remove.

src/test/xmeta/CMakeLists.txt
```diff
@@ -5,17 +5,32 @@ project(test_xmeta)
...
+    ${CMAKE_SOURCE_DIR}/tool/xmeta/
+)
+
 add_executable(test_xmeta "")
 
+target_link_libraries(test_xmeta
```
DefaultRyan: Move these down to the same block where windowsapp and ole32 are.

src/test/xmeta/CMakeLists.txt
```diff
@@ -27,7 +42,13 @@ if (MSVC)
     target_link_libraries(test_xmeta windowsapp ole32)
 endif()
 
-target_sources(test_xmeta PUBLIC main.cpp)
+# TODO Refactor the tool dependencies into a library that we can simply link against (instead of double compiling)
```
DefaultRyan: Remove these dependencies on cpp files in the tool. Instead, put that functionality into a lib that you can link the test against.

src/test/xmeta/CMakeLists.txt
```diff
@@ -5,17 +5,32 @@ project(test_xmeta)
...
+
 add_executable(test_xmeta "")
 
+target_link_libraries(test_xmeta 
+    WindowsApp.lib
+    comsuppw.lib
+)
+
+
 if (EXISTS "${CMAKE_SOURCE_DIR}/test/xmeta/example/midl3test")
```
DefaultRyan: Is this stuff still being used?

src/test/xmeta/CMakeLists.txt
```diff
@@ -5,17 +5,32 @@ project(test_xmeta)
...
+target_sources(test_xmeta PUBLIC
+    pch.cpp
+    testlexer.cpp
+    testmidlcompat.cpp
+    testxmetaemit.cpp
+    xlang_test_listener.cpp
+)
 
 target_include_directories(test_xmeta
     PUBLIC ${XLANG_LIBRARY_PATH} ${XLANG_TEST_INC_PATH} ${CMAKE_CURRENT_SOURCE_DIR}/../output/component/source)
```
DefaultRyan: Is ..output/component/source still being used?

src/test/xmeta/testxmetaemit.cpp
```diff
@@ -0,0 +1,120 @@
...
+    std::shared_ptr<xlang::xmeta::xmeta_emit> emitter = std::make_shared<xlang::xmeta::xmeta_emit>(assembly_name);
+    emitter->initialize();
+
+    xlang::meta::writer::pe_writer writer;
+    writer.add_metadata(emitter->save_to_memory());
+
+    xlang::meta::reader::database db{ writer.save_to_memory() };
+
+    REQUIRE(db.Assembly.size() == 1);
+    REQUIRE(db.Assembly[0].Name() == "testidl");
```
DefaultRyan: You should compare against assembly_name and common_assembly_ref here instead of the hardcoded literals.

src/test/xmeta/testxmetaemit.cpp
```diff
@@ -0,0 +1,120 @@
...
+    REQUIRE(db.AssemblyRef.size() == 1);
+    REQUIRE(db.AssemblyRef[0].Name() == "mscorlib");
+    emitter->uninitialize();
```
DefaultRyan: Needing to manually call initialize/uninitialized is an antipattern. Encapsulate that logic into the constructor and destructor of xmeta_emit.

src/test/xmeta/testxmetaemit.cpp
```diff
@@ -0,0 +1,120 @@
...
+    result.Sealed(true);
+    result.WindowsRuntime(true);
+    return result;
+}
+
+TEST_CASE("Assemblies metadata") 
+{
+    std::string assembly_name = "testidl";
+    std::string common_assembly_ref = "mscorlib";
+    std::shared_ptr<xlang::xmeta::xmeta_emit> emitter = std::make_shared<xlang::xmeta::xmeta_emit>(assembly_name);
```
DefaultRyan: Why is xmeta_emit in a shared_ptr?

src/test/xmeta/testxmetaemit.cpp
```diff
@@ -0,0 +1,120 @@
...
+                Blue
+            }
+        }
+    )" };
+    std::string assembly_name = "testidl";
+    xmeta_idl_reader reader{ "" };
+    REQUIRE(reader.read(test_idl) == 0);
+    xlang::xmeta::xlang_model_walker walker(reader.get_namespaces());
+
+    std::shared_ptr<xlang::xmeta::xmeta_emit> emitter = std::make_shared<xlang::xmeta::xmeta_emit>(assembly_name);
```
DefaultRyan: Why is the emitter in a shared_ptr?

src/test/xmeta/testxmetaemit.cpp
```diff
@@ -0,0 +1,120 @@
...
+    emitter->initialize();
+    walker.register_listener(emitter);
+    walker.walk();
+
+    xlang::meta::writer::pe_writer writer;
+    writer.add_metadata(emitter->save_to_memory());
+    xlang::meta::reader::database db{ writer.save_to_memory() };
+    REQUIRE(db.TypeDef.size() == 2);
+
+    auto const& enum_type = db.TypeDef[1];
```
DefaultRyan: You should also test the properties of enum_type (flags, ensure it doesn't have methods, its base type is System.Enum, etc).

src/tool/xmeta/main.cpp
```diff
@@ -6,25 +6,50 @@
 
 #include "xmeta_idl_reader.h"
 #include "ast_to_st_listener.h"
+#include "xlang_model_walker.h"
+#include "xmeta_emit.h"
 
 #if defined(_WIN32)
```
DefaultRyan: There isn't any alternative on non-Windows platforms for this. Instead, I think it makes more sense to simply restrict the entire xmeta tool executable to only WIN32 in CMakeFiles.txt.

src/tool/xmeta/main.cpp
```diff
@@ -6,25 +6,50 @@
...
+#include <winrt/base.h>
 #endif
 
 #pragma execution_character_set("utf-8")
 
 using namespace antlr4;
+using namespace winrt;
+using namespace xlang::xmeta;
+
+std::string remove_extension(const std::string& filename) {
```
DefaultRyan: Prefer to do these with string_view. It's not typically needed to allocate a whole new string for this sort of thing.

src/tool/xmeta/main.cpp
```diff
@@ -6,25 +6,50 @@
...
-    xlang::xmeta::xmeta_idl_reader reader{ argv[1] };
-    size_t num_syntax_errors = reader.read(stream);
-    if (num_syntax_errors)
-    {
-        return EXIT_FAILURE;
-    }
+    assembly_name = remove_extension(assembly_name);
+    xlang::meta::writer::pe_writer writer;
+    writer.add_metadata(emitter->save_to_memory());
+    writer.save_to_file(std::filesystem::current_path().concat("\\" + assembly_name + ".xmd"));
```
DefaultRyan: The output path and/or file should be a command-line argument. Also, prefer to use `append` instead of `concat`, as it will automatically add a system-appropriate directory separator.

DefaultRyan: Should it really be `current_path`? It seems like it's preferable to pass the output file or directory as a command-line argument.

src/tool/xmeta/main.cpp
```diff
@@ -6,25 +6,50 @@
...
 {
     std::ifstream stream;
-    printf("Opening %s \n", argv[1]);
     stream.open(argv[1]);
+    xmeta_idl_reader reader{ "" };
+    reader.read(stream);
+
+    std::string assembly_name = remove_path(std::string(argv[1]));
+    xlang_model_walker walker(reader.get_namespaces());
+    std::shared_ptr<xmeta_emit> emitter = std::make_shared<xmeta_emit>(assembly_name);
```
DefaultRyan: Why is this a shared_ptr?

src/tool/xmeta/xmeta_emit.h
```diff
@@ -0,0 +1,91 @@
...
+#include <meta_writer.h>
+
+namespace xlang::xmeta
+{
+    inline std::string type_semantics_to_string(model_ref<type_semantics> const& semantic_type);
+
+    class xmeta_emit : public xlang_model_listener
+    {
+    public:
+        xmeta_emit(std::string const& assembly_name)
```
DefaultRyan: Explicit?

src/tool/xmeta/xmeta_emit.h
```diff
@@ -0,0 +1,91 @@
...
+        xmeta_emit(std::string const& assembly_name) 
+            : m_assembly_name(assembly_name)
+        { };
+
+        void initialize();
+        void uninitialize();
+        void save_to_file() const;
+		std::vector<uint8_t> save_to_memory() const;
+
+        void listen_namespace_model(std::shared_ptr<namespace_model> const& model) override;
```
DefaultRyan: Consider `final` rather than `override`.

src/tool/xmeta/xmeta_emit.h
```diff
@@ -0,0 +1,91 @@
...
+        void listen_class_model(std::shared_ptr<class_model> const& model) override;
+        void listen_struct_model(std::shared_ptr<struct_model> const& model) override;
+        void listen_interface_model(std::shared_ptr<interface_model> const& model) override;
+        void listen_enum_model(std::shared_ptr<enum_model> const& model) override;
+        void listen_delegate_model(delegate_model const& model) override;
+
+        mdTypeDef define_type_def(std::string const& name, DWORD const& type_flag, mdToken token_extend, mdToken token_implements[]);
```
DefaultRyan: Should these "define" methods be public? They seem more appropriate as private methods.

src/tool/xmeta/xmeta_emit.h
```diff
@@ -0,0 +1,91 @@
...
+        void define_method(std::shared_ptr<method_model> const& model, mdTypeDef const& token_def);
+        void define_property(std::shared_ptr<property_model> const& model, mdTypeDef const& token_def);
+        void define_event(std::shared_ptr<event_model> const& model, mdTypeDef const& token_def);
+
+        void define_return(std::optional<type_ref> const& retun_type, mdTypeDef const& type_def, mdParamDef *token_return);
+        void define_parameters(formal_parameter_model const& model, mdMethodDef const& token_method, int parameter_index);
```
DefaultRyan: For the param index, prefer uint16_t.

src/tool/xmeta/xmeta_emit.h
```diff
@@ -0,0 +1,91 @@
...
+        void listen_interface_model(std::shared_ptr<interface_model> const& model) override;
+        void listen_enum_model(std::shared_ptr<enum_model> const& model) override;
+        void listen_delegate_model(delegate_model const& model) override;
+
+        mdTypeDef define_type_def(std::string const& name, DWORD const& type_flag, mdToken token_extend, mdToken token_implements[]);
+        void define_method(std::shared_ptr<method_model> const& model, mdTypeDef const& token_def);
+        void define_property(std::shared_ptr<property_model> const& model, mdTypeDef const& token_def);
+        void define_event(std::shared_ptr<event_model> const& model, mdTypeDef const& token_def);
+
+        void define_return(std::optional<type_ref> const& retun_type, mdTypeDef const& type_def, mdParamDef *token_return);
```
DefaultRyan: If you have a return value, return it instead of using an out param.

src/tool/xmeta/xmeta_emit.h
```diff
@@ -0,0 +1,91 @@
...
+
+        meta::reader::Module m_module;
+        mdAssembly token_assembly;
+        mdAssemblyRef token_mscorlib;
+        mdTypeRef token_enum;
+        mdTypeRef token_value_type;
+        mdTypeRef token_delegate;
+
+        // A generic assembly metadata struct.
+        static constexpr ASSEMBLYMETADATA s_genericMetadata =
```
DefaultRyan: Does this need to be a class member, or is it an implementation detail of a particular method?

src/tool/xmeta/xmeta_emit.h
```diff
@@ -0,0 +1,91 @@
...
+            nullptr,
+            // ulProcessor
+            0,
+            // rOS - OS independence
+            nullptr,
+            // ulOS
+            0,
+        };
+
+        // Windows specific methods
```
DefaultRyan: How are these methods Windows specific?

src/test/xmeta/CMakeLists.txt
kythant: Yup. The test idl files are not really public so they are not in the repo right now.

src/test/xmeta/CMakeLists.txt
kythant: Apparently not.

src/test/xmeta/testxmetaemit.cpp
kythant: In this case, this test is using xlang_model_walker which takes in a shared pointer for the listener. 

void xlang_model_walker::register_listener(std::shared_ptr<xlang_model_listener> const& listener)

src/tool/xmeta/xlang_model_walker.h
```diff
@@ -0,0 +1,30 @@
...
+#include "xmeta_models.h"
+#include "xlang_model_listener.h"
+
+namespace xlang::xmeta
+{
+    class xlang_model_walker
+    {
+    public:
+        xlang_model_walker(std::map<std::string_view, std::shared_ptr<namespace_model>, std::less<>> const& namespaces)
+            : m_namespaces(namespaces)
```
timr1126: `m_namespaces` should be reference, we don't want to copy the whole map.

src/tool/xmeta/xlang_model_walker.h
```diff
@@ -0,0 +1,30 @@
...
+        std::map<std::string_view, std::shared_ptr<namespace_model>, std::less<>> m_namespaces;
+
+        void enter_namespace_model(std::shared_ptr<namespace_model> const& model);
+        void enter_class_model(std::shared_ptr<class_model> const& model);
+        void enter_struct_model(std::shared_ptr<struct_model> const& model);
+        void enter_interface_model(std::shared_ptr<interface_model> const& model);
+        void enter_enum_model(std::shared_ptr<enum_model> const& model);
+        void enter_delegate_model(delegate_model const& model);
+    };
+}
```
timr1126: Newline before EOF

src/tool/xmeta/xmeta_emit.h
```diff
@@ -0,0 +1,91 @@
...
+        void define_assembly();
+        void define_common_reference_assembly();
+
+        static constexpr DWORD struct_type_flag = tdPublic | tdSealed | tdClass | tdSequentialLayout | tdWindowsRuntime; // Flags: Public | Sealed | Class |  Sequential
+        static constexpr DWORD runtimeclass_type_flag = tdPublic | tdSealed | tdClass | tdWindowsRuntime;                // Flags: class | public | sealed
+        static constexpr DWORD interface_type_flag = tdPublic | tdInterface | tdAbstract | tdWindowsRuntime;    // Flags: : Interface | Public | Abstract 
+        static constexpr DWORD delegate_type_flag = tdPublic | tdSealed | tdClass | tdWindowsRuntime;           // Flags: Public | Sealed | Class
+    };
+}
```
timr1126: Newline at EOF

src/tool/xmeta/xmeta_emit.h
kythant: How about both?

src/tool/xmeta/xmeta_emit.h
kythant: Yea I think we can make them private. There is no situation here where we have to call these define methods from the walker. Unless the walker is also walking through methods, properties, and events.

src/tool/xmeta/xmeta_emit.h
kythant: Noted

src/tool/xmeta/xmeta_emit.h
kythant: It is used both in define_assembly() and define_common_reference_assembly()

src/tool/xmeta/xmeta_emit.h
kythant: Whoops stray comment. The whole file is windows specific.

src/test/xmeta/testxmetaemit.cpp
```diff
@@ -0,0 +1,120 @@
...
+    TypeAttributes result{};
+    result.Visibility(TypeVisibility::Public);
+    result.Sealed(true);
+    result.WindowsRuntime(true);
+    return result;
+}
+
+TEST_CASE("Assemblies metadata") 
+{
+    std::string assembly_name = "testidl";
```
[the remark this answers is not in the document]
kythant: What is the benefit though? I don't see it.

src/tool/xmeta/CMakeLists.txt
```diff
@@ -37,6 +40,7 @@ target_sources(xmeta_antlr PRIVATE ${antlr_generated_src})
 target_include_directories(xmeta_antlr PUBLIC
     ${antlr_generated_dir}
     ${ANTLR4_INCLUDE_DIRS} # antlr4cpp artifacts
+    ${CMAKE_CURRENT_SOURCE_DIR}/cpp/models
```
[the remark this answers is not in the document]
kythant: Nope it does not need it

src/tool/xmeta/cpp/models/event_model.h
```diff
@@ -17,22 +17,30 @@ namespace xlang::xmeta
...
+                std::shared_ptr<method_model> const& remove_method,
+                std::variant<std::shared_ptr<class_model>, std::shared_ptr<interface_model>> const& parent,
+                type_ref&& t) :
             base_model{ id, decl_line, assembly_name },
             m_semantic{ sem },
-            m_add_method_ref{ add_method_ref },
-            m_remove_method_ref{ remove_method_ref },
+            m_add_method{ add_method },
+            m_remove_method{ remove_method },
+            m_parent{ parent },
```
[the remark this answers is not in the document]
kythant: Reverting changes for the commit that was pulled from that PR

src/tool/xmeta/xmeta_emit.h
DefaultRyan: Saying both is redundant. `final` implies `override`. A function should only have exactly one of `virtual`, `override`, or `final`.

https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#Rh-override

kythant: I see. Noted :)

src/test/xmeta/testxmetaemit.cpp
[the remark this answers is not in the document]
timr1126: Faster runtime. Also more obvious it isn't meant to be changed.

src/tool/xmeta/cpp/models/enum_model.h
```diff
@@ -46,11 +46,26 @@ namespace xlang::xmeta
...
             m_value = value;
         }
 
+        enum_value_semantics get_resolved_value() const
+        {
+            assert(m_value.is_resolved());
+            return std::visit(
+                [&](auto&& value) -> enum_value_semantics
+                {
+                    if constexpr (!std::is_integral_v<std::decay_t<decltype(value)>>)
```
timr1126: @DefaultRyan Won't this always be an integral type? `m_value.get_resolved_target()` is a variant of only integral types.

src/test/xmeta/testxmetaemit.cpp
DefaultRyan: And the point of my question is: why is the model walker using a `shared_ptr`? Is this really a case where we expect to have multiple walkers sharing ownership of a listener?

src/tool/xmeta/main.cpp
kythant: The walker takes in a shared_ptr

src/tool/xmeta/main.cpp
kythant: https://github.com/Microsoft/xlang/issues/350
Made this an issue. Will resolve in a PR later with cmd_reader.h

src/tool/xmeta/CMakeLists.txt
```diff
@@ -49,4 +55,12 @@ if (MSVC)
...
-target_link_libraries(xmeta xmeta_antlr xmeta_model)
+if (WIN32)
+    target_link_libraries(xmeta
+        xmeta_antlr
+        xmeta_model
+        xmeta_windows
+        WindowsApp.lib
+        comsuppw.lib
+    )
+endif()
```
timr1126: newline at EOF

src/tool/xmeta/cpp/xlang_model_listener.h
```diff
@@ -0,0 +1,17 @@
+#pragma once
+#include "xmeta_models.h"
+#include <map>
+
+namespace xlang::xmeta
+{
+    class xlang_model_listener
+    {
+    public:
+        virtual void listen_namespace_model(std::shared_ptr<namespace_model> const& model) = 0;
```
timr1126: Do these need to be pure virtual?

src/tool/xmeta/cpp/xlang_model_listener.h
```diff
@@ -0,0 +1,17 @@
+#pragma once
+#include "xmeta_models.h"
+#include <map>
+
+namespace xlang::xmeta
+{
+    class xlang_model_listener
```
timr1126: Can be struct, to remove `public:`

src/tool/xmeta/cpp/xlang_model_walker.h
```diff
@@ -0,0 +1,30 @@
...
+            : m_namespaces(namespaces), m_listener(listener)
+        { };
+
+        void register_listener(xlang_model_listener & listener);
+        void walk();
+        
+        xlang_model_walker() = delete;
+
+    private:
+        xlang_model_listener & m_listener;
```
timr1126: We can make this a `std::vector<std::shared_ptr<xlang_model_listener>> m_listeners;` to allow multiple listeners to be attached.

src/tool/xmeta/xmeta_emit.h
```diff
@@ -0,0 +1,88 @@
...
+    inline std::string type_semantics_to_string(model_ref<type_semantics> const& semantic_type);
+
+    class xmeta_emit : public xlang_model_listener
+    {
+    public:
+        explicit xmeta_emit(std::string_view const& assembly_name);
+        ~xmeta_emit();
+
+        void save_to_file() const;
+		std::vector<uint8_t> save_to_memory() const;
```
timr1126: Change tabs to spaces.

src/tool/xmeta/cpp/xlang_model_walker.h
kythant: Actually I don't really see the need for it right now so I will just leave it as is.

src/tool/xmeta/cpp/models/enum_model.h
DefaultRyan: That's correct. I think we can get away with:
```c++
static_assert(std::is_integral_v<std::decay_t<decltype(value)>>);
return value;
```

src/tool/xmeta/xmeta_emit.cpp
```diff
@@ -0,0 +1,674 @@
...
+}
+
+
+namespace xlang::xmeta
+{
+    xmeta_emit::xmeta_emit(std::string_view const& assembly_name)
+        : m_assembly_name(std::string(assembly_name))
+    {
+        // Getting the meta data dispenser
+        check_hresult(CoInitialize(NULL));
```
DefaultRyan: nullptr

src/tool/xmeta/xmeta_emit.cpp
```diff
@@ -0,0 +1,674 @@
...
+    };
+
+
+    // This doesn't really output in a PE format
+    void xmeta_emit::save_to_file() const
+    {
+        m_metadata_emitter->Save((s2ws(remove_extension(m_assembly_name)) + L".xmeta").c_str(), 0);
+    }
+
+	std::vector<uint8_t> xmeta_emit::save_to_memory() const
```
DefaultRyan: Indentation.

src/tool/xmeta/xmeta_emit.cpp
```diff
@@ -0,0 +1,674 @@
...
+        }
+    }
+
+    mdParamDef xmeta_emit::define_return(std::optional<type_ref> const& retun_type, mdTypeDef const& type_def)
+    {
+        /* Define return value */
+        mdParamDef token_param_return;
+        std::string return_name = type_semantics_to_string(retun_type->get_semantic());
+        check_hresult(m_metadata_emitter->DefineParam(
+            type_def,
```
DefaultRyan: This can't be right. The first arg to DefineParam needs to be a mdMethodDef. There's a case to be made for primarily using xlang's reader types for the type safety to avoid this sort of issue, and to convert to/from the tokens as close as possible to IMetaDataEmit.

DefaultRyan: Also, was this meant to go into this PR? I thought this was just about the enums, not methods.

kythant: Oh yes this is all going to be changed and not meant for it to be the PR

kythant: But I think the define_return method will stay the same except for the name of it. It' s just the way delegate will use it will change.

src/test/xmeta/test_xmeta_emit.cpp
```diff
@@ -0,0 +1,131 @@
+#include "pch.h"
+#include <iostream>
+
+#include "antlr4-runtime.h"
```
timr1126: No need for ANTLR headers anymore thanks to the IDL reader.

PR #348, opened 2019-04-24
Add tearoff support to C++/WinRT

kennykerr:
This is an advanced feature that most developers won't ever need to deal with. Occasionally there is a need to implement an interface as a COM tearoff and this update adds support for that in C++/WinRT. This also allows C++/WinRT implementations to integrate support for the fast abi forwarding implementation that Scott's been working on. An implementation can simply define the `query_interface_tearoff` function and provide a suitable implementation. Something like this:

```C++
struct Sample : implements<Sample, /* interfaces directly implemented */>
{
    HRESULT query_interface_tearoff(GUID const& id, void** result) const noexcept
    {
        if (is_guid_of</* some tearoff interface */>(id))
        {
            return make_tearoff(this, result);
        }

        *result = nullptr;
        return E_NOINTERFACE;
    }

    // rest of implementation...
};
```


src/tool/cppwinrt/test/tearoff.cpp
```diff
@@ -0,0 +1,277 @@
...
+            winrt::com_ptr<IPersist> persist2 = persist.as<IPersist>();
+            REQUIRE(winrt::get_abi(persist) == winrt::get_abi(persist2));
+        }
+
+        // Releasing a reference to the object does not destroy it since the tearoff keeps it alive.
+        REQUIRE(!InteropType::Destroyed);
+        closable = nullptr;
+        REQUIRE(!InteropType::Destroyed);
+
+        // Releasing the tearoof finally destroys the object.
```
Scottj1s: typo (and below)

src/tool/cppwinrt/test/tearoff.cpp
```diff
@@ -0,0 +1,277 @@
...
+        };
+
+        *result = new (std::nothrow) persist(object);
+        return *result ? S_OK : E_OUTOFMEMORY;
+    }
+
+    // Creates an implementation of IStringable as a tearoff.
+    HRESULT make_stringable(winrt::IInspectable const& object, winrt::hstring const& value, void** result) noexcept
+    {
+        struct stringable final : abi::IStringable
```
Scottj1s: could factor a base_tearoff for m_object mgmt, but if these are the only intended test cases, don't bother

kennykerr: Yes, just examples really.

PR #351, opened 2019-04-24
Improved error message to help diagnose invalid winmd files 

kennykerr:
This is a small targeted improvement to the diagnostics produced by `cppwinrt.exe` to help apply the blame accurately when encountering types in a winmd that use non-WinRT types in an ABI context. For example, given this invalid C++/CX code (that sadly compiles):

```C++
namespace BadCppCxComponent
{
    public ref struct Class sealed
    {
        Class()
        {
        }

        Platform::Type^ GetInvalidType()
        {
            return nullptr;
        }
    };
}
```

The cppwinrt compiler now produces a more helpful diagnostic pinpointing the source of the problem:

```
C:\scratch>cppwinrt.exe -in BadCppCxComponent.winmd
 error: Type 'System.Type' could not be found
 method: GetInvalidType
 type: BadCppCxComponent.__IClassPublicNonVirtuals
 database: C:\scratch\BadCppCxComponent.winmd
```

Given a large winmd with hundreds of types, this can make a world of difference.

### DIFF
--- a/src/tool/cppwinrt/cppwinrt/code_writers.h
+++ b/src/tool/cppwinrt/cppwinrt/code_writers.h
@@ -748,8 +748,18 @@ namespace xlang
 
         for (auto&& method : type.MethodList())
         {
-            method_signature signature{ method };
-            w.write(format, get_abi_name(method), bind<write_abi_params>(signature));
+            try
+            {
+                method_signature signature{ method };
+                w.write(format, get_abi_name(method), bind<write_abi_params>(signature));
+            }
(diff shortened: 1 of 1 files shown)


DefaultRyan:
Nice. The other place where this sort of thing frequently seems to comes up is when trying to resolve base types. Any chance of adding an improved diagnostic there as well?

kennykerr:
Thanks Ryan, send me the repro steps for that scenario I can certainly add something to help diagnose.

PR #353, opened 2019-04-24
Improve Python example

zooba:
Improves some of the code for the Python WinML sample.

I have tested this with the latest build I was sent, so I'm not sure how much newer than what's in this repo it is, but it works :)

### DIFF
--- a/samples/python/winml_tutorial/complete_winml_tutorial.py
+++ b/samples/python/winml_tutorial/complete_winml_tutorial.py
@@ -1,11 +1,16 @@
-import sys
-sys.path.append("./pywinrt_output")
-
-import pyrt
 import asyncio
+import csv
+import os
+import time
+import winrt
+import winrt.windows.ai.machinelearning as winml 
+
+from pathlib import Path
+from winrt.windows.graphics.imaging import BitmapDecoder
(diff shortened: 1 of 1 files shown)


samples/python/winml_tutorial/complete_winml_tutorial.py
```diff
@@ -24,92 +29,68 @@ def sync_wrapper(*args, **kwds):
...
         return ret
 
     return async_wrapper if asyncio.iscoroutinefunction(fun) else sync_wrapper
 
-import pyrt.windows.ai.machinelearning as winml 
-
 @timed_op
 def load_model(model_path):
-    return winml.LearningModel.LoadFromFilePath(model_path)
+    return winml.LearningModel.load_from_file_path(os.fspath(model_path))
```
devhawk: Currently, this tutorial is pinned against an older version of the pywinrt projection that doesn't have pythonic member names. How about you make these changes to src\test\python\complete_winml_tutorial.py and I'll update samples\python\winml_tutorial\complete_winml_tutorial.py as part of the overall effort to update samples\python\winml_tutorial

zooba:
Ah sure, didn't find that one. I'll just make a new PR

PR #356, opened 2019-04-25
Building Python/WinRT projection in Azure Pipelines

devhawk:
Note, this PR is squash merged branch so CodeFlow users can review this PR. 
* Original PR #331 
* Original topic branch: https://github.com/Microsoft/xlang/tree/devhawk/buildtest

This PR includes azure pipelies yaml files to enable building the full Python projection of WinRT in Azure Pipelines. An example Azure build of this branch is here: Example Build: https://dev.azure.com/msft-xlang/public/_build/results?buildId=1877. Additionally, the full signed proejction is available from test.pypi at https://test.pypi.org/project/winrt/.

Notes:
* src\package\pywinrt\projection contains two pipeline yml files - pipeline-internal and pipeline-public. The only difference between the two pipelines is the addition of a signing step for pipeline-internal. Both yml files use templates from src\package\pywinrt\projection\AzurePipelinesTemplates to eliminate duplication.
* src\package\pywinrt\projection\AzurePipelinesTemplates\steps-build-projection.yml reads the projection type from a queue time build variable. If the value of ProjectionType is "full", it generates the full WinRT projection (except for Windows.UI.Xaml and Windows.UI.Composition which are not supported by Py/WinRT yet). Otherwise it generates the small partial projection. The [public projection build pipeline](https://dev.azure.com/msft-xlang/public/_build?definitionId=19&_a=summary) defaults to partial projection. Internal projection build pipeline defaults to full projeciton. 

Other Changes:
* For packaging purposes, the binary extension now lives inside the top level package rather than as a sibling of the python code package
* removed code from generated setup.py file for compiling the extension module.
* wrap functions in generated cpp files in a namespace to enable unity builds (i.e. combining multiple cpp files into a single compilation unit). Not currently using unity builds, but we experimented with using them. We kept the code gen change in order to enable maximum flexibility in build pipeline creation down the road
* Add precompiled header support to generated projection
* The full projection uses a custom setup.py file rather than the one generated by pywinrt


devhawk:
here's the link for the publci build of this branch: https://dev.azure.com/msft-xlang/public/_build/results?buildId=1888

src/package/pywinrt/projection/CMakeLists.txt
```diff
@@ -1,32 +1,73 @@
...
+foreach(file ${sources})
+    message(${file})
+endforeach()
 
 set(CMAKE_CXX_STANDARD 17)
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
-add_compile_options(/await /bigobj /GL /GR- /permissive-)
+
+# TODO: re-enable /GL and /LTCG when we get a bigger build machine
+add_compile_options(/await /bigobj /GR- /permissive-)
```
Scottj1s: this applies to both x86 and x64 builds?  workaround just means slightly longer build times?

src/package/pywinrt/projection/generate.ps1
```diff
@@ -0,0 +1,43 @@
+param ([switch]$clean, [switch]$fullProjection)
+
+$windows_sdk = '10.0.17763.0'
+$repoRootPath = (get-item $PSScriptRoot).parent.Parent.parent.Parent.FullName
```
Scottj1s: >.parent.Parent.parent.Parent [](start = 40, length = 28)

camel parenting, nice

src/package/pywinrt/projection/CMakeLists.txt
devhawk: Yes, this applies to both x86 and x64. I was running out of heap space compiling the full projection for both archtiectures w/ LTCG turned on. I am looking into getting access to a build agent with more memory for internal builds, but I am not going to hold up this PR for that.

src/package/pywinrt/projection/generate.ps1
devhawk: these scripts are only for local use, so I'm not going to worry about this casing thing

Scottj1s: I thought it was intentional, to keep track of the number of them

---
In reply to: [278704485](https://github.com/Microsoft/xlang/pull/356#discussion_r278704485) [](ancestors = 278704485)

devhawk: Uhh, yeah that's it! It was intentional to keep track of the number of them!

src/package/pywinrt/projection/AzurePipelinesTemplates/steps-build-projection.yml
```diff
@@ -0,0 +1,85 @@
...
+    downloadType: single
+    artifactName: 'pywinrt'
+    downloadPath: '$(Build.ArtifactStagingDirectory)'
+
+# I have been completely unable to wire up queue-time variables to template parameters so that they work with conditional insertion
+# so instead, I'm passing the projectionType value into the script and switching between full/partial via a script if statement
+- script: |
+    echo %ProjectionType%
+    IF %ProjectionType%==full (
+        $(Build.ArtifactStagingDirectory)/pywinrt/pywinrt.exe -input 10.0.17763.0 -output src/package/pywinrt/projection/pywinrt -verbose -include Windows. -exclude Windows.UI.Composition -exclude Windows.UI.Xaml
```
Scottj1s: >10.0.17763.0 [](start = 69, length = 12)

consider creating an SdkVersion parameter

devhawk: yeah, I thought about that. Given the issues I had getting the full vs. partial build parameter working, I am not going to tackle SDK version as a param right now

PR #359, opened 2019-04-26
Convert IInspectable to IXlangObject

DefaultRyan:
In addition to the conversion, this also provides the default implementations for GetObjectInfo and Equals.


src/tool/cppxlang/strings/base_array.h
```diff
@@ -454,24 +454,31 @@ namespace xlang
...
 
-    inline Windows::Foundation::TrustLevel get_trust_level(Windows::Foundation::IInspectable const& object)
+    inline hstring get_StringRepresentation(Windows::Foundation::IXlangObject const& object)
     {
-        Windows::Foundation::TrustLevel value{};
-        check_hresult((*(impl::inspectable_abi**)&object)->GetTrustLevel(&value));
-        return value;
+        void* value{};
+        check_hresult((*(impl::xlang_object_abi**)&object)->GetObjectInfo(XlangObjectInfoCategory::StringRepresentation, &value));
+        return { static_cast<xlang_string>(value), take_ownership_from_abi };
```
kennykerr: The `take_ownership_from_abi` constructors should all accept a void* as a first parameter so the static_cast should not be necessary.

src/tool/cppxlang/code_writers.h
```diff
@@ -1219,7 +1219,7 @@ namespace xlang
 
     private:
 
-        IReference<T>(IInspectable const& value) : IReference<T>(value.as<IReference<T>>())
+        IReference<T>(IXlangObject const& value) : IReference<T>(value.as<IReference<T>>())
```
devhawk: We haven't speced out boxing/nullable yet. Should we still have IReference here?

DefaultRyan: It should be removed from the Foundation PR.

PR #360, opened 2019-04-26
Add foundation + collection types to xlang

DefaultRyan:
This brings the collection interfaces, async interfaces, and foundation structs into the Foundation namespace, with the following changes:

- No observable collections
- IMapView.Split is removed
- Only specifying single interface inheritance in the metadata
- No "progress" async interfaces
- No IAsyncInfo
- The Id property/methods have been removed from the remaining async interfaces
- Point, Size, and Rect are removed


src/CMakeLists.txt
```diff
@@ -139,6 +139,6 @@ if (WIN32)
 endif()
 
 add_subdirectory(tool)
-add_subdirectory(test)
 add_subdirectory(platform)
```
devhawk: nit: if we're going to move these around, can we make them alphabetical?

DefaultRyan: Currently they have to be ordered a certain way because of dependencies.

src/tool/cppxlang/strings/base_chrono.h
```diff
@@ -17,32 +17,32 @@ namespace xlang
...
+    template <> struct abi<Foundation::DateTime>
     {
         using type = int64_t;
     };
 
-    template <> struct name<Windows::Foundation::TimeSpan>
+    template <> struct name<Foundation::TimeSpan>
     {
-        static constexpr auto & value{ u8"Windows.Foundation.TimeSpan" };
+        static constexpr auto & value{ u8"Foundation.TimeSpan" };
```
devhawk: Shouldn't this name be "xlang.Foundation.TimeSpan"?

src/tool/cppxlang/type_writers.h
```diff
@@ -209,40 +209,20 @@ namespace xlang
 
             // TODO: get rid of all these renames once parity with cppxlang.exe has been reached...
 
-            if (name == "EventRegistrationToken" && ns == "Windows.Foundation")
+            if (name == "EventRegistrationToken" && ns == "Foundation")
```
devhawk: For language projections that build on top of C++ projection, it's a pain in the ass to undo the customization of these core structs. Is there any way to add a command line option (perhaps in a later PR) to emit a C++ proecjtion w/o them?

src/tool/cppxlang/strings/base_chrono.h
devhawk: I looked thru the .il file and see that you've removed the xlang namespace across the board. I'm OK with this, though I worry a bit about name collisions. I suspect "Foundation" is a lot more common than "xlang.Foundation"

src/tool/cppxlang/helpers.h
```diff
@@ -176,7 +176,7 @@ namespace xlang
         {
             auto const&[type_namespace, type_name] = get_type_namespace_and_name(type.GenericType());
 
-            if (type_namespace == "Windows.Foundation")
+            if (type_namespace == "Foundation")
```
devhawk: haven't we removed progress versions?

devhawk:
FYI, I manually kicked off PR validation builds. Not sure why they didn't kick off automatically

src/tool/cppxlang/strings/base_chrono.h
DefaultRyan: Not yet, it isn't. We're the only ones. :)

But seriously, shouldn't the platform developers have the prerogative to reserve a good namespace for themselves? The C++ standard library has `std`, C++/CX has/had `Platform`.

src/tool/cppxlang/type_writers.h
DefaultRyan: It goes further than that. Now that I've laid down the initial cppx language projection (and your Python winrt work), we've got a much better idea of where the common code is, and where the customization points need to be. Let's make sure we have a task to perform some consolidation work into a core projection helper library. That consolidation work is the perfect time to account for these behavior differences.

src/tool/cppxlang/helpers.h
DefaultRyan: Good catch.

src/tool/cppxlang/strings/base_chrono.h
devhawk: Good point. Works for me.

PR #361, opened 2019-04-26
Implementaton of error handling for xlang

manodasanW:
This change defines the xlang error codes and implements the error origination API and the error info interface exposed in the xlang PAL.  Existing code have been updated to use them.


src/platform/activation_abi.cpp
```diff
@@ -19,17 +19,17 @@ namespace xlang::impl
...
                 }
-                else if (result != xlang_error_class_not_available)
+                else if (result != xlang_result::xlang_no_interface)
                 {
                     throw_result(result);
                 }
             }
         }
-        return originate_error(xlang_error_class_not_available);
+        return xlang_originate_error(xlang_result::xlang_no_interface);
```
devhawk: It seems like class not available should be one of the core errors listed in XDN11. I think you removed it as part of that review, but perhaps we should not have? Is it important for language projection code interacting with activation factories to distinguish between "that class doesn't exist in this binary" and "that class exists but doesn't implement the interface you asked for"?

src/platform/error_abi.cpp
```diff
@@ -0,0 +1,191 @@
...
+        {
+            auto result = --m_count;
+            if (result == 0)
+            {
+                delete this;
+            }
+            return result;
+        }
+
+        void GetError(xlang_result* error) noexcept override
```
devhawk: feels odd to have the type be snake_case and the methods be PascalCase. I get it for GetUnk/AddRef/Release, but should we standardize on snake_case for everything else?

manodasanW: Yea I was thinking about that.  Wasn't sure whether we were going to have a convention for interfaces.  But, if we are fine with special casing IUnknown, I will update these functions.

devhawk: I would use snake_case and add a comment to QI/AddRef/Release indicating why they have different naming convention

src/platform/activation_abi.cpp
manodasanW: I looked up what we did with C++/WinRT for this and we do have a different error (returned by the OS).  I think this error maps up with the xlang_type_load error that I have in our list of error codes.  I will update the spec to call out that error for use in this scenario and and use it here.

src/platform/error_abi.cpp
manodasanW: We probably need to come to an agreement on what to use for interfaces and make sure we are consistent through out as there is other new interfaces where we currently use pascal casing.  Opened #363 to track this.

src/platform/published/pal.h
```diff
@@ -192,6 +184,58 @@ extern "C"
     };
 #endif
 
+#ifdef __cplusplus
+    enum class xlang_result : uint32_t
+    {
+        xlang_success = 0,
```
Scottj1s: >xlang_success [](start = 8, length = 13)

we should be consistent on case convention.  also note that c++ enum classes enforce scope, so the "xlang_" prefix is redundant.  (see xlang_string_encoding for both)

manodasanW: I agree, I will address the removal of the xlang prefix.  I am going to leave the casing for now as that is something we need to come to an agreement on for interfaces and this (the pal functions use snake casing while interfaces use pascal casing).  I opened a issue to track that.

src/platform/published/pal.h
```diff
@@ -301,16 +362,7 @@ constexpr xlang_string_encoding& operator&=(xlang_string_encoding& lhs, xlang_st
...
-inline constexpr xlang_result xlang_error_string_not_null_terminated{ static_cast <int32_t>(0x80000017) };
-inline constexpr xlang_result xlang_error_no_interface{ static_cast<int32_t>(0x80004002) };
-inline constexpr xlang_result xlang_error_pointer{ static_cast<int32_t>(0x80004003) };
-inline constexpr xlang_result xlang_error_sadness{ static_cast<int32_t>(0x80004005) };
-inline constexpr xlang_result xlang_error_out_of_memory{ static_cast<int32_t>(0x8007000e) };
-inline constexpr xlang_result xlang_error_invalid_arg{ static_cast<int32_t>(0x80070057) };
-inline constexpr xlang_result xlang_error_untranslatable_string{ static_cast<int32_t>(0x80070459) };
-inline constexpr xlang_result xlang_error_class_not_available{ static_cast<int32_t>(0x80040111) };
+inline constexpr int32_t xlang_hresult_no_interface{ static_cast<int32_t>(0x80004002) };
 #endif
```
Scottj1s: I understand that we need E_NOINTERFACE for QI, for interop scenarios.  But it's a bit confusing that we also have xlang_no_interface.  What prevents a QI impl from returning that?  Maybe QI should be strongly typed with a com_interop_result enum that contains two values, including this one?

Scottj1s:
>         virtual int32_t XLANG_CALL QueryInterface(xlang_guid const& id, void** object) XLANG_NOEXCEPT = 0;

consider an enum here to make interop values clear (see below)

---
Refers to: src/platform/published/pal.h:144 in e14c780. [](commit_id = e14c780b9531d87ec9c5822bad621dfe7f3ebc17, deletion_comment = False)

src/platform/published/pal.h
manodasanW: Addressed.  Tracking casing as part of separate bug.

src/platform/published/pal.h
manodasanW: The stance I was thinking of taking is the return value of QueryInterface isn't a xlang error (which would be xlang_error_info*), but rather a convention documented for that function (similar to the return values of AddRef/Release).  If the caller of QueryInterface decides that the value they recieved from it is something they consider an error / unexpected, they can then generate an xlang error for it.  In the case, QueryInterface returns the hresult version of no interface and that is determined by the caller to be an error in their case, they would originate an xlang error with the error as xlang_result::no_interface.

manodasanW: I am going to complete this PR, but I do want to continue this conversation to determine whether my stance is the right model and whether it would be clear to a developer implementing QueryInterface or whether it should be changed to something as you mentioned or something else.

src/platform/error_abi.cpp
```diff
@@ -0,0 +1,191 @@
...
+            while (last_propagated_error->m_next_propagated_error != nullptr)
+            {
+                last_propagated_error = static_cast<error_info*>(last_propagated_error->m_next_propagated_error.get());
+            }
+            last_propagated_error->m_next_propagated_error = propagated_error;
+        }
+
+    private:
+        xlang_result m_result{};
+        xlang_string m_message{ nullptr };
```
kennykerr: Are these handle types? If so, where is the destructor that releases them?

src/platform/error_abi.cpp
```diff
@@ -0,0 +1,191 @@
...
+        }
+
+        void GetError(xlang_result* error) noexcept override
+        {
+            *error = m_result;
+        }
+
+        void GetMessage(xlang_string* message) noexcept override
+        {
+            *message = m_message;
```
kennykerr: Shouldn't these be calling the equivalent of WindowsDuplicateString?

src/platform/error_abi.cpp
manodasanW: Good catch, I will have a PR to address this and your other comment.

PR #366, opened 2019-04-26
Timr1126/populate sym tables1

timr1126:
Populate delegates and properties for classes and interfaces.
TODO: Populate methods, events and structs.


src/test/xmeta/test_symbol_table.cpp
```diff
@@ -110,3 +110,176 @@ TEST_CASE("Enum circular dependency")
...
     REQUIRE(reader.get_num_semantic_errors() == 1);
 }
+
+TEST_CASE("Delegate test")
+{
+    std::istringstream test_idl{ R"(
+        namespace N
+        {
+            enum E { }
+enum F{}
```
kythant: looks weird here

kythant:
Taking over this PR.

src/tool/xmeta/grammar/XlangParser.g4
```diff
@@ -386,7 +406,7 @@ interface_method_declaration
     ;
 
 interface_property_declaration
-    : attributes? type property_identifier OPEN_BRACE property_accessors CLOSE_BRACE SEMICOLON?
+    : property_declaration
```
kythant: @BrentRe 
According to the CDL we don't allow interface property declarations to have modifiers. Am I right?

src/test/xmeta/test_symbol_table.cpp
```diff
@@ -110,3 +110,176 @@ TEST_CASE("Enum circular dependency")
...
+                Int32 property_3 { set; };
+                Int32 property_2 { set; };
+                Int32 property_3 { get; };
+                Int32 property_4 { get; set; };
+                Int32 property_5 { set; get; }
+            }
+        }
+    )" };
+
+    std::istringstream incorrect_property_test_idl{ R"(
```
DefaultRyan: This test is pretty big. Split off these "invalid" tests into their own test case, so they don't get lost in the sea of validation for valid properties.

src/test/xmeta/test_symbol_table.cpp
```diff
@@ -110,3 +110,176 @@ TEST_CASE("Enum circular dependency")
...
+    REQUIRE(properties[3]->get_get_method()->get_id() == "get_property_4");
+    REQUIRE(properties[3]->get_set_method()->get_id() == "set_property_4");
+    REQUIRE(properties[3]->is_get_method_declared_first());
+    REQUIRE(properties[4]->get_get_method() != nullptr);
+    REQUIRE(properties[4]->get_set_method() != nullptr);
+    REQUIRE(properties[4]->get_get_method()->get_id() == "get_property_5");
+    REQUIRE(properties[4]->get_set_method()->get_id() == "set_property_5");
+    REQUIRE(!properties[4]->is_get_method_declared_first());
+
+    auto classes = ns_body->get_classes();
```
DefaultRyan: Should check that the synthesized default interface exists, and that the property accessors on the runtime class are overriding those of the interface.

src/test/xmeta/test_symbol_table.cpp
```diff
@@ -110,3 +110,176 @@ TEST_CASE("Enum circular dependency")
...
+    REQUIRE((del0_formal_param_0_type.is_resolved() && std::get<simple_type>(del0_formal_param_0_type.get_resolved_target()) == simple_type::Int32));
+    REQUIRE((del0_formal_param_1_type.is_resolved() && std::get<simple_type>(del0_formal_param_1_type.get_resolved_target()) == simple_type::Double));
+    REQUIRE((!del0_formal_param_2_type.is_resolved() && del0_formal_param_2_type.get_ref_name() == "E"));
+
+    REQUIRE(delegates.find("D2") != delegates.end());
+    REQUIRE(delegates.at("D2")->get_return_type() == std::nullopt);
+    REQUIRE(delegates.at("D2")->get_formal_parameters().empty());
+}
+
+TEST_CASE("Property test")
```
DefaultRyan: Split the invalid case into separate test cases. We also need test cases to check
- When a runtime class property getter and setter are defined in different interfaces.
- When a runtime class declares a property name that collides with an existing property one of its interfaces.
- Static properties, and verifying the relationship with the properties defined on the synthesized static interface.

src/test/xmeta/test_symbol_table.cpp
```diff
@@ -110,3 +110,176 @@ TEST_CASE("Enum circular dependency")
...
+    REQUIRE(interfaces.find("I") != interfaces.end());
+    auto iface = interfaces.at("I");
+    auto properties = iface->get_properties();
+    REQUIRE(properties.size() == 5);
+    REQUIRE(properties[0]->get_id() == "property_1");
+    REQUIRE(properties[1]->get_id() == "property_2");
+    REQUIRE(properties[2]->get_id() == "property_3");
+    REQUIRE(properties[3]->get_id() == "property_4");
+    REQUIRE(properties[4]->get_id() == "property_5");
+    REQUIRE(properties[0]->get_get_method() != nullptr);
```
DefaultRyan: Need to verify the parameters and return types of the getters and setters.

src/test/xmeta/test_symbol_table.cpp
```diff
@@ -110,3 +110,176 @@ TEST_CASE("Enum circular dependency")
...
+    REQUIRE(properties[3]->get_set_method()->get_id() == "set_property_4");
+    REQUIRE(properties[3]->is_get_method_declared_first());
+    REQUIRE(properties[4]->get_get_method() != nullptr);
+    REQUIRE(properties[4]->get_set_method() != nullptr);
+    REQUIRE(properties[4]->get_get_method()->get_id() == "get_property_5");
+    REQUIRE(properties[4]->get_set_method()->get_id() == "set_property_5");
+    REQUIRE(!properties[4]->is_get_method_declared_first());
+
+    reader.reset("");
+    REQUIRE(reader.read(incorrect_property_test_idl) == 0);
```
DefaultRyan: What is this return value for? Seems like an implementation detail that should be abstracted away so we don't have to perform as much surgery down the road.

src/tool/xmeta/cpp/models/property_model.h
```diff
@@ -57,9 +69,38 @@ namespace xlang::xmeta
...
+                std::vector<formal_parameter_model>{},
+                method_semantics{});
+        }
+
     private:
+        type_ref m_type;
+        bool m_is_array;
         property_semantics m_semantic;
-        property_type_semantics m_type;
+        bool m_get_declared_first;
```
DefaultRyan: This member is of dubious value. What we *really* care about is the ordering of the methods.

src/test/xmeta/test_symbol_table.cpp
```diff
@@ -110,3 +110,176 @@ TEST_CASE("Enum circular dependency")
...
+    REQUIRE(properties[4]->get_id() == "property_5");
+    REQUIRE(properties[0]->get_get_method() != nullptr);
+    REQUIRE(properties[0]->get_get_method()->get_id() == "get_property_1");
+    REQUIRE(properties[0]->get_set_method() != nullptr);
+    REQUIRE(properties[0]->get_set_method()->get_id() == "set_property_1");
+    REQUIRE(properties[1]->get_get_method() != nullptr);
+    REQUIRE(properties[1]->get_set_method() != nullptr);
+    REQUIRE(properties[1]->get_get_method()->get_id() == "get_property_2");
+    REQUIRE(properties[1]->get_set_method()->get_id() == "set_property_2");
+    REQUIRE(properties[1]->is_get_method_declared_first());
```
DefaultRyan: I alluded to this elsewhere, but instead of this narrow check of whether a particular getter or setter is declared first, what we really care about is the ordering of the methods, particularly on the interface(s) this class implements. Check that ordering instead of this flag.

src/tool/xmeta/cpp/models/property_model.h
```diff
@@ -57,9 +69,38 @@ namespace xlang::xmeta
...
         }
 
+        auto const& get_semantic() const noexcept
+        {
+            return m_semantic;
+        }
+
+        void create_get_method()
+        {
+            m_get_method = std::make_shared<method_model>(
```
DefaultRyan: The getter and setter are owned by the parent interface/class, not by this property. In other words, they need to be created on the owning interface/class, and the property holds a non-owning reference to the getter and/or setter method.

src/tool/xmeta/cpp/xmeta_idl_reader.h
```diff
@@ -22,7 +22,7 @@ namespace xlang::xmeta
         xmeta_idl_reader(std::string_view const& idl_assembly_name)
         {
             m_assembly_names.emplace_back(idl_assembly_name);
-            m_current_assembly = m_assembly_names.back();
+            m_cur_assembly = m_assembly_names.back();
```
DefaultRyan: Let's have either an assembly model, or use the meta reader's Assembly type. Either way, we need to hold more information than just the name.

src/test/xmeta/test_symbol_table.cpp
kythant: read() is returning the number of syntax error that the parser encountered. This makes sense in a test but you're right, for xmeta.exe, it might be more useful to return errors with more information instead.

PR #375, opened 2019-04-30
Improve Python/WinRT Readme

devhawk:
plus a few fixes to WinML sample readme


Scottj1s:
Steve White recommends stripping the en-us tag from permalinks

PR #396, opened 2019-05-08
C++/WinRT fast ABI forwarding and offloading

kennykerr:
This update now removes the versioned interfaces so that they are offloaded to the fast forwarder by default.


src/tool/cppwinrt/cppwinrt/component_writers.h
```diff
@@ -212,11 +212,16 @@ catch (...) { return winrt::to_hresult(); }
...
     static void write_component_interfaces(writer& w, TypeDef const& type)
     {
+        bool const fastabi = has_fastabi(type);
+
         for (auto&&[interface_name, info] : get_interfaces(w, type))
         {
             if (!info.base && !info.is_default && !is_always_disabled(info.type))
             {
-                w.write(", @", interface_name);
+                if (!fastabi || !info.fastabi)
```
BenJKuhn: can the fastabi check be lifted out of the loop?

PR #397, opened 2019-05-09
Isolate fast forwarder support into standalone header for more general use

Scottj1s:


src/tool/cppwinrt/strings/base_fast_forward.h
```diff
@@ -1,3 +1,13 @@
+#include <cstddef>
+#include <atomic>
```
BenJKuhn: is std::atomic safe for non-throwing code?

kennykerr: Yes, it just compiles down to the various intrinsics and is explicitly noexcept.

BenJKuhn: ideally would be good to add a test case for this from the repo validating usage of just this header with /Eh-

PR #406, opened 2019-05-14
Fast ABI forwarder optimizations

Scottj1s:
This PR resolves some bloat issues related to fast ABI:

For components:
Shrank x64 thunks from 32 to 16 bytes by removing a 6 byte padding (the thunk code itself is only 12 bytes).  

For clients:
There is no easy way to implement /Gy-like behavior (function-level linking, or “COMDAT packaging”) in assembler, which is required for /OPT:REF to discard unneeded thunks.  To address this, the nuget now includes a lib instead.


Scottj1s:
Discussed with VS team the potential for ABI breaking changes between toolset versions of lib.exe and link.exe, and was assured there is no risk.

DefaultRyan:
Might want to look into that failed Windows build. Seems due to a dependency on the 142 toolset?

Scottj1s:
> Might want to look into that failed Windows build. Seems due to a dependency on the 142 toolset?

Yes, thanks!

PR #411, opened 2019-05-16
Kythant/interface population

kythant:
This PR includes model population, (Most) semantic checks, and metadata emit for interfaces. Also includes refactors and infrastructure to import metadata.

What's missing in this PR?
tests for the imported metadata.
Semantic checking for overloading methods in interfaces


src/test/xmeta/test_symbol_table.cpp
```diff
@@ -517,3 +515,377 @@ TEST_CASE("Resolving types across namespaces test")
...
+            }
+            interface ITextBox requires IControl
+            {
+                void SetText(String text);
+            }
+            interface IListBox requires IControl
+            {
+               void SetItem(String items);
+            }
+            interface IComboBox requires ITextBox, IListBox {}
```
DefaultRyan: This case seems odd for two reasons:
1. I didn't think xlang was supporting multiple interface inheritance. And multiple "requires" isn't supported without some sort of attribute specifying which required interface becomes the base for inheritance purposes.
2. This makes IComboBox.Paint() ambiguous. This should require some sort of disambiguation on the method.

src/test/xmeta/CMakeLists.txt
```diff
@@ -45,7 +45,13 @@ if (MSVC)
     target_link_libraries(test_xmeta windowsapp ole32 WindowsApp.lib comsuppw.lib)
 endif()
 
-target_sources(test_xmeta PUBLIC main.cpp)
+get_target_property(xmeta_path foundation_metadata Foundation_xmeta)
+add_custom_command(TARGET test_xmeta POST_BUILD
```
devhawk: Should we add be running these tests as part of our Azure pipelines build?

src/test/xmeta/test_symbol_table.cpp
```diff
@@ -517,3 +515,377 @@ TEST_CASE("Resolving types across namespaces test")
...
+    std::istringstream test_idl{ R"(
+        namespace N
+        {
+            struct S1
+            {
+            };
+
+            interface IControl
+            {
+                E1 Draw(S1 p1, M.S2 p2);
```
DefaultRyan: Nice.

src/test/xmeta/test_symbol_table.cpp
```diff
@@ -517,3 +515,377 @@ TEST_CASE("Resolving types across namespaces test")
...
+}
+
+TEST_CASE("Interface property test")
+{
+    std::istringstream test_idl{ R"(
+        namespace N
+        {
+            interface IControl
+            {
+                Int32 property1 { get; set; };
```
DefaultRyan: I see validation that there are three methods, and two properties. However, the model should also be linking the property to its getter and setter (or vice versa) - I don't see that relationship being checked.

src/test/xmeta/test_symbol_table.cpp
```diff
@@ -517,3 +515,377 @@ TEST_CASE("Resolving types across namespaces test")
...
+    REQUIRE(method1->get_id() == "set_property1");
+    REQUIRE(!method1->get_return_type());
+
+    auto const& method2 = methods[2];
+    REQUIRE(method2->get_id() == "get_property2");
+    auto method2_return_type = method2->get_return_type()->get_semantic();
+    REQUIRE((method2_return_type.is_resolved() && std::get<simple_type>(method2_return_type.get_resolved_target()) == simple_type::Int32));
+}
+
+TEST_CASE("Interface property implicit accessors test")
```
DefaultRyan: Think we should have the following tests:
- property getter and setter declared on separate lines
- Rejecting duplicate properties
- Rejecting multiple getters and setters
- Rejecting accessors other than `get` and `set`.
- Rejecting unresolved types in properties.

src/test/xmeta/test_symbol_table.cpp
DefaultRyan: I'd like to see some test coverage for rejecting things like unresolved types for: bases, return types, parameter types.

src/test/xmeta/test_symbol_table.cpp
```diff
@@ -517,3 +515,377 @@ TEST_CASE("Resolving types across namespaces test")
...
+    }
+    {
+        REQUIRE(properties[1]->get_id() == "property2");
+        auto const& property_type = properties[1]->get_type().get_semantic();
+        REQUIRE(property_type.is_resolved());
+        REQUIRE(std::get<std::shared_ptr<struct_model>>(property_type.get_resolved_target())->get_id() == "S2");
+    }
+}
+
+TEST_CASE("Interface event test")
```
DefaultRyan: Similar to the property tests, seems like we should have tests validating that the event is properly linked to its add/remove methods.

src/test/xmeta/test_symbol_table.cpp
```diff
@@ -517,3 +515,377 @@ TEST_CASE("Resolving types across namespaces test")
...
+    auto const& events = model->get_events();
+    REQUIRE(events[0]->get_id() == "Changed");
+
+    auto const& property_type = events[0]->get_type().get_semantic();
+    REQUIRE(property_type.is_resolved());
+    REQUIRE(std::holds_alternative<std::shared_ptr<delegate_model>>(property_type.get_resolved_target()));
+    REQUIRE(std::get<std::shared_ptr<delegate_model>>(property_type.get_resolved_target())->get_id() == "StringListEvent");
+}
+
+TEST_CASE("Interface circular inheritance test")
```
DefaultRyan: I alluded to this above, but I'd like to see more tests showing that we properly reject invalid IDL.

src/test/xmeta/test_xmeta_emit.cpp
```diff
@@ -210,31 +263,34 @@ void test_delegate_type_properties(TypeDef const& delegate_type)
...
-
-    xlang::meta::writer::pe_writer writer;
-    writer.add_metadata(emitter.save_to_memory());
+    REQUIRE(reader.get_num_semantic_errors() == 0);
+    return reader.save_to_memory();
+}
 
-    return writer.save_to_memory();
+template<typename T>
+auto find_type_by_name(table<T> const& type, std::string name, std::string namespace_name)
```
DefaultRyan: If this is coming up frequently, it might be worth considering `reader::cache` which will store these name-based lookups more efficiently.

src/test/xmeta/test_xmeta_emit.cpp
```diff
@@ -543,29 +589,285 @@ TEST_CASE("Struct class type metadata")
...
+    auto const& property1 = property_list.first[0];
+    REQUIRE(property1.Name() == "property1");
+    REQUIRE(property1.Flags().value == method_attributes_no_flags().value);
+    
+    REQUIRE(db.MethodSemantics.size() == 2);
+    for (size_t i = 0; i < db.MethodSemantics.size(); i++)
+    {
+        auto const& assoc = db.MethodSemantics[i].Association().type();
+        REQUIRE(assoc == HasSemantics::Property);
+        auto const& property_sig = db.MethodSemantics[i].Association().Property().Type();
```
DefaultRyan: `Association().Property()` should actually be equal to the `property1` variable declared above. So you can check the property signature and name outside the loop. All that needs to happen down in this loop is that the entry in the semantics table is equal to the `property1`.

src/test/xmeta/test_xmeta_emit.cpp
```diff
@@ -543,29 +589,285 @@ TEST_CASE("Struct class type metadata")
...
+        REQUIRE(std::get<coded_index<TypeDefOrRef>>(property_sig.Type().Type()).TypeRef() == S1_ref);
+        REQUIRE(db.MethodSemantics[i].Association().Property().Name() == property1.Name());
+    }
+    auto const& get_method = db.MethodSemantics[0].Method();
+    REQUIRE(get_method.Name() == "get_property1");
```
DefaultRyan: Check that the `Semantic` flag is set property for the getter and setter.

src/test/xmeta/test_xmeta_emit.cpp
```diff
@@ -543,29 +589,285 @@ TEST_CASE("Struct class type metadata")
...
+    auto const& add_method = db.MethodSemantics[0].Method();
+    REQUIRE(add_method.Name() == "add_Changed");
+    {
+        auto const& sig = add_method.Signature();
+        REQUIRE(std::get<coded_index<TypeDefOrRef>>(sig.ReturnType().Type().Type()).TypeRef() == event_registration_token_ref);
+        REQUIRE(size(sig.Params()) == 1);
+        auto const& param_sig = sig.Params().first[0];
+        REQUIRE(std::get<coded_index<TypeDefOrRef>>(param_sig.Type().Type()).TypeRef() == string_list_event_ref);
+    }
+    auto const& remove_method = db.MethodSemantics[1].Method();
```
DefaultRyan: Apply the above comments about property metadata tests to events.

src/test/xmeta/CMakeLists.txt
kythant: I don't see why not. You will have to show me how to set it up though.

src/test/xmeta/test_xmeta_emit.cpp
kythant: That would require alot of changes to the code and this is just a test so the performance isn't a worry.

src/test/xmeta/test_symbol_table.cpp
kythant: I see what you're saying. I can disable this test for now and come back to it later once we have attributes.

kythant: And also add a comment

PR #422, opened 2019-05-21
Add rules file containing RootNamespace property editor

wjk:
This property is used by the default item templates to populate the namespace of newly generated IDL files. I have no clue why this isn't already provided by the core C++ project infrastructure, seeing as it is (apparently) not C++/WinRT-specific.


Scottj1s:
This is a really powerful project extensibility mechanism I didn't know about (and which I intend to use, thanks!)  But I'm not sure I understand the problem.  When you create a new C++/WinRT item (from template, such as ViewModel), then the namespace is populated with the project's:
namespace $rootnamespace$

Where are you seeing that not be the case?  By adding the ability to override the project's rootnamespace property at any time, we open the possibility for all kinds of build errors.

kennykerr:
Scott just showed me what this looks like in Visual Studio - very nifty!

Scottj1s:
Discussing with Kenny, and in favor of this change, thanks!

PR #423, opened 2019-05-21
Introduce the C++/WinRT winrt::terminate function to provide richer crash dumps in the event of a WinRT exception

kennykerr:
This update introduces the `winrt::terminate` function that may be used in place of `std::terminate` to fail fast without losing rich error information from a WinRT exception in flight. `fire_and_forget` has also been updated to use it instead of `std::terminate` to address debuggability issues reported by OS developers. The `winrt::terminate` function will call `winrt::to_hresult` to rethrow and then release the error information to the ABI. This has the effect of calling `SetRestrictedErrorInfo` with the WinRT exception information before calling `RoFailFastWithErrorContext` to capture this information before terminating.

### DIFF
--- a/src/tool/cppwinrt/strings/base_error.h
+++ b/src/tool/cppwinrt/strings/base_error.h
@@ -380,10 +380,6 @@ namespace winrt
         {
             return hresult_error(impl::error_fail, to_hstring(e.what())).to_abi();
         }
-        catch (...)
-        {
-            std::terminate();
-        }
     }
 
     [[noreturn]] inline void throw_last_error()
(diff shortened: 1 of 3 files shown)


src/tool/cppwinrt/strings/base_error.h
```diff
@@ -380,10 +380,6 @@ namespace winrt
         {
             return hresult_error(impl::error_fail, to_hstring(e.what())).to_abi();
         }
-        catch (...)
```
kennykerr: This ensures that stray exceptions are caught (by noexcept) without unwinding, making for easier debugging.

PR #424, opened 2019-05-22
Optimize C++/WinRT fast ABI forwarding to minimize code size

kennykerr:
This update ensures that the fast_forward.h header is sized to content, meaning that the resulting forwarding vtable is only as big as the largest fast ABI vtable in the component. Developers can still override WINRT_FAST_ABI_SIZE if needed to handle multiple libs and assertions are in place to ensure that these values are always aligned.


src/tool/cppwinrt/cppwinrt/component_writers.h
```diff
@@ -1180,4 +1185,34 @@ namespace winrt::@::implementation
             type.TypeNamespace(),
             bind<write_component_member_definitions>(type));
     }
+
+    static void write_component_fast_abi_thunk(writer& w)
+    {
+        for (uint32_t slot = 6; slot < 1024; ++slot)
+        {
+            auto format = R"(
+#if WINRT_FAST_ABI_SIZE > %
```
Scottj1s: not needed

src/tool/cppwinrt/strings/base_fast_forward.h
```diff
@@ -2164,3 +135,8 @@ namespace winrt
         return new impl::fast_abi_forwarder(owner, *reinterpret_cast<ff_guid const*>(&guid), offset);
     }
 }
+
+#undef WINRT_IMPL_NOVTABLE
```
Scottj1s: naming convention change is fine, lets use pragma push/pop for macros

Issue #415, opened 2019-05-19
CppWinRTOptimized is not documented


ChrisGuzak:
This .vcxproj property enables cppwinrt.exe `-opt` but it is not mentioned in the docs.

Seems that should go here:
https://github.com/microsoft/xlang/blob/d7e00e126fe7727df59ec31a91db5a2fd11bc140/src/package/cppwinrt/nuget/readme.md#customizing

Scottj1s:
Added the table entry, assigning to Kenny to add a hyperlink to blog post, msdn doc, etc

PR #425, opened 2019-05-22
Added C++/WinRT property page

Scottj1s:
decided to leave the propertysheet.props files in place, as they offer a convenient way to drop in custom references (e.g., cross-platform winmds) and targets.


Scottj1s:
@jlaanstra, this PR provides C++/CX projection support

Scottj1s:
fixes #415

PR #435, opened 2019-05-28
C++/WinRT now handles failing async/event delegates the same way that WRL does

### LINKED ISSUE #409: C++/WinRT should handle disconnected async completion handlers much like it does event handlers

kennykerr:
Fixes #409

### DIFF
--- a/src/tool/cppwinrt/strings/base_events.h
+++ b/src/tool/cppwinrt/strings/base_events.h
@@ -335,6 +335,29 @@ namespace winrt::impl
 #pragma warning(suppress: 6386)
         return { new(raw) event_array<T>(capacity), take_ownership_from_abi };
     }
+
+    template <typename Delegate, typename... Arg>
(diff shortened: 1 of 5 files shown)


src/tool/cppwinrt/strings/base_events.h
```diff
@@ -335,6 +335,29 @@ namespace winrt::impl
...
+    bool invoke(Delegate const& delegate, Arg const&... args) noexcept
+    {
+        try
+        {
+            delegate(args...);
+        }
+        catch (...)
+        {
+            int32_t const code = to_hresult();
+            WINRT_RoTransformError(code, 0, nullptr);
```
devhawk: Why are we adding RoTransformError here?

kennykerr: This adds telemetry so that instead of the error info simply disappearing when the exception is destroyed (and the ref count reaches zero), there is an event indicating that the error was transformed explicitly.

PR #440, opened 2019-05-29
Update to Xmeta Model

kythant:
Update to model so that they can depend on each other.

Also includes methods to validate and resolve types in the models.


src/tool/xmeta/cpp/models/base_model.h
```diff
@@ -26,12 +31,65 @@ namespace xlang::xmeta
...
+
+    enum class method_association
+    {
+        None,
+        Property,
+        Event,
+        Constructor
+    };
+
+    enum class simple_type
```
devhawk: These are refered to as "fundamental" types in the type system spec: https://github.com/microsoft/xlang/blob/master/design_notes/XDN03%20-%20xlang%20Type%20System.md#fundamental-types

src/tool/xmeta/cpp/models/base_model.h
```diff
@@ -26,12 +31,65 @@ namespace xlang::xmeta
...
+        UInt32,
+        UInt64,
+        Char16,
+        Single,
+        Double,
+    };
+
+    struct object_type {};
+
+    using class_type_semantics = std::variant<std::monostate,
```
devhawk: The meta_reader project refers to these ase type categories. And it seems odd to call them "class" type semantics, when only one of the possible variant options is class

src/tool/xmeta/cpp/models/base_model.h
```diff
@@ -26,12 +31,65 @@ namespace xlang::xmeta
...
+    using type_semantics = std::variant<
         std::shared_ptr<class_model>,
         std::shared_ptr<enum_model>,
         std::shared_ptr<interface_model>,
         std::shared_ptr<struct_model>,
-        std::shared_ptr<delegate_model>>;
+        std::shared_ptr<delegate_model>,
+        std::shared_ptr<xlang::meta::reader::TypeDef>,
+        simple_type,
+        object_type>;
```
devhawk: I can't leave a comment on the "get_it" function below, but that's an awful function name. "get_by_id" would be better.

devhawk: Also, what does "id" refer to in this context. Is that a generalized concept for "name"?

src/tool/xmeta/cpp/models/struct_model.h
```diff
@@ -32,72 +38,11 @@ namespace xlang::xmeta
...
-                        error_manager.write_unresolved_type_error(get_decl_line(), symbol);
-                    }
-                    else
-                    {
-                        field_type.set_semantic(iter->second);
-                    }
-                }
-            }
-        }
+        bool has_circular_struct_declarations(xlang_error_manager & error_manager);
```
devhawk: this feels like a semantic check that shouldn't be baked directly into the model

src/tool/xmeta/cpp/models/base_model.h
kythant: I have to agree. Will change the function name get_by_id. Id refers to the name of the member without the namespace part.

src/tool/xmeta/cpp/models/struct_model.h
kythant: I've baked alot of the semantic checking into the models. Especially to resolve type refs within it.

src/tool/xmeta/cpp/models/class_model.h
```diff
@@ -5,32 +5,41 @@
...
-#include "model_ref.h"
 #include "class_or_interface_model.h"
-#include "property_model.h"
+#include "model_types.h"
 
 namespace xlang::xmeta
 {
-    // Note: All classes are implicitly public, and explicitly specifying this is not allowed.
-    enum class class_semantics
+    struct class_semantics
```
devhawk: You might be overusing the term semantics. It's not clear (from the name) the difference between type_semantics, class_semantics and class_type_semantics

src/tool/xmeta/cpp/models/class_model.h
```diff
@@ -42,22 +51,39 @@ namespace xlang::xmeta
...
-        {
-            m_static_interfaces.emplace_back(static_interface_ref);
-        }
+        void add_static_interface_ref(std::shared_ptr<interface_model> const& static_interface_ref);
+
+        void add_factory_interface_ref(std::shared_ptr<interface_model> const& static_interface_ref);
+
+        void add_instance_interface_ref(std::shared_ptr<interface_model> const& static_interface_ref);
+
+        //void add_protected_interface_ref(std::shared_ptr<interface_model> const& protected_interface_ref);
```
devhawk: Add a TODO that composition/inheritance is coming in a later update

src/tool/xmeta/cpp/models/struct_model.h
kythant: I think by making the model semantically check itself. I can easily use my walker to go through the models and call the semantic checking methods. The only downside is that the  models become bigger with these extra methods.

src/tool/xmeta/cpp/models/formal_parameter_model.h
```diff
@@ -25,6 +25,12 @@ namespace xlang::xmeta
             m_type{ std::move(type) }
         { }
 
+        formal_parameter_model(std::string_view const& id, parameter_semantics sem, type_ref&& type) :
```
devhawk: nit picky, but are there other types of parameter models OTHER than formal? Shouldn't this just be parameter_model?

src/tool/xmeta/cpp/models/struct_model.h
devhawk: type ref resolution should be baked in, but I worry that baking in other types of validation will make it harder to add semantic checks down the road.

src/tool/xmeta/cpp/models/base_model.h
devhawk: maybe then "get_by_name", standardizing "name" to be "type name w/o namespace" and "qualified_name" to be "type name w/ namespace".  "fully_namespace_qualified_name" would be most descriptive, but that's too verbose I think .

src/tool/xmeta/cpp/models/class_model.h
kythant: I've change this to class_modifier

src/tool/xmeta/cpp/models/formal_parameter_model.h
kythant: I will change this in a later PR

src/tool/xmeta/cpp/models/struct_model.h
kythant: I will have that in mind when I write validation stuff.

src/tool/xmeta/cpp/models/property_model.h
```diff
@@ -57,9 +65,21 @@ namespace xlang::xmeta
...
+
+        void resolve(symbol_table & symbols, xlang_error_manager & error_manager, std::string const& qualified_name);
+
     private:
-        property_semantics m_semantic;
-        property_type_semantics m_type;
+        property_modifier m_modifier;
+        model_ref<std::shared_ptr<property_model>> m_implemented_property_ref;
+        type_ref m_type;
+        bool m_is_array;
```
devhawk: I wonder if we should bake array semantics into a "parameter type" model, instead of having to code array semantics into property directly

src/tool/xmeta/cpp/models/method_model.h
```diff
@@ -5,30 +5,57 @@
...
-            m_semantic{ sem },
+            m_modifier{ sem },
+            m_association{ assoc },
+            m_implemented_method_ref{ "" }
+        { }
+
+        method_model(std::string_view const& id,
+                size_t decl_line,
+                std::string_view const& assembly_name,
+                std::optional<type_ref>&& return_type,
```
devhawk: you're missing a way of encoding that the return type or any of the parameter types are arrays

src/tool/xmeta/cpp/models/formal_parameter_model.h
```diff
@@ -25,6 +26,12 @@ namespace xlang::xmeta
...
+            base_model{ id, 0, "test_only" },
+            m_semantic{ sem },
+            m_type{ std::move(type) }
```
devhawk: you're missing a mechanism to capture the parameter is an array

src/tool/xmeta/cpp/models/property_model.h
kythant: I think we should encode that into the type_ref in model_types. It is the type_refs that can be an array anyway. This should cover this, return types and the parameter types.

src/tool/xmeta/cpp/models/base_model.h
```diff
@@ -58,27 +116,27 @@ namespace xlang::xmeta
...
 
     template<class T>
-    inline auto get_it(std::map<std::string_view, std::shared_ptr<T>, std::less<>> const& m, std::string_view const& id)
+    inline auto get_by_name(std::map<std::string_view, std::shared_ptr<T>, std::less<>> const& m, std::string_view const& id)
     {
         return m.find(id);
     }
 
     template<class T>
     inline bool contains_id(T const& v, std::string_view const& id)
```
devhawk: still using id here

src/tool/xmeta/cpp/models/class_or_interface_model.h
```diff
@@ -34,46 +43,36 @@ namespace xlang::xmeta
...
-            m_interface_base_refs.emplace_back(std::string(interface_base_ref));
-        }
+        void add_interface_base_ref(std::string_view const& interface_base_ref);
 
-        void add_interface_base_ref(size_t index, std::shared_ptr<interface_model> interface_base_ref)
-        {
-            assert(index < m_interface_base_refs.size());
-            m_interface_base_refs[index].resolve(interface_base_ref);
-        }
+        compilation_error add_member(std::shared_ptr<method_model> const& member);
```
devhawk: are add_member overloads called during parsing? If so, why would it return a compilation_error?

src/tool/xmeta/cpp/models/class_or_interface_model.h
```diff
@@ -34,46 +43,36 @@ namespace xlang::xmeta
...
-        {
-            m_methods.emplace_back(member);
-        }
 
-        void add_member(std::shared_ptr<event_model> const& member)
-        {
-            m_events.emplace_back(member);
-        }
+        bool member_id_exists(std::string_view const& id);
+        bool event_or_property_id_exists(std::string_view const& id);
```
devhawk: standardize on "name" instead of "id" across the tool

devhawk: Note, you can have this method signature be "bool event_or_property_exists(std::string_view const& name)"

src/tool/xmeta/cpp/models/compilation_unit.cpp
```diff
@@ -0,0 +1,62 @@
+#include "compilation_unit.h"
+
+namespace xlang::xmeta
+{
+    compilation_error symbol_table::set_symbol(std::string_view symbol, type_category const& class_type)
```
devhawk: Compilation_error isn't clear here - are we talking errors that are raised during parsing or during a later validation step?

src/tool/xmeta/cpp/models/class_or_interface_model.h
kythant: I used compilation_error to return the message that add_member can't add the member since the member with the same name already exist in the model. They are not called during parsing. They are called when we are walking the AST and putting them into the models.

src/tool/xmeta/cpp/models/compilation_unit.cpp
kythant: They are suppose to be semantic errors. I will change the name to reflect that.

src/tool/xmeta/cpp/models/enum_model.h
```diff
@@ -8,12 +8,12 @@
...
 #include <variant>
 
-#include "model_ref.h"
+#include "base_model.h"
 #include "namespace_member_model.h"
 
 namespace xlang::xmeta
 {
-    enum class enum_semantics
+    enum class enum_types
```
devhawk: nit, should be enum_type (not plural).

kythant: whoops why did I do that

PR #444, opened 2019-05-29
Update azure-pipelines.yml

devhawk:
update windows azdo pipeline file to use hosted windows-2019 image instead of container

### DIFF
--- a/src/scripts/windows/azure-pipelines.yml
+++ b/src/scripts/windows/azure-pipelines.yml
@@ -32,12 +32,12 @@ steps:
   displayName: 'Generate xlang build files via cmake'
   inputs:
     cwd: '.'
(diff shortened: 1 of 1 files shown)


Scottj1s:
found the cause of the build failure.  armasm.exe and armasm64.exe are on the VM, and the fast forwarder lib is being built for all arches, but all into the same folder, since it doesn't contain a platform element.  I think if we just restore this, it should work:
-C _build\Windows\x86\Release 

also, are the cmake generation and build steps supposed to have the same name, CMake@1 ?

devhawk:
Yes, I'm using the CMake@1 task to both generate the ninja build file and run ninja. cmake's [--build command](https://cmake.org/cmake/help/v3.14/manual/cmake.1.html#build-a-project) is used "to build an already-generated project binary tree"

devhawk:
parsing the build folder path is fragile. Filed #447

Scottj1s:
are you going to remove the dockerfile now?

devhawk:
As per a conversation with @scottj1s IRL, I did NOT delete the windows dockerfile. But since this PR moves xlang over to the hosted VS2019 image, we certainly can.

Note, I paused but not delete the [create-xlang-windows-build-image](https://dev.azure.com/msft-xlang/public/_build?definitionId=9) job

PR #445, opened 2019-05-30
Runtime class models

kythant:
Setting up models for runtime class.

Synthesized static, factory and instance interfaces for the class. 

Also includes a bunch of varaible name changes.


src/library/impl/meta_reader/schema.h
```diff
@@ -61,6 +61,11 @@ namespace xlang::meta::reader
             return get_string(2);
         }
 
+        auto FullyQualifiedName() const
```
devhawk: It seems like a mistake to add this to meta reader. First, you don't have a delimiter between NS and name. Second, different languages use different symbols to delimit - dot for C# vs double colon for C++ for example.

src/tool/xmeta/cpp/models/class_model.h
```diff
@@ -51,13 +51,40 @@ namespace xlang::xmeta
             return m_modifier;
         }
 
-        void add_class_base_ref(std::string_view const& class_base_ref);
+        auto const& get_synthesized_static_interface() const noexcept
```
devhawk: FYI, versioning will likely require changes here. In WinRT, types can introduce new factory methods, static members and instance members with each release. In IDL parsing terms, this means the possibility of multiple synthesized factory/static/instance interfaces for each type. For example, one interface with all the static members introduced in v1, another interface with all the static members introduced in v2. Versioning story for xlang isn't worked out yet, but the deeper you bake the "one synthesized factory/static/instance interface per runtime class" into the tool, the harder it will be to change

kythant: Ahh noted. I was wondering why they were vectors. This makes sense now.

PR #446, opened 2019-05-30
Update azure-pipelines.yml

Scottj1s:
Use ubuntu 18.04 container image


Scottj1s:
just an experiment to see what's missing from base 18.04 image... lots

devhawk:
Thus would probably work if you dropped in the apt-get commands currently in src\scripts\ubuntu\Dockerfile into the yaml file

PR #449, opened 2019-05-30
C++/WinRT single_threaded_observable_vector now supports IInspectable automatically

kennykerr:
The `single_threaded_observable_vector<T>()` helper function creates an implementation of `IObservableVector<T>` and if `T` is not `IInspectable` then the resulting object will now also implement `IObservableVector<IInspectable>`. Observable collections are predominantly used for Xaml binding and the Xaml framework requires `IObservableVector<IInspectable>` in order to support binding. 

And every time I work on collections I always end up with way more test code than shipping code...

I have also confirmed that this works beautifully with Xaml binding.


kennykerr:
Thanks! I don't think there's a concern since Xaml already does copious amounts of boxing/unboxing and the underlying boxing mechanics continues to rely on the implementation provided by `Windows::Foundation::PropertyValue`.

PR #452, opened 2019-05-31
Use get_category consistently in cppwinrt implementation

kennykerr:
Some time back I introduced the `get_category` helper to solve some more challenging parameter categorization issues in the cppwinrt compiler. This made the clunky `can_take_ownership_of_return_type` and `wrap_abi` helper functions redundant. This update just removes those functions and makes use of `get_category` throughout. Internal bug #21214501.

### DIFF
--- a/src/tool/cppwinrt/cppwinrt/code_writers.h
+++ b/src/tool/cppwinrt/cppwinrt/code_writers.h
@@ -675,23 +675,20 @@ namespace xlang
         if (method_signature.return_signature())
         {
             s();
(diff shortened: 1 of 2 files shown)


devhawk:
There are a bunch of useful neta reader helpers spread across cppwinrt, pywinrt and other projects. get_category seems like another such useful helper. Can we start a "meta_reader contrib" file in library to host these and stop duplicating effort across tools?

kennykerr:
Yes, we should definitely keep an eye out for redundant helpers and fold them into the xlang library if they are widely applicable.

PR #453, opened 2019-05-31
Runtime class metadata

kythant:
Runtime class models, metadata and the tests for it. 

Synthesizing I<classname>, I<classname>Factory, I<classname>Statics interfaces for the runtime class. The runtime class will have a method_def for instance and factory interfaces. 

Also includes:
Fix to correctly parse array types.
Fix to meta_reader code so that methodimpl list works correctly.

What was scoped out:
Certain error reporting for semantically wrong idl files
Imported interface types
Protected, override, and composable.


src/test/xmeta/test_xmeta_emit.cpp
```diff
@@ -1115,24 +1292,310 @@ TEST_CASE("Interface type metadata")
...
+
+    ExpectedClass c1{ "C1", {}, {}, {}, {
+        { ExpectedMethod{ "Draw", ElementType::I4, { ExpectedParam{ "param1", ElementType::I4 } }, class_method_attributes() }, "IC1.Draw" },
+    }, "N.C0"};
+
+    ExpectedInterface ic1{ "IC1",
+        { ExpectedMethod{ "Draw", ElementType::I4, { ExpectedParam{ "param1", ElementType::I4 } }, interface_method_attributes() } }
+};
+
+    std::vector<std::shared_ptr<ExpectedType>> fileTypes =
```
Scottj1s: curious about the pattern - who are you sharing with?  why not unique_ptr?

kythant: I was just thinking to turn all of them into non pointer types or just unique_ptrs like you said since it is just mostly going to be contained in a test. But I was racing against the clock.

src/tool/xmeta/cpp/ast_to_st_listener.cpp
```diff
@@ -374,61 +404,431 @@ listener_error ast_to_st_listener::extract_property_accessors(XlangParser::Inter
...
+            }
+            if (class_member->class_method_declaration())
+            {
+                auto class_method = class_member->class_method_declaration();
+                std::string method_id = class_method->IDENTIFIER()->getText();
+
+                std::optional<type_ref> tr = type_ref{ class_method->return_type()->getText() };
+                extract_type(class_method->return_type(), tr);
+
+                // TODO: semantic checks
```
Scottj1s: repeated pattern - factor out a semantic_check() ?

src/tool/xmeta/xmeta_emit.cpp
```diff
@@ -756,21 +840,36 @@ namespace xlang::xmeta
...
+                    {
+                        return TypeSig{ ElementType::Class, type_ref.coded_index<TypeDefOrRef>() };
+                    }
+                    else
+                    {
+                        return TypeSig{ ElementType::ValueType, type_ref.coded_index<TypeDefOrRef>() };
+                    }
                 }
-                else
+                // else it is either one of the tyep categories
```
Scottj1s: typo

src/test/xmeta/test_xmeta_emit.cpp
kythant: Okay I tried, ctrl+f replace std::shared_ptr<ExpectedType> to unique_ptr. Getting C2280 error. I am guessing somewhere deep in this code it is being shared somehow?

Scottj1s:
wow, 266 commits! ;-)  codeflow has a limit of 64, which is a good upper bound

kythant:
oh whoops hahaha. I've been branching out of the branches I created back from two months ago.

src/tool/xmeta/cpp/ast_to_st_listener.cpp
kythant: This may require a deeper rework of the grammar which i have something in mind that will make this easier (Combining all method, properties, and events into one grammar production and just semantically check what is allowed later). Will do so this in a refactor later.

PR #460, opened 2019-06-03
Refactoring xmeta tests

EricJohnson327:
Refactorying xmeta tests to prepare for future xmeta additions.

### DIFF
--- a/src/test/xmeta/test_xmeta_emit.cpp
+++ b/src/test/xmeta/test_xmeta_emit.cpp
@@ -338,6 +338,43 @@ std::vector<uint8_t> run_and_save_to_memory(std::istringstream & test_idl, std::
     return reader.save_to_memory();
 }
 
+const std::string ConvertTypeToString(ElementType const& type)
+{
+    switch (type)
+    {
+    case ElementType::Void:
+        return "void";
+    case ElementType::Boolean:
+        return "Boolean";
+    case ElementType::Char:
+        return "Char16";
+    case ElementType::I1:
+        return "Int8";
+    case ElementType::U1:
+        return "UInt8";
+    case ElementType::I2:
+        return "Int16";
+    case ElementType::U2:
+        return "UInt16";
+    case ElementType::I4:
+        return "Int32";
+    case ElementType::U4:
+        return "UInt32";
+    case ElementType::I8:
+        return "Int64";
+    case ElementType::U8:
+        return "UInt64";
+    case ElementType::R4:
+        return "Single";
+    case ElementType::R8:
+        return "Double";
+    case ElementType::String:
+        return "String";
+    default:
+        return "";
+    }
+}
+
 void VerifyTypeRefName(std::string const& expectedName, std::string_view currentNamespace, TypeRef const& typeRef)
 {
     std::string combineName("");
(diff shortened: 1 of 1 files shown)


src/test/xmeta/test_xmeta_emit.cpp
```diff
@@ -350,23 +387,30 @@ void VerifyTypeRefName(std::string const& expectedName, std::string_view current
     REQUIRE(combineName == expectedName);
 }
 
-void VerifyReturnType(ElementType const& expectedType, std::string const& expectedTypeRef, MethodDef const& method, TypeDef const& parent)
+using value_type = std::variant<ElementType, coded_index<TypeDefOrRef>, GenericTypeIndex, GenericTypeInstSig, GenericMethodTypeIndex>;
```
DefaultRyan: This is a duplicate of `TypeSig::value_type`. Instead of duplicating it here, either use it directly, or define your local type in terms of it, aka `using value_type = TypeSig::value_type`.

src/test/xmeta/test_xmeta_emit.cpp
```diff
@@ -338,6 +338,43 @@ std::vector<uint8_t> run_and_save_to_memory(std::istringstream & test_idl, std::
...
+    case ElementType::R8:
+        return "Double";
+    case ElementType::String:
+        return "String";
+    default:
+        return "";
+    }
+}
+
 void VerifyTypeRefName(std::string const& expectedName, std::string_view currentNamespace, TypeRef const& typeRef)
```
DefaultRyan: Passing string inputs by `std::string_view` provides more flexibility than `std::string`.

DefaultRyan: Also, consider having this function take a `coded_index<TypeDefOrRef>` for flexibility.

src/test/xmeta/test_xmeta_emit.cpp
```diff
@@ -350,23 +387,30 @@ void VerifyTypeRefName(std::string const& expectedName, std::string_view current
...
+using value_type = std::variant<ElementType, coded_index<TypeDefOrRef>, GenericTypeIndex, GenericTypeInstSig, GenericMethodTypeIndex>;
+void VerifyTypeName(std::string const& expectedType, std::string_view currentNamespace, value_type const& type)
+{
+    if (std::holds_alternative<ElementType>(type))
+    {
+        REQUIRE(ConvertTypeToString(std::get<ElementType>(type)) == expectedType);
+    }
+    else
+    {
+        VerifyTypeRefName(expectedType, currentNamespace, std::get<coded_index<TypeDefOrRef>>(type).TypeRef());
```
DefaultRyan: The exception thrown by `std::get` may be cryptic to some. Since this is test code, prefer an explicit `REQUIRE(std::holds_alternative<coded_index<TypeDefOrRef>>(type));

src/test/xmeta/test_xmeta_emit.cpp
```diff
@@ -469,38 +503,26 @@ struct ExpectedEnum : ExpectedType
...
-        if (type == ElementType::End)
-        {
-            VerifyTypeRefName(typeRef, parentNamespace, std::get<coded_index<TypeDefOrRef>>(paramSig.Type().Type()).TypeRef());
-        }
-        else
-        {
-            REQUIRE(std::get<ElementType>(paramSig.Type().Type()) == type);
-        }
+
+        VerifyTypeName(typeRef, parentNamespace, paramSig.Type().Type());
```
DefaultRyan: This interface still lacks the capacity to check things we should already be checking:
- Is a parameter in or out?
- Is it passed by reference?
- Is it an array? (this also applies to return types)

Strongly recommend that instead of adding a pile of arguments to this function, you simply have a ToString for `MethodDef`, `MethodDefSig`, `ParamSig`, `RetTypeSig`, and `TypeSig`, so you only need to compare the expected and actual strings. Then we have something like this ability:
```
ExpectedMethod m{"SomeFunc", "void (Int32[] param1, out ref Object)"};
m.Verify(method_def);
```

DefaultRyan:
Are you planning to add the "ConvertToString" functionality for the other constructs? Adding it only for the fundamental types (int, float, string), misses the purpose of doing this in the first place: to allow the xmeta tests to easily scale up to more complex scenarios, while preserving the ability to check things that we should already have today. In other words, this really needs the ability to convert an entire `TypeSig` to a string, along with a MethodDefSig.

PR #462, opened 2019-06-04
Further macro isolation in C++/WinRT to prepare for C++20 modules

kennykerr:
Macros continue to be abused. C++ modules are a forcing function for libraries to be macro-free and consumers not to rely on library internals. In preparation for the (re)introduction of C++ module support in C++/WinRT, this update further isolates macro usage from consumers. Macros that are only used internally are now undefined and macros that must span headers but are only for internal use are renamed to WINRT_IMPL_XXX to underscore the fact that they are for internal use only. 

The WINRT_EXTERNAL_CATCH_CLAUSE macro will be removed once I coordinate the change with WIL.


Scottj1s:
I thought some of these macros were to isolate compiler differences (e.g., clang)

kennykerr:
No, they were originally to allow C++/WinRT to be cross-compiled before the xlang fork was created. For example (and ironically) Clang on Windows also defines `_MSC_VER`. 😄

PR #472, opened 2019-06-06
Fix bug in winrt::delegate return call generation

kennykerr:
Typo introduced by https://github.com/microsoft/xlang/pull/467.


kennykerr:
It has now. :)

PR #473, opened 2019-06-09
Remove magic statics from to_hstring and start using std::to_chars

DefaultRyan:


src/tool/cppwinrt/strings/base_string.h
```diff
@@ -289,10 +289,9 @@ namespace winrt::impl
         }
     };
 
-    inline _locale_t get_default_locale()
+    inline handle_type<locale_handle_traits> get_default_locale()
```
kennykerr: Can we remove this function now?

DefaultRyan: Ah, yeah. Good point. I think this may also remove our dependency on `<clocale>`.

PR #474, opened 2019-06-10
Add a test for cool

stephenatwork:
This adds a test_cool C# project and rebuilds the WinMD from IDL & the projection from the WinMD on changes.

The basic workflows work, but there are a few gotchas:
* it's awkward to get the VS developer environment initialized (see hardcoded call to $ENV{VS140COMNTOOLS})
* cmake handles generated files, but only if the list is known at _configuration time_. This is awkward for coolrt which doesn't know the list of generated files until _build time_.

### DIFF
--- a/src/test/cool/CMakeLists.txt
+++ b/src/test/cool/CMakeLists.txt
@@ -1,0 +1,76 @@
+
+project(test_cool LANGUAGES CSharp)
+
+# CMAKE doesn't have a good way to deal with tools which may
+# generate multiple files not known a priori. See
+# https://stackoverflow.com/questions/44076307/cmake-globbing-generated-files
+# We could try to extract this list from "ildasm /text foo.winmd".
+# For now, let's hardcode.
+set(generated_files
+    ${CMAKE_CURRENT_BINARY_DIR}/base-interop.cs
+    ${CMAKE_CURRENT_BINARY_DIR}/TestCool.BasicEnum.cs
+    ${CMAKE_CURRENT_BINARY_DIR}/TestCool.BasicStruct.cs
+    ${CMAKE_CURRENT_BINARY_DIR}/TestCool.Class.cs
+    ${CMAKE_CURRENT_BINARY_DIR}/TestCool.IClass.cs
+    ${CMAKE_CURRENT_BINARY_DIR}/TestCool.IInterface.cs)
+
+add_executable(test_cool "")
+target_sources(test_cool
+    PRIVATE main.cs
+    testcool.idl
+    ${generated_files}
+)
+set(CMAKE_CSharp_FLAGS "/langversion:7.1") # otherwise cmake defaults to 3
+set_target_properties(test_cool PROPERTIES VS_DOTNET_TARGET_FRAMEWORK_VERSION "4.5.1")
+set_target_properties(test_cool PROPERTIES VS_GLOBAL_LangVersion "7.1")
+set_target_properties(test_cool PROPERTIES VS_GLOBAL_AllowUnsafeBlocks "true")
+
+
+# Inputs and outputs. MIDL cannot handle / in paths.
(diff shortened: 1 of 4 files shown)


stephenatwork:
cc @dfields-msft

src/test/cool/CMakeLists.txt
```diff
@@ -0,0 +1,76 @@
...
+    ${CMAKE_CURRENT_BINARY_DIR}/TestCool.IInterface.cs)
+
+add_executable(test_cool "")
+target_sources(test_cool
+    PRIVATE main.cs
+    testcool.idl
+    ${generated_files}
+)
+set(CMAKE_CSharp_FLAGS "/langversion:7.1") # otherwise cmake defaults to 3
+set_target_properties(test_cool PROPERTIES VS_DOTNET_TARGET_FRAMEWORK_VERSION "4.5.1")
```
Scottj1s: should target .NET Standard 2.0 (not framework, which is obsolete as of 4.8)

src/test/cool/CMakeLists.txt
```diff
@@ -0,0 +1,76 @@
...
+
+# Find WindowsKitsRoot10 (from registry)
+get_filename_component(WindowsKitsRoot10 "[HKEY_LOCAL_MACHINE\\SOFTWARE\\Microsoft\\Windows Kits\\Installed Roots;KitsRoot10]" ABSOLUTE)
+if(NOT EXISTS "${WindowsKitsRoot10}")
+    message(FATAL_ERROR "Could not find the Windows SDK Root")
+endif()
+
+# Find WindowsSDKContractDir (from CMAKE_VS_WINDOWS_TARGET_PLATFORM_VERSION)
+file(TO_NATIVE_PATH
+    ${WindowsKitsRoot10}/References/${CMAKE_VS_WINDOWS_TARGET_PLATFORM_VERSION}/Windows.Foundation.FoundationContract/3.0.0.0
```
Scottj1s: this could be hard to maintain - why not just use the VS generator and allow it to automatically find the SDK, etc?

src/test/cool/CMakeLists.txt
```diff
@@ -0,0 +1,76 @@
...
+set_target_properties(test_cool PROPERTIES VS_DOTNET_TARGET_FRAMEWORK_VERSION "4.5.1")
+set_target_properties(test_cool PROPERTIES VS_GLOBAL_LangVersion "7.1")
+set_target_properties(test_cool PROPERTIES VS_GLOBAL_AllowUnsafeBlocks "true")
+
+
+# Inputs and outputs. MIDL cannot handle / in paths.
+file(TO_NATIVE_PATH ${CMAKE_CURRENT_SOURCE_DIR}\\testcool.idl   TestCool_idl)
+file(TO_NATIVE_PATH ${CMAKE_CURRENT_BINARY_DIR}\\testcool.winmd TestCool_winmd)
+
+if( NOT EXISTS "$ENV{VS140COMNTOOLS}")
```
Scottj1s: can't we let either [CMake figure this out](https://cmake.org/cmake/help/v3.14/generator/Visual%20Studio%2016%202019.html), or just use the VS generators?

src/test/cool/main.cs
```diff
@@ -0,0 +1,15 @@
+using System;
+namespace HelloWorld
+{
+    class Hello
+    {
+        static void Main()
```
Scottj1s: ideally this would be a unit test with junit xml output for Azure Devops build pipelines to consume.

Scottj1s:
Looks like wildcard cmake source detection [is supported, but discouraged](https://cmake.org/cmake/help/v3.4/command/aux_source_directory.html?highlight=aux_source_directory).

Scottj1s:
@stephenatwork, going to ask you to hold off on this PR.  The original author of this branch (devhawk/cool1) is no longer with Microsoft, and we're re-evaluating project structure, testing strategy, etc.

PR #477, opened 2019-06-12
add winrt::com_ptr, winrt::param::hstring, winrt::...::IUnknown to .natvis file

ChrisGuzak:
this is minor but aids debugging. 

I really want to fix winrt::implements too. I was able to improve WRL::RuntimeClass and think I might be able to apply that to this case but I need to play with it more. Scott, if you have ideas on that let me know.

### DIFF
--- a/src/package/cppwinrt/natvis/cppwinrt.natvis
+++ b/src/package/cppwinrt/natvis/cppwinrt.natvis
@@ -20,6 +20,12 @@
     <DisplayString Condition="this == 0">null</DisplayString>
   </Type>
   <!--Primitive type visualizers-->
+  <Type Name="winrt::com_ptr&lt;*&gt;">
+    <DisplayString>{m_ptr}</DisplayString>
+    <Expand>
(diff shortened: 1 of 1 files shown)


Scottj1s:
> this is minor but aids debugging.
> 
> I really want to fix winrt::implements too. I was able to improve WRL::RuntimeClass and think I might be able to apply that to this case but I need to play with it more. Scott, if you have ideas on that let me know.

Yes, been thinking the visualizer should support not just projected types but implementations too.  Unfortunately, that's all static natvis (not dynamic via GetRuntimeClassName, etc).  So it would be a fairly hefty natvis file, but they do parse fairly quickly.

Scottj1s:
winrt::param::hstring is fine

jonwis:
Consider adding some documentation on the wiki about how to use [msvc's link.exe -natvis](https://docs.microsoft.com/en-us/cpp/build/reference/natvis-add-natvis-to-pdb?view=vs-2019) option. Makes debugging magically better.

PR #488, opened 2019-06-24
Update linux docker image to Ubuntu 19.40 (disco) with Clang/Libc++ 8

Scottj1s:
This may be needed if PR #486 is not sufficient.


Scottj1s:
Even though the Ubuntu build has been decommissioned, I'll go ahead and complete this for future use.

PR #489, opened 2019-06-28
Incorporating xlang error handling into cppxlang

manodasanW:
- Updating the cppxlang projection to represent the state of error handling in xlang
- Addressing some PR feedback from the previous PR for error handling in xlang with how to represent HRESULT errors from functions whose interface can't change like QueryInterface by introducing com_interop_result


src/platform/error_abi.cpp
```diff
@@ -22,18 +22,30 @@ namespace xlang::impl
...
+            {
+                // Even if any of the string copies fail, we still have a xlang_result to represent the error.
+            }
         }
 
-        int32_t XLANG_CALL QueryInterface(xlang_guid const& id, void** object) noexcept final
+        com_interop_result XLANG_CALL QueryInterface(xlang_guid const& id, void** object) noexcept final
         {
-            if (id == xlang_unknown_guid)
+            if (!object)
```
kennykerr: Just AV here.

src/platform/error_abi.cpp
```diff
@@ -88,7 +102,8 @@ namespace xlang::impl
 
         void GetProjectionIdentifier(xlang_string* projection_identifier) noexcept override
         {
-            *projection_identifier = m_projection_identifier;
+            *projection_identifier = nullptr;
```
kennykerr: I think you can/should safely assume that copy_to_abi will clear the out param. If that's not the case then we should make it so.

src/platform/published/pal.h
```diff
@@ -147,9 +147,25 @@ extern "C"
...
 
+#ifdef __cplusplus
+    enum class com_interop_result : int32_t
+    {
+        success = 0,
+        no_interface = static_cast<int32_t>(0x80004002),
+        pointer = static_cast<int32_t>(0x80004003)
+    };
+#else
+    enum com_interop_result
```
kennykerr: This sounds like a very long-winded way of saying "HRESULT" :wink:

src/platform/published/pal.h
```diff
@@ -218,18 +234,18 @@ extern "C"
 #else
     enum xlang_result
```
kennykerr: Seems confusing that there both com_results and xlang_results.

src/test/abi_component/widget.cpp
```diff
@@ -70,11 +70,11 @@ struct widget_factory : iwidget_factory
         else
         {
             *object = nullptr;
-            return xlang_hresult_no_interface;
+            return com_interop_result::no_interface;
         }
 
         AddRef();
-        return 0;
+        return com_interop_result::success;
```
kennykerr: This wouldn't be necessary if the return type was simply int32_t. Seems overly verbose for an ABI.

src/tool/cppxlang/code_writers.h
```diff
@@ -1074,15 +1074,22 @@ namespace xlang
             if constexpr (std::is_base_of_v<Foundation::IUnknown, V>)
             {
                 V result{ nullptr };
-                XLANG_SHIM(Foundation::Collections::IMapView<K, V>)->Lookup(get_abi(key), put_abi(result));
+                com_ptr<xlang_error_info> error_info{
```
kennykerr: The Lookup call's result is intentionally ignored. It should probably stay that way in xlang due to the "Try" nature of this helper.

kennykerr: Oh, nevermind. You're doing the right thing by capturing and destroying any error object.

src/tool/cppxlang/strings/base_composable.h
```diff
@@ -41,7 +41,7 @@ namespace xlang::impl
 
             if (shim().outer())
             {
-                check_hresult(shim().QueryInterface(guid_of<I>(), &result));
+                check_com_interop_result(shim().QueryInterface(guid_of<I>(), &result));
```
kennykerr: check_com_interop_result vs check_xlang_error seems inconsistent. e.g. xxx_result vs xxx_error

src/tool/cppxlang/strings/base_meta.h
```diff
@@ -1,9 +1,8 @@
 
 namespace xlang
 {
-    void check_hresult(hresult const result);
-    void check_hresult(xlang_error_info* result);
-    hresult to_hresult() noexcept;
+    void check_com_interop_result(com_interop_result const result);
```
kennykerr: Here's the inconsistency.

src/tool/cppxlang/strings/base_natvis.h
```diff
@@ -31,11 +31,11 @@ namespace xlang::impl
             }
             value{};
             guid iid{};
-            if (XLANG_IIDFromString(iid_str, &iid) == error_ok)
+            if (XLANG_IIDFromString(iid_str, &iid) == 0)
```
kennykerr: Zero here and bla::success below. Zero is so much more concise.

src/platform/published/pal.h
kennykerr: To be clear, I'm not saying there's no merit for this just that we may want to bury the HRESULT in the ABI and just stick with a handful of int32_t values where absolutely necessary (e.g. QI).

src/platform/error_abi.cpp
BenJKuhn: agreed. Since it has no failure mode, it must always set its out parameter to a stable value.

src/platform/published/pal.h
BenJKuhn: This is a side effect of the long conversation we had about reducing & simplifying the xlang error codes to a new set of values that don't align with the COM values. It's useful to note, then that the ABI for the first three methods is actually COM-compatible including failure values. It's a little verbose but nicely explicit.

src/platform/error_abi.cpp
manodasanW: copy_to_abi seems to taken in a reference and asserts that the passed reference == nullptr.

src/tool/cppxlang/strings/base_meta.h
manodasanW: addressed.

src/tool/cppxlang/strings/base_composable.h
manodasanW: addressed.

src/platform/error_abi.cpp
manodasanW: addressed.

src/platform/published/pal.h
manodasanW: The reason for this change was to avoid confusion on what error is returned here.  We have an xlang error for no interface and an COM error for no interface, both of which have different values.  To keep compat with COM for IUnknown, we made the choice to keep this as a COM error.  If we were to represent that as an int32, it can introduce confusion to a developer on what to return here as the first instinct would be to return / check for the xlang version of no interface as we are in xlang.  But, they have to actually return the COM version of the error.  To make that clear, the int32 was changed to com_interop_result.  I do agree that it is a little verbose, but this should really only affect QueryInterface and related functions.

kennykerr: Sure, I just think that if anyone is actually implementing QueryInterface and thus having to figure this out they're either one of us or we've failed because nobody should be dealing with the ABI directly.

manodasanW: I agree no one else should probably be dealing with the ABI directly.  This change also makes sure we ourselves don't accidentally use the wrong error value when we return from the com interop functions by essentially having a compile time safety check.  In addition, in cross platform scenarios, it ensures that we don't introduce a new com error without explicitly being aware we are adding a new error for interop scenarios and providing a mapping to an xlang error for it.

In theory, we can make returning success less verbose by having a define S_OK or something similar as com_interop_result::success and returning that instead.  What do you think about that?

kennykerr: Macros are evil. Let's leave it as is. :shipit:

manodasanW:
Build status not being reflected, but it passed.

PR #490, opened 2019-06-28
Some basic Clang 8 conformance for C++/WinRT

kennykerr:
Some minor tweaks after manual testing. Mainly `impl::wait_for` and `impl::wait_get` being used before they are declared.

Tested with:  `-fno-delayed-template-parsing  -fno-ms-compatibility`

Also added support for using coroutines with Clang. Since Clang uses a different coroutine ABI compared with Visual C++, a replacement for the <coroutine> header is required.

Also some build cleanup to simplify Clang testing.

Need to get a nightly Clang build going...

### DIFF
--- a/src/tool/cppwinrt/strings/base_error.h
+++ b/src/tool/cppwinrt/strings/base_error.h
@@ -185,7 +185,7 @@ namespace winrt
 #endif
 
         impl::bstr_handle m_debug_reference;
-        uint32_t const m_debug_magic{ 0xAABBCCDD };
+        uint32_t m_debug_magic{ 0xAABBCCDD };
         hresult m_code{ impl::error_fail };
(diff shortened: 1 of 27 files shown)


src/tool/cppwinrt/strings/base_error.h
```diff
@@ -185,7 +185,7 @@ namespace winrt
 #endif
 
         impl::bstr_handle m_debug_reference;
-        uint32_t const m_debug_magic{ 0xAABBCCDD };
+        uint32_t m_debug_magic{ 0xAABBCCDD };
```
BenJKuhn: BTW, what is this magic number for?

kennykerr: This magic number is used by the VS debugger as a sanity check before it attempts to reach in and grab the debug reference string to light up restricted error information.

PR #491, opened 2019-06-30
Use ResolveProjectReferences target and limit WinMDs passed to mdmerge.exe

jlaanstra:
This PR fixes a few issues with the way CppWinRT handles project references, that prevent certain project types from not being included in the reference projection or causes duplicate type errors because the winmd ends up being referenced twice:

- When a Runtime Component and a Static Library reference the same NuGet package containing a WinMD file, it will be referenced twice because the WinMD will always be copied to the output folder. 
- A CX project referenced by a CppWinRT project won't have it's WinMD included in the reference projection.
- You have to set project to project references to not copy the reference locally in the properties window to prevent duplicate type errors.

To fix those problems I made two changes.
1. Hook into the ResolveProjectReferences target instead of ResolvedLinkLib.
2. Copy the reference projection into a Metadata folder using hard links if possible instead of using wildcards into the project output folders.


jlaanstra:
Seems like I have some whitespace changes that I will revert.

jlaanstra:
@kennykerr probably want to wait for @scottj1s to take a look before merging.

kennykerr:
Yes, Scott will be back in a week. This can wait.

src/package/cppwinrt/nuget/Microsoft.Windows.CppWinRT.targets
```diff
@@ -319,17 +250,15 @@ $(XamlMetaDataProviderPch)
...
                 <WinMDPath>%(Midl.OutputDirectory)%(Midl.MetadataFileName)</WinMDPath>
             </_MdMergeInputs>
-            <_MdMergeInputs Include="@(CppWinRTStaticWinMDReferences)"/>
+            <_MdMergeInputs Include="@(CppWinRTStaticLibraryProjectWinMDReferences)"/>
             <_MdMergeReferences Remove="@(_MdMergeReferences)" />
             <_MdMergeReferences Include="@(CppWinRTDirectWinMDReferences)" />
-            <_MdMergeReferences Include="@(CppWinRTDynamicWinMDReferences)" />
+            <_MdMergeReferences Include="@(CppWinRTDynamicLibraryProjectWinMDReferences)" />
             <_MdMergeReferences Include="@(CppWinRTPlatformWinMDReferences)" />
-            <_MdMergeRefsDistinct Remove="@(_MdMergeRefsDistinct)" />
```
Scottj1s: mdmerge can't handle duplicate winmd files - why is this no longer needed?

src/package/cppwinrt/nuget/Microsoft.Windows.CppWinRT.targets
```diff
@@ -319,17 +250,15 @@ $(XamlMetaDataProviderPch)
...
-            <_MdMergeRefsDistinct Include="@(_MdMergeReferences->'%(RelativeDir)'->Distinct())" />
         </ItemGroup>
         <PropertyGroup>
-          <_MdMergeParameters>-v @(_MdMergeRefsDistinct->'-metadata_dir &quot;%(RelativeDir).&quot;', '&#x0d;&#x0a;')</_MdMergeParameters>
-          <_MdMergeParameters>$(_MdMergeParameters) -o &quot;$(CppWinRTMergedDir)&quot; -i &quot;$(CppWinRTUnmergedDir)&quot; -partial $(_MdMergeDepth)</_MdMergeParameters>
+            <_MdMergeParameters>-v -metadata_dir &quot;$(CppWinRTMetadataDir)&quot;</_MdMergeParameters>
```
Scottj1s: ah I think I see - just a single metadata_dir?

src/package/cppwinrt/nuget/Microsoft.Windows.CppWinRT.targets
```diff
@@ -338,14 +267,24 @@ $(XamlMetaDataProviderPch)
...
             WriteOnlyWhenDifferent="true" />
-        <Copy SourceFiles="@(_MdMergeInputs->'%(WinMDPath)')" DestinationFolder="$(CppWinRTUnmergedDir)\" />
+        <Copy UseHardlinksIfPossible="$(CppWinRTUseHardlinksIfPossible)"
+            SkipUnchangedFiles="$(CppWinRTSkipUnchangedFiles)"
+            SourceFiles="@(_MdMergeInputs->'%(WinMDPath)')"
+            DestinationFolder="$(CppWinRTUnmergedDir)" />
+        <Copy UseHardlinksIfPossible="$(CppWinRTUseHardlinksIfPossible)"
+            SkipUnchangedFiles="$(CppWinRTSkipUnchangedFiles)"
+            SourceFiles="@(_MdMergeReferences->'%(WinMDPath)')" 
+            DestinationFolder="$(CppWinRTMetadataDir)\" />
```
Scottj1s: ah, and the level of indirection thru CppWinRTMetadataDir with SkipUnchangedFiles - effectively eliminates dupes

src/package/cppwinrt/nuget/Microsoft.Windows.CppWinRT.targets
```diff
@@ -270,17 +201,17 @@ $(XamlMetaDataProviderPch)
     <!--Insert Midl /references to Platform WinMDs, library reference WinMDs, and direct reference WinMDs-->
     <Target Name="CppWinRTSetMidlReferences"
             BeforeTargets="Midl"
-            DependsOnTargets="GetCppWinRTPlatformWinMDReferences;CppWinRTAddXamlMetaDataProviderIdl;GetCppWinRTStaticWinMDReferences;GetCppWinRTDynamicWinMDReferences;GetCppWinRTDirectWinMDReferences;$(CppWinRTSet_MidlReferencesDependsOn)">
+            DependsOnTargets="GetCppWinRTPlatformWinMDReferences;CppWinRTAddXamlMetaDataProviderIdl;GetCppWinRTProjectWinMDReferences;GetCppWinRTDirectWinMDReferences;$(CppWinRTSet_MidlReferencesDependsOn)">
```
Scottj1s: I don't see where GetCppWinRTProjectWinMDReferences is defined

src/package/cppwinrt/nuget/Microsoft.Windows.CppWinRT.targets
kennykerr: XML at it's best. :wink:

src/package/cppwinrt/nuget/Microsoft.Windows.CppWinRT.targets
Scottj1s: or ResolveProjectReferences hooked for that matter

Scottj1s:
Lots of tweaks (thanks!) - how has this been tested?  Both VS 2017 and 2019, with various p2p reference scenarios?

src/package/cppwinrt/nuget/Microsoft.Windows.CppWinRT.targets
jlaanstra: We copy the files to a separate metadata dir now.

src/package/cppwinrt/nuget/Microsoft.Windows.CppWinRT.targets
jlaanstra: Correct.

src/package/cppwinrt/nuget/Microsoft.Windows.CppWinRT.targets
jlaanstra: We also only copy the reference instead of using a wildcard into the output folder which elminates dupes as well.

jlaanstra:
I tested this on the YourPhone solution with 34 projects. It has:
- Static lib -> Nuget package
- Dynamic lib -> NuGet package
- Dynamic lib -> Static lib
- Dynamic lib -> Dynamic lib
- CX -> WinRT
- WinRT -> CX

src/package/cppwinrt/nuget/Microsoft.Windows.CppWinRT.targets
jlaanstra: Looks like it got removed in the second iteration. I'll add it back.

Scottj1s:
Great - thank you!

---
In reply to: [509732892](https://github.com/microsoft/xlang/pull/491#issuecomment-509732892) [](ancestors = 509732892)

src/package/cppwinrt/nuget/Microsoft.Windows.CppWinRT.targets
```diff
@@ -130,77 +134,30 @@ Copyright (C) Microsoft Corporation. All rights reserved.
...
-            <CppWinRTDynamicWinMDReferences Include="@(_CppWinRTDynamicWinMDReferences);@(_CppWinRTManagedWinMDReferences)">
-                <WinMDPath>%(Identity)</WinMDPath>
-            </CppWinRTDynamicWinMDReferences>
+            <!-- Get static library project references -->
+            <_CppWinRTStaticProjectReferences Remove="@(_CppWinRTStaticProjectReferences)"/>
+            <_CppWinRTStaticProjectReferences Include="@(_ResolvedProjectReferencePaths)"
+                Condition= "'%(_ResolvedProjectReferencePaths.ProjectType)'=='StaticLibrary' AND '%(_ResolvedProjectReferencePaths.WinMDFile)' == 'true'"/>
+            <CppWinRTStaticLibraryProjectWinMDReferences Remove="@(CppWinRTStaticLibraryProjectWinMDReferences)"/>
+            <CppWinRTStaticLibraryProjectWinMDReferences Include="@(_CppWinRTStaticProjectReferences)">
+                <WinMDPath>%(FullPath)</WinMDPath>
```
Scottj1s: isn't this going to create a path to lib file, not a winmd?  (see original _CppWinRTStaticWinMDReferenceWildcard)

Scottj1s: oh, that's the _ResolvedProjectReferencePaths.WinMDFile filter?

---
In reply to: [301731509](https://github.com/microsoft/xlang/pull/491#discussion_r301731509) [](ancestors = 301731509)

jlaanstra: That's not what I've seen, but that is just one data point. I can try this in another project to verify.

src/package/cppwinrt/nuget/Microsoft.Windows.CppWinRT.targets
```diff
@@ -130,77 +134,30 @@ Copyright (C) Microsoft Corporation. All rights reserved.
...
+            </CppWinRTStaticLibraryProjectWinMDReferences>
+            <!--Get dynamic library project references-->
+            <_CppWinRTDynamicProjectReferences Remove="@(_CppWinRTDynamicProjectReferences)"/>
+            <_CppWinRTDynamicProjectReferences Include="@(_ResolvedProjectReferencePaths)"
+                Condition= "'%(_ResolvedProjectReferencePaths.ProjectType)'!='StaticLibrary'  AND '%(_ResolvedProjectReferencePaths.WinMDFile)' == 'true'"/>
+            <CppWinRTDynamicLibraryProjectWinMDReferences Remove="@(CppWinRTDynamicLibraryProjectWinMDReferences)"/>
+            <CppWinRTDynamicLibraryProjectWinMDReferences Include="@(_CppWinRTDynamicProjectReferences)">
+                <WinMDPath>%(FullPath)</WinMDPath>
```
Scottj1s: without wildcard support, how is a dynamic project reference with multiple winmds handled? e.g., when the project has CppWinRTNamespaceMergeDepth set to > 1?  (some users require that for publishing specific namespace winmds)

jlaanstra: Do you have an example of this? I was definitely under the assumption that a single project outputs a single WinMD.

Scottj1s: That's true (and required) for Xaml projects, but we have customers that specifically require multiple winmds, not merged.  Easy enough to test with a couple nested namespaces under the project root namespace -> root.a.winmd, root.b.winmd -> an app should be able to project and consume classes from both

src/package/cppwinrt/nuget/Microsoft.Windows.CppWinRT.targets
jlaanstra: This is now guaranteed with the CppWInRTGetResolvedWinMD target.

src/test/package/nuget/NuGetTest/TestRuntimeComponent2/module.cpp
```diff
@@ -0,0 +1,56 @@
...
+        return false;
+    }
+
+    winrt::clear_factory_cache();
+    return true;
+}
+
+void* __stdcall winrt_get_activation_factory(std::wstring_view const& name)
+{
+    void* factory = TestRuntimeComponent2_get_activation_factory(name);
```
jlaanstra: @kennykerr Is this what you were thinking with -lib? I'm getting some identifier not found error which makes sense as I don't have a separate definition, but wondering how you imagined this to work.

src/test/package/nuget/NuGetTest/Directory.Build.props
```diff
@@ -0,0 +1,27 @@
+<?xml version="1.0" encoding="utf-8"?>
```
Scottj1s: Nice!  I wasn't aware of this technique

src/test/package/nuget/NuGetTest/TestApp/packages.config
```diff
@@ -0,0 +1,4 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<packages>
+  <package id="Microsoft.Windows.CppWinRT" version="2.0.190712.2" targetFramework="native" />
```
Scottj1s: This will test the published NuGet package version N-1, so regressions could still slip thru.  Instead, we should remove all the packages.config files and restructure all the vcxproj files to import Microsoft.Windows.CppWinRT.props and Microsoft.Windows.CppWinRT.targets via relative paths "as if" the Nuget package had been added.  I did this originally for the fast forwarder test project - see 
src/tool/cppwinrt/test_fast_fwd/test_fast_fwd.vcxproj in PR #344

Scottj1s:
great test collateral - just need to point to props/targets in repo instead of using last published Nuget.  

I was just discussing with Kenny scripting the creation of something like this from project templates (also testing the VSIX).  Seems doable with dotnet CLI, which can create projects, references, etc - but we can do that later.

src/test/package/nuget/NuGetTest/Directory.Build.props
jlaanstra: I can :)

src/test/package/nuget/NuGetTest/Directory.Build.props
```diff
@@ -0,0 +1,25 @@
...
+        <!-- Build overrides the roots. -->
+        <GenerateProjectSpecificOutputFolder>false</GenerateProjectSpecificOutputFolder>
+        <OutDirWasSpecified>true</OutDirWasSpecified>
+        <OutDirRoot Condition="'$(OutDirRoot)' == ''" >$(SolutionDir)bin\$(Platform)\$(Configuration)\</OutDirRoot>
+        <OutDir>$(OutDirRoot)$(MSBuildProjectName)\</OutDir>
+        <IntDirRoot Condition="'$(IntDirRoot)' == ''" >$(SolutionDir)obj\$(Platform)\$(Configuration)\</IntDirRoot>
+        <IntDir>$(IntDirRoot)$(MSBuildProjectName)\</IntDir>
+        <GeneratedFilesDir>$(IntDir)Generated Files\</GeneratedFilesDir>
+        <CppWinRTVerbosity>high</CppWinRTVerbosity>
+        <CppWinRTPath>..\..\..\..\..\..\_build\Windows\x64\Debug\tool\cppwinrt\</CppWinRTPath>
```
Scottj1s: >x64\ [](start = 55, length = 4)

$(Platform)\$(Configuration) ?

src/test/package/nuget/NuGetTest/Directory.Build.targets
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project TreatAsLocalProperty="OutDir;IntDir" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+
+    <!-- This file is automatically imported by MSBuild for each project in under this folder, 
+         see https://docs.microsoft.com/en-us/visualstudio/msbuild/customize-your-build -->
+
+    <Import Condition="Exists($([MSBuild]::GetPathOfFileAbove('Directory.Build.targets', '$(MSBuildThisFileDirectory)../')))" Project="$([MSBuild]::GetPathOfFileAbove('Directory.Build.targets', '$(MSBuildThisFileDirectory)../'))" />
```
Scottj1s: >../ [](start = 117, length = 3)

I think this is a mistake (same for second one).  It's mentioned on the docs.microsoft.com page above, where it's also a mistake.  In my experiments of chaining directory.build.* file imports, the GetPathOfFileAbove does what you want and expect - the ".." is redundant, and maybe incorrect - have you tested?

src/test/package/nuget/NuGetTest/NuGetTest.sln
```diff
@@ -0,0 +1,179 @@
...
+EndProject
+Project("{8BC9CEB8-8B4A-11D0-8D11-00A0C91BC942}") = "TestStaticLibrary1", "TestStaticLibrary1\TestStaticLibrary1.vcxproj", "{2158F418-CA97-4599-8103-EFC133850BAA}"
+EndProject
+Project("{8BC9CEB8-8B4A-11D0-8D11-00A0C91BC942}") = "TestStaticLibrary2", "TestStaticLibrary2\TestStaticLibrary2.vcxproj", "{0011F69F-2363-4DFD-B02A-1E7E909BCE89}"
+EndProject
+Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Solution Items", "Solution Items", "{41E05048-C8B0-4AB7-BE1B-D8D53E60A5E4}"
+	ProjectSection(SolutionItems) = preProject
+		..\..\..\..\package\cppwinrt\nuget\Microsoft.Windows.CppWinRT.props = ..\..\..\..\package\cppwinrt\nuget\Microsoft.Windows.CppWinRT.props
+		..\..\..\..\package\cppwinrt\nuget\Microsoft.Windows.CppWinRT.targets = ..\..\..\..\package\cppwinrt\nuget\Microsoft.Windows.CppWinRT.targets
+	EndProjectSection
```
Scottj1s: may want to also drop the directory.build.* files in here too, to highlight them and as a convenience for maintenance

src/test/package/nuget/NuGetTest/TestRuntimeComponentCX/TestRuntimeComponentCX.vcxproj
```diff
@@ -0,0 +1,288 @@
...
+    <ApplicationType>Windows Store</ApplicationType>
+    <WindowsTargetPlatformVersion>10.0.18362.0</WindowsTargetPlatformVersion>
+    <WindowsTargetPlatformMinVersion>10.0.18362.0</WindowsTargetPlatformMinVersion>
+    <ApplicationTypeRevision>10.0</ApplicationTypeRevision>
+  </PropertyGroup>
+  <Import Project="$(VCTargetsPath)\Microsoft.Cpp.Default.props" />
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|Win32'" Label="Configuration">
+    <ConfigurationType>DynamicLibrary</ConfigurationType>
+    <UseDebugLibraries>true</UseDebugLibraries>
+    <PlatformToolset>v142</PlatformToolset>
```
Scottj1s: Can you edit these to:
    <PlatformToolset>v140</PlatformToolset>
    <PlatformToolset Condition="'$(VisualStudioVersion)' == '15.0'">v141</PlatformToolset>
    <PlatformToolset Condition="'$(VisualStudioVersion)' == '16.0'">v142</PlatformToolset>

as with the C++ projects, to enable building in both VS 2017 and 2019?

src/test/package/nuget/NuGetTest/TestStaticLibrary2/TestStaticLibrary2.vcxproj
```diff
@@ -0,0 +1,230 @@
...
+    </ProjectConfiguration>
+    <ProjectConfiguration Include="Release|x64">
+      <Configuration>Release</Configuration>
+      <Platform>x64</Platform>
+    </ProjectConfiguration>
+  </ItemGroup>
+  <PropertyGroup Label="Configuration">
+    <ConfigurationType>StaticLibrary</ConfigurationType>
+    <PlatformToolset>v141</PlatformToolset>
+    <PlatformToolset Condition="'$(VisualStudioVersion)' == '16.0'">v142</PlatformToolset>
```
Scottj1s: TestStaticLibrary4 needs this too

src/test/package/nuget/NuGetTest/Directory.Build.targets
jlaanstra: It works this way, but it might be redundant.

PR #492, opened 2019-07-03
Major Clang update for C++/WinRT

kennykerr:
This turned from a minor update to a pretty major update as I decided to finally get all of the C++/WinRT unit tests to build with Clang. And we have a lot of tests...

Clang support has always been mostly a best effort by smoke testing with simple apps. This was mainly done in support of conformance. Being able to compile all our unit tests means that we can do much much better. Notably with this update, you can finally use C++ coroutines with Clang. This has some interesting possibilities as Clang has a much more advanced optimizer when it comes to coroutines. 

Beyond coroutine support, I've fixed a number of conformance issues highlighted by Clang and done a lot of work to clean up the build support to make it simpler and easier to build with either VC or Clang. It's not quite warning-free for Clang just yet, but its a lot closer. I've dealt with all the errors and most of the warnings. More to come as I can now far more easily test with Clang.

### DIFF
--- a/src/tool/cppwinrt/strings/base_coroutine.h
+++ b/src/tool/cppwinrt/strings/base_coroutine.h
@@ -1,0 +1,125 @@
+
+#ifndef __clang__
+
+#include <experimental/coroutine>
+
+#else
+
+namespace std::experimental
+{
+    template <typename R, typename...> struct coroutine_traits
+    {
+        using promise_type = typename R::promise_type;
+    };
+
+    template <typename Promise = void> struct coroutine_handle;
+
+    template <> struct coroutine_handle<void>
+    {
+        coroutine_handle(decltype(nullptr)) noexcept
+        {
+        }
+
+        coroutine_handle() noexcept
+        {
+        }
+
+        static coroutine_handle from_address(void* address) noexcept
+        {
+            coroutine_handle result;
+            result.ptr = address;
+            return result;
+        }
+
+        coroutine_handle& operator=(decltype(nullptr)) noexcept
+        {
+            ptr = nullptr;
+            return *this;
+        }
+
+        explicit operator bool() const noexcept
+        {
+            return ptr;
+        }
+
+        void* address() const noexcept
+        {
+            return ptr;
+        }
+
+        void operator()()
+        {
+            resume();
+        }
+
+        void resume() const
+        {
+            __builtin_coro_resume(ptr);
+        }
+
+        void destroy() const
+        {
+            __builtin_coro_destroy(ptr);
+        }
+
+        bool done() const
+        {
+            return __builtin_coro_done(ptr);
+        }
+
+    protected:
+        void* ptr{};
+    };
+
+    template <typename Promise> struct coroutine_handle : coroutine_handle<>
+    {
+        using coroutine_handle<>::operator=;
+
+        static coroutine_handle from_address(void* address) noexcept
+        {
+            coroutine_handle result;
+            result.ptr = address;
+            return result;
+        }
+
+        Promise& promise() const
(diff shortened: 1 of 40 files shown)


src/tool/cppwinrt/strings/base_identity.h
```diff
@@ -224,8 +224,12 @@ namespace winrt::impl
     template <typename T>
     struct name
     {
+#ifdef __clang__
+        inline static const auto value
```
kennykerr: This is as a result of guid_of not always being a constexpr expression when the type is a classic COM interface. See next comment.

src/tool/cppwinrt/strings/base_meta.h
```diff
@@ -136,7 +136,11 @@ namespace winrt::impl
     template <typename T>
     struct guid_storage
     {
+#ifdef __clang__
+        inline static const guid value{ __uuidof(T) };
```
kennykerr: While Clang generally honors MS compatibility, it doesn't extend to __uuidof. VC treats it as a constexpr expression but Clang does not. Fortunately, this only impacts classic COM interfaces and not WinRT types that do not depend on __uuidof.

src/tool/cppwinrt/test_component/test_auto.cpp
```diff
@@ -14,5 +14,5 @@ void test_auto()
 
     Class::StaticTestReturn();
 
-    auto discarded = Class::StaticProperty(); discarded;
+    [[maybe_unused]] auto discarded = Class::StaticProperty();
```
kennykerr: Clang isn't fooled...

src/tool/cppwinrt/strings/base_coroutine.h
```diff
@@ -0,0 +1,125 @@
+
+#ifndef __clang__
```
kennykerr: Clang uses a completely different ABI for coroutine support. This means that the VC <experimental/coroutine> header does not support the Clang frontend.

src/tool/cppwinrt/strings/base_coroutine.h
```diff
@@ -0,0 +1,125 @@
...
+
+#include <experimental/coroutine>
+
+#else
+
+namespace std::experimental
```
BenJKuhn: Since there's a compiler difference that you're smoothing over, it seems likely that projects could accidentally take a dependency on your adapter layer. Would it make sense to put the adapter in a cppwinrt::details namespace, then in the VC version, just do namespace cppwinrt::details { using std::experimental; }, then consume from there so that clang projects using cppwinrt can't easily take an accidental dependency on something that you might want to change later once things standardize a bit more?

src/tool/cppwinrt/strings/base_meta.h
BenJKuhn: This seems worthy of a compatibility note in the code.

src/tool/cppwinrt/strings/base_coroutine.h
kennykerr: The trick is that everything here is part of the public library support for coroutines. This is intentionally designed for libraries to take a dependency on and is unavoidable. As such, care is taken that this accurately represents the same library surface as the VC header and the C++20 standard - specifically nothing *more* is provided. The only implementation detail is the implementation of the functions that call the Clang intrinsics.

src/tool/cppwinrt/strings/base_meta.h
kennykerr: Will add a comment. :thumbsup:

kennykerr:
Thanks Ben!

PR #497, opened 2019-07-05
Optimize C++/WinRT's GetMany implementation

kennykerr:
The `GetMany` implementation for `IIterable<T>` was performing multiple traversals of the range when random access iterators weren't available. This update ensures that only a single traversal occurs in all cases and memcpy is used in more cases when the range offers random access.

I've also moved the scratch test into a separate project as the tests have become too extensive.

### DIFF
--- a/src/tool/cppwinrt/strings/base_collections_base.h
+++ b/src/tool/cppwinrt/strings/base_collections_base.h
@@ -103,7 +103,7 @@ namespace winrt
 
             uint32_t GetMany(array_view<T> values)
             {
-                if constexpr (std::is_same_v<decltype(std::iterator_traits<iterator_type>::iterator_category()), std::random_access_iterator_tag>)
+                if constexpr (std::is_same_v<decltype(typename std::iterator_traits<iterator_type>::iterator_category()), std::random_access_iterator_tag>)
                 {
                     uint32_t const actual = (std::min)(static_cast<uint32_t>(m_end - m_current), values.size());
                     m_owner->copy_n(m_current, actual, values.begin());
--- a/src/tool/cppwinrt/strings/base_collections_base.h
+++ b/src/tool/cppwinrt/strings/base_collections_base.h
@@ -103,17 +103,35 @@ namespace winrt
 
             uint32_t GetMany(array_view<T> values)
             {
-                uint32_t const actual = (std::min)(static_cast<uint32_t>(std::distance(m_current, m_end)), values.size());
-                m_owner->copy_n(m_current, actual, values.begin());
-                std::advance(m_current, actual);
-                return actual;
+                if constexpr (std::is_same_v<decltype(std::iterator_traits<iterator_type>::iterator_category()), std::random_access_iterator_tag>)
+                {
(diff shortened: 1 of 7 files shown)


BenJKuhn:
This seems likely to cause a lot of pain for ingestion, esp. if those includes have been around for a while. Might I suggest putting them behind a macro that you can enable / disable, so that you (and other customers) can enable it incrementally?

src/tool/cppwinrt/strings/base_collections_base.h
```diff
@@ -103,17 +103,35 @@ namespace winrt
 
             uint32_t GetMany(array_view<T> values)
             {
-                uint32_t const actual = (std::min)(static_cast<uint32_t>(std::distance(m_current, m_end)), values.size());
-                m_owner->copy_n(m_current, actual, values.begin());
-                std::advance(m_current, actual);
-                return actual;
+                if constexpr (std::is_same_v<decltype(typename std::iterator_traits<iterator_type>::iterator_category()), std::random_access_iterator_tag>)
```
DefaultRyan: `std::is_same` is the wrong way to light up behavior for random access iterators, as it will improperly reject C++20's `contiguous_iterator_tag`. The iterator tags have an inheritance chain, basically representing the strength of the iterator guarantees - `continuous_iterator_tag` inherits from `random_access_iterator_tag`, which inherits from `bidirectional_iterator_tag`, and so on.

What you really need to do here is use overload dispatch to specialize the behavior based on iterator "strength", as in the example at https://en.cppreference.com/w/cpp/iterator/iterator_tags.

src/tool/cppwinrt/strings/base_collections_base.h
```diff
@@ -103,17 +103,35 @@ namespace winrt
...
+                    auto output = values.begin();
+
+                    while (output < values.end() && m_current != m_end)
+                    {
+                        *output = Current();
+                        ++output;
+                        ++m_current;
+                    }
+
+                    return static_cast<uint32_t>(output - values.begin());
```
DefaultRyan: Directly subtracting iterators is an operation that is only valid for `random_access_iterator_tag`, which this 'else' clause is explicitly not supposed to use. Do you have test coverage exercising this branch?

kennykerr: Yes, this is heavily tested. In fact, I just added even more rigorous tests for this branch. #499

kennykerr: I think the confusion here is that the subtraction is not on the input but on the output. ;)

src/tool/cppwinrt/strings/base_collections_base.h
kennykerr: Ah good idea. I'll do that.

src/tool/cppwinrt/strings/base_collections_base.h
DefaultRyan: Oh yeah, that's totally safe to do on the output. Thanks for pointing that out. 👍

PR #499, opened 2019-07-08
Improved GetMany test for C++/WinRT

kennykerr:
Updated the test to use an iterator that may only be advanced once, ensuring the GetMany complexity optimization is actually enforced with this test. This updates #497


src/tool/cppwinrt/test/GetMany.cpp
```diff
@@ -15,33 +15,51 @@ using namespace Windows::Foundation::Collections;
...
             }
 
         private:
-
-            std::list<T> m_values;
+            generator_container m_container;
         };
 
-        return make<list>(std::move(values));
+        auto v = single_threaded_vector<T>(std::move(values));
```
DefaultRyan: Add the ability to do this for a container that is definitely not random-access, like `std::list`.

kennykerr: That's what `generator_container` ensures as it only offers an `IIterator<T>`. I initially used `std::list` alone and while that ensures that random access is not available, it doesn't ensure that multiple traversals aren't employed. Here a vector is used merely for convenience to create the container that the iterator is backed by.

PR #500, opened 2019-07-08
Add fast reverse iterators to C++/WinRT

### LINKED ISSUE #141: Containers like IVector/IObservableVector should support reverse iterators
I was recommending to a colleague that they use an STL algorithm instead of a manual for loop but in their case they were searching backwards so they couldn't use rbegin/rend with std::find_first_of. 

Please add rbegin/rend support to all the random access containers in C++/WinRT. Thanks!

kennykerr:
Add `rbegin` and `rend` ADL functions to complement `begin` and `end` and provide reverse iterators for `IVector<T>` and `IVectorView<T>`. 

Fixes #141

### DIFF
--- a/src/tool/cppwinrt/strings/base_collections.h
+++ b/src/tool/cppwinrt/strings/base_collections.h
@@ -41,8 +41,50 @@ namespace winrt::impl
 
     private:
 
-        T const* m_collection = nullptr;
-        uint32_t m_index = 0;
+        T const* m_collection{};
+        uint32_t m_index{};
+    };
+
+    template <typename T>
+    struct rfast_iterator
+    {
+        using iterator_category = std::input_iterator_tag;
+        using value_type = T;
+        using difference_type = ptrdiff_t;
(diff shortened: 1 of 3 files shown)


src/tool/cppwinrt/strings/base_collections.h
```diff
@@ -41,8 +41,50 @@ namespace winrt::impl
...
+        using difference_type = ptrdiff_t;
+        using pointer = T *;
+        using reference = T &;
+
+        rfast_iterator(T const& collection, int64_t const index) noexcept :
+            m_collection(&collection),
+            m_index(index)
+        {}
+
+        rfast_iterator& operator++() noexcept
```
BenJKuhn: any reason not to make this bidirectional, or even random access?

kennykerr: No reason, I've just not come across a motivating use case. I'll add an issue to address this for both begin/rbegin.

PR #516, opened 2019-07-16
Add C++/WinRT when_all async helper

kennykerr:
The `winrt::when_all` helper function creates an `IAsyncAction` object that will complete when all of the supplied awaitable objects have completed. The resulting async object can then either be waited upon using `co_await when_all(...)` or `when_all(...).get()`.

While this is easy to implement in C++/WinRT, it turns out not to be too obvious to many developers and is helpful for developers particularly coming from a C# or PPL background that are accustomed to using `Task.WhenAll` in C# or `concurrency::when_all` in C++.

### DIFF
--- a/src/tool/cppwinrt/strings/base_coroutine_foundation.h
+++ b/src/tool/cppwinrt/strings/base_coroutine_foundation.h
@@ -663,3 +663,12 @@ namespace std::experimental
         };
     };
 }
+
+namespace winrt
+{
+    template <typename... Args>
+    Windows::Foundation::IAsyncAction when_all(Args&&... args)
+    {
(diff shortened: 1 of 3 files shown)


bschoepke:
@kennykerr how would you do when_any? Also very useful for doing things like timeouts.

https://docs.microsoft.com/en-us/dotnet/api/system.threading.tasks.task.whenany?view=netframework-4.8

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/race

kennykerr:
@bschoepke I'll add a `when_any` next. 😉

PR #520, opened 2019-07-17
Add C++/WinRT when_any async helper

kennykerr:
Complements #516 by adding a `when_any` helper function that creates an `IAsyncAction` object that will complete when any of the supplied awaitable objects have completed. The resulting async object can then either be waited upon using `co_await when_any(...)` or `when_any(...).get()`.

This is a little harder to implement than `when_all`, but just as useful for things like redundant operations or cooperative timeouts. As I mentioned in #341, WinRT only allows waiting once. That means that any objects that may not yet have completed may not be waited on again in any way.

Here's an example:

```C++
using namespace std::literals;
using namespace winrt;

IAsyncAction wait_for(TimeSpan timeout)
{
    co_await timeout;
    printf("wait_for %lld\n", timeout.count());
}

int main()
{
    init_apartment();

    when_all(wait_for(3s), wait_for(2s), wait_for(1s)).get();
    puts("> when_all");

    puts("");

    when_any(wait_for(3s), wait_for(2s), wait_for(1s)).get();
    puts("> when_any");

    getchar();
}
```

And this program produces the following output:

```
wait_for 10000000
wait_for 20000000
wait_for 30000000
> when_all

wait_for 10000000
> when_any
wait_for 20000000
wait_for 30000000
```


src/tool/cppwinrt/strings/base_coroutine_foundation.h
```diff
@@ -666,9 +666,36 @@ namespace std::experimental
...
+        void* process = WINRT_GetCurrentProcess();
+
+        auto duplicate = [&](handle const& source) noexcept
+        {
+            void* target{};
+            WINRT_VERIFY(WINRT_DuplicateHandle(process, source.get(), process, &target, 0, false, 2 /* DUPLICATE_SAME_ACCESS */));
+            return handle{ target };
+        };
+
+        auto completed = [&](auto&& async)
```
kennykerr: This "completed" lambda isn't technically needed, but it is here to [work around a compiler bug](https://developercommunity.visualstudio.com/content/problem/648507/c-pack-expansion-crashes-compile-with-lambda-expre.html).

kennykerr:
Working on an improved version that returns the "winning" result.

PR #530, opened 2019-07-19
Add support for PE32+ (64-bit) winmd parsing, to re-enable managed assemblies

### LINKED ISSUE #527: Support referencing a c# managed winmd
This keeps coming up in 3rd party library scenarios. If you attempt to reference a managed winmd, cppwinrt fails with "PE section containing CLI header not found". Managed WinMDs include both the IL and the metadata.

Scottj1s:
Fixes #527

### DIFF
--- a/src/library/impl/meta_reader/database.h
+++ b/src/library/impl/meta_reader/database.h
@@ -41,7 +41,7 @@ namespace xlang::meta::reader
 
             auto dos = file.as<impl::image_dos_header>();
 
-            if (dos.e_magic != 0x5A4D) // IMAGE_DOS_SIGNATURE
+            if (dos.e_signature != 0x5A4D) // IMAGE_DOS_SIGNATURE
             {
                 return false;
             }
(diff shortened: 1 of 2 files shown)


src/library/impl/meta_reader/database.h
```diff
@@ -57,19 +57,33 @@ namespace xlang::meta::reader
             {
                 return false;
             }
-
-            auto com = pe.OptionalHeader.DataDirectory[14]; // IMAGE_DIRECTORY_ENTRY_COM_DESCRIPTOR
-            auto sections = &file.as<impl::image_section_header>(dos.e_lfanew + sizeof(impl::image_nt_headers32));
+            
+            impl::image_section_header const* sections{};
+            uint32_t com_virtual_address{};
+            if (pe.OptionalHeader.Magic == 0x10B) // PE32
```
DefaultRyan: Prefer to define these two values as constants. The metadata writer should share the value.

Scottj1s: Yes, should do that for the DOS magic number too

PR #536, opened 2019-07-23
Use CreateFile2 and CreateFileMapping

micahl:


Scottj1s:
> Thanks, @Scottj1s! I initially had that and then I ran across this [stackoverflow post](https://stackoverflow.com/questions/27220/how-to-convert-stdstring-to-lpcwstr-in-c-unicode) that suggested that approach wouldn't always work as expected if the input was anything other than ASCII. I can switch it back if you think it wouldn't be an issue.

oh, that's good to know.  ok, leave as is, but please add some error handling (see https://github.com/microsoft/xlang/blob/d2cff46d73a2b5f320f8a101e0c39413f069d849/src/platform/win32_string_convert.cpp).

src/library/impl/meta_reader/view.h
```diff
@@ -237,7 +242,7 @@ namespace xlang::meta::reader
                 return{};
             }
 
-            handle mapping{ CreateFileMappingA(file.value, nullptr, PAGE_READONLY, 0, 0, nullptr) };
+            handle mapping{ CreateFileMapping(file.value, nullptr, PAGE_READONLY, 0, 0, nullptr) };
```
kennykerr: Any reason for this change? This makes the code dependent on whether or not UNICODE is defined. Avoid using these function-like macros and explicitly pick the function you wish to call.

src/library/impl/meta_reader/view.h
```diff
@@ -222,7 +222,12 @@ namespace xlang::meta::reader
         static byte_view open_file(std::string_view const& path)
         {
 #if XLANG_PLATFORM_WINDOWS
-            file_handle file{ CreateFileA(c_str(path), GENERIC_READ, FILE_SHARE_READ, nullptr, OPEN_EXISTING, FILE_ATTRIBUTE_NORMAL, nullptr) };
+            auto cstr = c_str(path);
+            int len = MultiByteToWideChar(CP_ACP, 0, cstr, path.length() + 1, 0, 0);
```
kennykerr: CP_UTF8

kennykerr: Prefer to use readable variable names, e.g. length (not len), buffer (not buf), and so on.

src/library/impl/meta_reader/view.h
micahl: I was trying to use the meta data reader in a c++/winrt app and hitting the limitation that the A version isn't available.  Would you recommend doing something like this...
#ifndef UNICODE
    // use CreateFileMappingA
#else 
    // use CreateFileMappingW
#endif

Scottj1s: You always want CreateFileMappingW with this change - just make that explicit

---
In reply to: [306941300](https://github.com/microsoft/xlang/pull/536#discussion_r306941300) [](ancestors = 306941300)

kennykerr: Yep what Scott said.

Scottj1s:
thanks Micah

PR #546, opened 2019-08-05
Improve scalability of C++/WinRT's coroutine resumption

### LINKED ISSUE #544: IContextCallback blocks coroutine's thread of execution indefinitely
Standalone repro:

```C++
#include <winrt/Windows.Foundation.h>
#include <winrt/Windows.System.h>
#include <Windows.h>

using namespace std::literals;
using namespace winrt;
using namespace Windows::System;

IAsyncAction ForegroundAsync(DispatcherQueue dispatcher)
{
    co_await resume_foreground(dispatcher);
}

IAsyncAction BackgroundAsync(DispatcherQueue dispatcher)
{
    co_await resume_background();
    int counter{};

    while (true)
    {
        // co_await ForegroundAsync(dispatcher);
        Sleep(1000);
        printf("%d\n", ++counter);
    }
}

int main()
{
    auto controller = DispatcherQueueController::CreateOnDedicatedThread();
    auto dispatcher = controller.DispatcherQueue();
    auto timer = dispatcher.CreateTimer();
    timer.Interval(100ms);
    timer.Tick([](auto&& ...) { printf("."); });
    timer.Start();

    BackgroundAsync(dispatcher).get();
}
```

This program prints something like this:

```
..........1
..........2
..........3
..........4
..........5
..........6
..........7
..........8
......
```

A new line is printed roughly every second. Uncomment the `co_await ForegroundAsync(dispatcher)` and the ticker no longer prints as the dispatcher thread is blocked waiting on the background thread. The result is as follows:

```
..........1
2
3
4
5
6
7
8
```

The problem is that `IContextCallback::ContextCallback` blocks the caller indefinitely but typically while the waiting coroutine executes along to its next suspension point. `await_suspend` should never block.

kennykerr:
This update fixes #544. Although this is strictly a bug fix, the OS is nearing feature complete so I'll let @BenJKuhn decide whether we can take this update right now or whether it should wait a while. 

I have been meaning to improve C++/WinRT coroutine resumption for a long time, but never really had a sufficiently motivating case to do so. Well @oldnewthing recently hit on a compelling scenario where this becomes a problem [that I described here](https://github.com/microsoft/xlang/issues/544).

The problem is that `IContextCallback::ContextCallback` blocks the caller indefinitely but typically while the waiting coroutine executes along to its next suspension point. `IContextCallback` is used to ensure that resumption following a `co_await` on an WinRT async object will always bring the caller back to the same apartment where suspension originally occurred. This was originally deemed necessary to avoid confusion. However the use of `IContextCallback` in every case isn't really necessary and can lead to some undesirable consequences, chief of which is the [pathological case illustrated here](https://github.com/microsoft/xlang/issues/544). 

Beyond that example, using `IContextCallback` has a few drawbacks:

1. While it is a relatively efficient way to switch to a foreground (STA) thread, it is a rather expensive way of introducing a context switch to a background thread. Not only is the use of the `IContextCallback` more costly than interacting directly with the thread pool, COM's implementation relies on a custom thread pool which then oversubscribes the available processors. 

2. It always introduces a context switch even when one isn't really needed. If suspension occurred on a background (MTA) thread then resumption can simply execute directly without queuing a work item.

3. It always blocks the caller until the resuming coroutine suspends or returns. This can be disastrous for background threads as has already been illustrated. It can also lead to a stack overflow if a corouine has many `co_await` expressions (e.g. using `co_await` in a loop). This forces `await_ready` to always short-circuit in an attempt to avoid this recursion. Ideally, `await_ready` can always return `true` or `false` unconditionally. Checking the status also requires a `QueryInterface` call, which is rather expensive.

4. It requires COM to be initialized. C++/WinRT is designed to initialize the MTA on demand in most cases, but if the first thing the app does is wait on a coroutine then COM still needs to be initialized explicitly. 

This update fixes all of these problems. `IContextCallback` is only used when suspension occurs on a foreground thread. Resumption onto a background thread will resume directly if the async object completed on a background thread, otherwise it will use the thread pool to resume without blocking the caller.


src/tool/cppwinrt/strings/base_coroutine_foundation.h
```diff
@@ -86,28 +86,43 @@ namespace winrt::impl
...
-            return async.Status() == Windows::Foundation::AsyncStatus::Completed;
+            return false;
         }
 
         void await_suspend(std::experimental::coroutine_handle<> handle) const
         {
-            auto context = capture<IContextCallback>(WINRT_CoGetObjectContext);
-
-            async.Completed([handle, context = std::move(context)](auto const&, Windows::Foundation::AsyncStatus)
+            if (is_sta())
```
oldnewthing: This is an improvement as far as it goes. There is still a problem if a task starts one one STA and finishes on another STA. The second STA will be blocked until the suspension of the awaiting coroutine on the first STA. This is less likely to be a problem in practice because coroutines on STA are not supposed to block for long periods in the first place. (The original problem was when the initiating apartment was MTA and therefore thought it was at liberty to perform long blocking operations.)

src/tool/cppwinrt/test/await_adapter.cpp
```diff
@@ -0,0 +1,94 @@
...
+        // This excercises one MTA thread waiting on another and just completing
+        // directly without the overhead of a context switch.
+        co_await OtherBackgroundAsync();
+        REQUIRE(!impl::is_sta());
+
+        // Wait for a coroutine that completes on a foreground (STA) thread.
+        co_await ForegroundAsync(dispatcher);
+
+        // Resumption should automatically switch to a background (MTA) thread
+        // without blocking the Completed handler and thus blocking the foreground thread.
```
oldnewthing: Missing word?
```cpp
// without blocking the Completed handler and thus **not** blocking the foreground thread.
```

Perhaps reword

```cpp
// Resumption should automatically switch to a background (MTA) thread
// without blocking the Completed handler (which would in turn block the foreground thread).
```

src/tool/cppwinrt/test/await_adapter.cpp
```diff
@@ -0,0 +1,94 @@
...
+        REQUIRE(SetEvent(signal.get()));
+    }
+
+    IAsyncAction BackgroundAsync(DispatcherQueue dispatcher)
+    {
+        // Switch to a background (MTA) thread.
+        co_await resume_background();
+        REQUIRE(!impl::is_sta());
+
+        // This excercises one MTA thread waiting on another and just completing
```
oldnewthing: Typo: exercises

oldnewthing: This exercises but does not validate. How can we validate that there was no context switch? Should we match up thread IDs?

src/tool/cppwinrt/strings/base_coroutine_foundation.h
kennykerr: Yes, I considered this a tolerable blocking operation since either foreground thread would/should not have long-running operations that would cause the other to starve.

src/tool/cppwinrt/test/await_adapter.cpp
kennykerr: Yes, for the case where we're not queuing we could check the thread Id but that in itself doesn't guarantee a context switch didn't occur since the thread pool may and typically does reuse threads for short-lived work items. I'm not too concerned. The tests are primarily to ensure there's no deadlock and provide a way to manually verify the behavior when desired.

oldnewthing: Yeah, matching up thread IDs is not sufficient to prove that no context switch occurred. Was not aware what the "standard of testing" is.

src/tool/cppwinrt/test/await_adapter.cpp
kennykerr: Sounds better - thanks!

PR #559, opened 2019-08-27
Update to_hresult to handle any exception types.

### LINKED ISSUE #558: winrt::to_hresult doesn't handle thrown objects other than std::exception/winrt::hresult_error.
When winrt::terminate is used in code that throw something else than std::exception or winrt::hresult_error, the call to to_hresult throws itself.

jlaanstra:
Fixes #558.


jlaanstra:
Closing based on conversation.

PR #561, opened 2019-08-28
Ensure the platform winmds are always referenced.

jlaanstra:
The VS build targets only add the platform WinMD references if the following result is not empty:

```
@(ClCompile->WithMetadataValue('CompileAsWinRT', 'true'))@(Midl->WithMetadataValue('EnableWindowsRuntime', 'true'))'
```

For projects where the dll project is empty and all idl and code is in a static library, this evaluates to false resulting in missing references. 
Since we always want to reference the platform WinMDs, we set WinMDAssembly to true which causes the VS targets to do the right thing.


src/package/cppwinrt/nuget/Microsoft.Windows.CppWinRT.props
```diff
@@ -13,6 +13,8 @@ Copyright (C) Microsoft Corporation. All rights reserved.
         <CppWinRTPackage Condition="'$(CppWinRTPackage)' != 'true'">false</CppWinRTPackage>
         <XamlLanguage>CppWinRT</XamlLanguage>
         <IsNativeLanguage>true</IsNativeLanguage>
+        <!-- This causes VS to add the platform WinMD refetrences for the target SDK -->
```
Scottj1s: typo

PR #582, opened 2019-09-16
C++/WinRT support for non-WinRT error information

### LINKED ISSUE #580: C++/WinRT hresult_error struct loses error message
When `Windows::Data::Xml::Dom::XmlDocument::LoadXML()` encounters an XML parsing error, the `hresult_error` exception loses the message string containing the error details. `hresult_error::message()` returns `The text associated with this error code could not be found.`

Debugging, the issue appears to be because the object backing the error information only supports the `IErrorInfo` interface and not `IRestrictedErrorInfo`. One potential fix would be to have the appropriate `hresult_error` constructor call `GetErrorInfo()` to retrieve the `IErrorInfo` interface instead of calling `GetRestrictedErrorInfo()`. For error objects that support it, the `IRestrictedError` interface could then be returned from `QueryInterface()`.

This also could be a defect in the Windows Runtime. The underying implementation of `Windows::Data::Xml::Dom::XmlDocument::LoadXML()` appears to be MSXML6, and the Windows Runtime just passes through the MSXML6 error object instance rather than translating it to an error object that implements `IRestrictedErrorInfo`.

The MSXML6 HRESULT error codes do not appear to be translatable to message strings using standard techniques.

To reproduce, call `Windows::Data::Xml::Dom::XmlDocument::LoadXML()`  with a string containing a document type definition. This should throw an exception with `HRESULT` 0xC00CE584. The `IErrorInfo` description is `DTD is prohibited.` -- this is the message that should get passed through.

Develpoment environment is:
Windows 10 Enterprise 1903
Visual Studio 16.2.3
C++/WinRT 2.0.190730.2
Windows Implementation Library 1.0.190716.2

kennykerr:
Fixes #580 

C++/WinRT assumes error information is propagated using WinRT error origination APIs. This update ensures that APIs that only use the older `IErrorInfo` interface for error reporting still get their error messages reported by C++/WinRT.

### DIFF
--- a/src/tool/cppwinrt/strings/base_error.h
+++ b/src/tool/cppwinrt/strings/base_error.h
@@ -95,13 +95,10 @@ WINRT_EXPORT namespace winrt
 
         hresult_error(hresult const code, take_ownership_from_abi_t) noexcept : m_code(code)
         {
-            WINRT_GetRestrictedErrorInfo(m_info.put_void());
+            com_ptr<impl::IErrorInfo> info;
+            WINRT_GetErrorInfo(0, info.put_void());
 
-            if (m_info == nullptr)
-            {
-                originate(code, nullptr);
-            }
-            else
+            if (m_info = info.try_as<impl::IRestrictedErrorInfo>())
             {
                 WINRT_VERIFY_(impl::error_ok, m_info->GetReference(m_debug_reference.put()));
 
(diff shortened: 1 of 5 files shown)


src/tool/cppwinrt/strings/base_error.h
```diff
@@ -110,6 +107,24 @@ WINRT_EXPORT namespace winrt
...
+            else
+            {
+                impl::bstr_handle legacy;
+
+                if (info)
+                {
+                    info->GetDescription(legacy.put());
+                }
+
+                hstring message;
```
manodasanW: Consider if you get want to append GetGuid / GetSource to the message.  Not sure how often they are used.

kennykerr: Rarely in my experience, and then I'd have to worry about how to format it. 😨

Issue #551, opened 2019-08-12
Add support for win32 service hosted local servers, the ability to hook winrt::get_module_lock()


ChrisGuzak:
Win32 services that host object need to inform the service controller about their ref count via CoAddRefSharedService()/CoReleaseSharedService

winrt::get_module_lock() does not have a way to redirection the operator++ and – to other functions, instead it maintains a single module specific ref count that is only visible to the C++ winrt machinery. 

If an outofproc server implemented in a service needs to idle down – and almost all should – this count isn’t very useful because the server would need to poll it, and it is not possible to prevent activation failures when a new activation races the server shutting down.

kennykerr:
Looks like `CoAddRefSharedService`/`CoReleaseSharedService` are not public functions. It would be simpler if whatever OS functions this feature requires were actually public and thus callable from the C++/WinRT implementation. That way we can naturally support services without having to add a bunch of hooks that make the feature harder to support.

kennykerr:
Implemented. 👍

PR #583, opened 2019-09-16
Module locking improvements for C++/WinRT

### LINKED ISSUE #579: C++/WinRT delegates should affect the module lock count
Interface and coroutine implementations in C++/WinRT update the module lock count, but delegates do not. They should, otherwise there is no way to reliably support unloading with outstanding delegates.

kennykerr:
By default, C++/WinRT provides a global module lock via the `winrt::get_module_lock` function that keeps track of outstanding references so that  `DllCanUnloadNow` can report back to the COM runtime appropriately. In some cases, custom hosting environments need something they can hook to deal with lifetime a little differently. Also, apps don't really need to keep track of lifetime as they don't typically rely on something like `DllCanUnloadNow` to determine when to shut down. `TerminateProcess` will do just fine. 😉 

This update now supports these scenarios. There are three options:

1. By default, everything remains the same with a thread-safe global module lock.

2. If you define `WINRT_CUSTOM_MODULE_LOCK` then you must provide your own `winrt::get_module_lock` implementation. 

3. If you define `WINRT_NO_MODULE_LOCK` then all of the locking overhead is elided. The `winrt::get_module_lock` function still exists, but it no longer reports anything meaningful.

This update also fixes an issue where delegates weren't participating in the module lock.

Fixes #579 and #551.

### DIFF
--- a/src/tool/cppwinrt/strings/base_delegate.h
+++ b/src/tool/cppwinrt/strings/base_delegate.h
@@ -26,7 +26,7 @@ WINRT_EXPORT namespace winrt
 
 #elif defined (WINRT_CUSTOM_MODULE_LOCK)
 
-    // When WINRT_CUSTOM_MODULE_LOCK is defined, you must provide an implementaiton of winrt::get_module_lock()
+    // When WINRT_CUSTOM_MODULE_LOCK is defined, you must provide an implementation of winrt::get_module_lock()
     // that returns an object that implements operator++ and operator--.
 
 #else
--- a/src/tool/cppwinrt/strings/base_delegate.h
+++ b/src/tool/cppwinrt/strings/base_delegate.h
@@ -1,4 +1,36 @@
 
+namespace winrt::impl
+{
+    struct atomic_ref_count
+    {
+        uint32_t operator++() noexcept
+        {
+            return m_count.fetch_add(1, std::memory_order_relaxed) + 1;
+        }
+
+        uint32_t operator--() noexcept
+        {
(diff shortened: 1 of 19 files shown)


src/tool/cppwinrt/strings/base_delegate.h
```diff
@@ -1,10 +1,61 @@
...
+    // When WINRT_CUSTOM_MODULE_LOCK is defined, you must provide an implementation of winrt::get_module_lock()
+    // that returns an object that implements operator++ and operator--.
+
+#else
+
+    // This is the default implementation for use with DllCanUnloadNow.
+
+    inline std::atomic<uint32_t>& get_module_lock() noexcept
+    {
+        static std::atomic<uint32_t> s_lock;
```
DefaultRyan: While we're here, let's improve the performance here by not forcing a full sequential consistency barrier on every ++ (which I believe is the default behavior for `std::atomic`.

Instead, let's apply the lessons we've learned and use a more performant "atomic_ref_count", with `std::memory_ordering::relaxed` ordering on `++`, and `std::memory_ordering::release` on `--`, with an added `acquire` thread fence in `--` when the ref count hits zero. I can resend an email I posted about this some time back if you want.

kennykerr: Sure, that makes sense as we're already doing that for AddRef/Release.

kennykerr: Done, I've added a `atomic_ref_count` helper as you suggested. Thanks!

PR #585, opened 2019-09-18
Update the GetCppWinRTMdMergeInputs target to not depend on the Midl target.

jlaanstra:
With #491, the Midl target is invoked twice in multi-pass builds used in razzle. This PR removes the target dependency that caused this as it is not necessary. 

It also adds <target>DependsOn properties so the build dependencies are more clear.

### DIFF
--- a/src/package/cppwinrt/nuget/Microsoft.Windows.CppWinRT.targets
+++ b/src/package/cppwinrt/nuget/Microsoft.Windows.CppWinRT.targets
@@ -132,7 +132,7 @@ Copyright (C) Microsoft Corporation. All rights reserved.
          Make sure that we've ran the GetCppWinRTProjectWinMDReferences target
          and then remove them. -->
     <Target Name="CppWinRTRemoveStaticLibraries"
-            DependsOnTargets="GetCppWinRTProjectWinMDReferences;"
+            DependsOnTargets="GetCppWinRTProjectWinMDReferences"
             Returns="@(_ResolvedProjectReferencePaths)">
         <ItemGroup>
             <_ResolvedProjectReferencePaths Remove="@(_ResolvedProjectReferencePaths)" Condition="'%(_ResolvedProjectReferencePaths.ProjectType)' == 'StaticLibrary'" />
--- a/src/package/cppwinrt/nuget/Microsoft.Windows.CppWinRT.targets
+++ b/src/package/cppwinrt/nuget/Microsoft.Windows.CppWinRT.targets
@@ -37,15 +37,10 @@ Copyright (C) Microsoft Corporation. All rights reserved.
             $(ResolveAssemblyReferencesDependsOn);GetCppWinRTProjectWinMDReferences;CppWinRTRemoveStaticLibraries
         </ResolveAssemblyReferencesDependsOn>
         <ComputeCompileInputsTargets>
-            $(ComputeCompileInputsTargets);
-            CppWinRTHeapEnforcementOptOut;
+            $(ComputeCompileInputsTargets);CppWinRTHeapEnforcementOptOut;
         </ComputeCompileInputsTargets>
         <BeforeClCompileTargets>
-            $(BeforeClCompileTargets);
-            CppWinRTCalculateEnabledProjections;
-            CppWinRTMakePlatformProjection;
-            CppWinRTMakeReferenceProjection;
-            CppWinRTMakeComponentProjection
+            $(BeforeClCompileTargets);CppWinRTMakeProjections;
         </BeforeClCompileTargets>
         <MarkupCompilePass1DependsOn>
             $(MarkupCompilePass1DependsOn);CppWinRTAddXamlReferences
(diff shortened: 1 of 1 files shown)


Scottj1s:
I'm not in favor of the DependsOn properties.  If they're meant as extensibility points, then they should append to their existing values.  If not, then they just move the dependencies far away from their points of reference, making things less clear, and the targets file larger.  I'd rather we be intentional about adding extensibility points.

jlaanstra:
@Scottj1s They were helpful to understand the dependencies and you can already inject targets anywhere using BeforeTargets/AfterTargets, but I can certainly remove them again.

src/package/cppwinrt/nuget/Microsoft.Windows.CppWinRT.targets
```diff
@@ -93,7 +88,7 @@ Copyright (C) Microsoft Corporation. All rights reserved.
     </Target>
 
     <Target Name="CppWinRTCalculateEnabledProjections"
-            DependsOnTargets="GetCppWinRTProjectWinMDReferences;GetCppWinRTMdMergeInputs"
+            DependsOnTargets="GetCppWinRTProjectWinMDReferences;GetCppWinRTMdMergeInputs;$(CppWinRTCalculateEnabledProjectionsDependsOn)"
```
Scottj1s: left the DependsOn prop.  If not intentional, let's remove here and elsewhere.  If intentional (somebody needs a hook), let's add to all targets for consistency (we do that later, it's not related to this PR).  

Again, we shouldn't implement our own dependencies in turn on public DependsOn props, as users could easily break that if they forget the pattern of appending to existing DependsOn values.

src/package/cppwinrt/nuget/Microsoft.Windows.CppWinRT.targets
```diff
@@ -137,7 +132,7 @@ Copyright (C) Microsoft Corporation. All rights reserved.
          Make sure that we've ran the GetCppWinRTProjectWinMDReferences target
          and then remove them. -->
     <Target Name="CppWinRTRemoveStaticLibraries"
-            DependsOnTargets="GetCppWinRTProjectWinMDReferences"
+            DependsOnTargets="GetCppWinRTProjectWinMDReferences;"
```
Scottj1s: semicolon

src/package/cppwinrt/nuget/Microsoft.Windows.CppWinRT.targets
```diff
@@ -160,7 +155,7 @@ Copyright (C) Microsoft Corporation. All rights reserved.
 
     <!--Define platform WinMD references for modern IDL compilation-->
     <Target Name="GetCppWinRTPlatformWinMDReferences"
-            DependsOnTargets="ResolveAssemblyReferences"
+            DependsOnTargets="ResolveAssemblyReferences;$(GetCppWinRTPlatformWinMDReferencesDependsOn)"
```
Scottj1s: another orphan DependsOn prop

src/package/cppwinrt/nuget/Microsoft.Windows.CppWinRT.targets
```diff
@@ -315,7 +310,7 @@ $(XamlMetaDataProviderPch)
     <!--Insert Midl /references to Platform WinMDs, library reference WinMDs, and direct reference WinMDs-->
     <Target Name="CppWinRTSetMidlReferences"
             BeforeTargets="Midl"
-            DependsOnTargets="GetCppWinRTPlatformWinMDReferences;CppWinRTAddXamlMetaDataProviderIdl;GetCppWinRTProjectWinMDReferences;GetCppWinRTDirectWinMDReferences;$(CppWinRTSet_MidlReferencesDependsOn)">
+            DependsOnTargets="GetCppWinRTPlatformWinMDReferences;GetCppWinRTDirectWinMDReferences;GetCppWinRTProjectWinMDReferences;CppWinRTAddXamlMetaDataProviderIdl;$(CppWinRTSetMidlReferencesDependsOn)">
```
Scottj1s: I don't see any references to CppWinRTSet_MidlReferencesDependsOn, so probably safe to rename, but we should be very intentional about that.

src/package/cppwinrt/nuget/Microsoft.Windows.CppWinRT.targets
jlaanstra: I left this on intentionally, as it seems like a useful place for extension.

I don't think there any any dependencies implemented with our public DependsOn props anymore?

src/package/cppwinrt/nuget/Microsoft.Windows.CppWinRT.targets
jlaanstra: I would say some targets are us hooking into MSBuild to make functionality work, like fixing the references for static libs. I don't think those targets should have DependsOn.

src/package/cppwinrt/nuget/Microsoft.Windows.CppWinRT.targets
jlaanstra: Ah good catch. I can add the underscore back. Didn't realize it had an underscore there before. If there are no refs let's keep the rename.

PR #599, opened 2019-09-23
Deprecate from_abi in C++/WinRT

### LINKED ISSUE #586: Remove winrt::from_abi from C++/WinRT
This function tends to confuse developers and should have been deprecated and removed a long time ago. Simply use `winrt::get_self` instead.

Just need to make sure there are no dependent tools (e.g. Xaml compiler) in the OS.

kennykerr:
Fixes #586


kennykerr:
I don't expect any issues as we don't actually refer to from_abi anywhere.

PR #601, opened 2019-09-23
Add formal test case for WinRT generic identifier generation in C++/WinRT

kennykerr:
Previously, this was tested in a variety of places but it really helps to have this consolidated in order to more easily verify that changes to generics under the hood don't break these invariants. It can also help to bring up new languages.

### DIFF
--- a/src/tool/cppwinrt/test/generic_types.cpp
+++ b/src/tool/cppwinrt/test/generic_types.cpp
@@ -105,7 +105,9 @@ TEST_CASE("generic_types")
     REQUIRE_EQUAL_GUID("FA0B7D80-7EFA-52DF-9B69-0574CE57ADA4", EventHandler<A>);
     REQUIRE_EQUAL_GUID("EDB31843-B4CF-56EB-925A-D4D0CE97A08D", TypedEventHandler<A, B>);
 
+    //
     // Generated Windows.Foundation.Collections GUIDs
+    //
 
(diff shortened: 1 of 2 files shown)


src/tool/cppwinrt/test/generic_types.cpp
```diff
@@ -0,0 +1,180 @@
...
+                static_cast<uint8_t>(to_uint(value[26]) * 16 + to_uint(value[27])),
+                static_cast<uint8_t>(to_uint(value[28]) * 16 + to_uint(value[29])),
+                static_cast<uint8_t>(to_uint(value[30]) * 16 + to_uint(value[31])),
+                static_cast<uint8_t>(to_uint(value[32]) * 16 + to_uint(value[33])),
+                static_cast<uint8_t>(to_uint(value[34]) * 16 + to_uint(value[35])),
+            }
+        };
+    }
+
+    constexpr bool equal(guid const& left, guid const& right) noexcept
```
Scottj1s: why not use the operator in base_types.h?

kennykerr: Good point. I'll add those as well.

kennykerr: Added tests for all primitives, as well as enums, structs, IInspectable, classes, and delegates.

PR #603, opened 2019-09-24
C++/WinRT support for implementing COM interfaces that derive from ::IInspectable

kennykerr:
Ideally this would not be required, but regrettably there exists some COM interop interfaces that just happen to derive from `::IInspectable`. This update makes it easier to implement such interfaces in C++/WinRT.

This now "just works" provided you include `<inspectable.h>` before including any C++/WinRT headers. This is distinct from general COM interop support that merely requires that you include `<unknwn.h>` because changing that now would break the world.

```C++
struct __declspec(uuid("ed0dd761-c31e-4803-8cf9-22a2cb20ec47")) IBadInterop : ::IInspectable
{
    virtual int32_t __stdcall JustSayNo() noexcept = 0;
};

struct Sample : implements<Sample, Windows::Foundation::IActivationFactory, IBadInterop>
{
    Windows::Foundation::IInspectable ActivateInstance()
    {
        throw hresult_not_implemented();
    }

    hstring GetRuntimeClassName()
    {
        return L"Sample";
    }

    Windows::Foundation::TrustLevel GetTrustLevel()
    {
        return Windows::Foundation::TrustLevel::PartialTrust;
    }

    int32_t __stdcall JustSayNo() noexcept final
    {
        return 123;
    }
};
```

### DIFF
--- a/src/tool/cppwinrt/test/inspectable_interop.cpp
+++ b/src/tool/cppwinrt/test/inspectable_interop.cpp
@@ -1,0 +1,64 @@
+#include <inspectable.h>
+#include "winrt/Windows.Foundation.h"
+#include "catch.hpp"
+
+using namespace winrt;
+
+namespace
+{
+    struct __declspec(uuid("ed0dd761-c31e-4803-8cf9-22a2cb20ec47")) IBadInterop : ::IInspectable
+    {
+        virtual int32_t __stdcall JustSayNo() noexcept = 0;
+    };
+
+    struct Sample : implements<Sample, Windows::Foundation::IActivationFactory, IBadInterop>
+    {
+        Windows::Foundation::IInspectable ActivateInstance()
+        {
+            throw hresult_not_implemented();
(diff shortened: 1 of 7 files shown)


src/tool/cppwinrt/test/inspectable_interop.cpp
```diff
@@ -0,0 +1,64 @@
+#include <inspectable.h>
+#include "winrt/Windows.Foundation.h"
+#include "catch.hpp"
+
+using namespace winrt;
+
+namespace
+{
+    struct __declspec(uuid("ed0dd761-c31e-4803-8cf9-22a2cb20ec47")) IBadInterop : ::IInspectable
```
oldnewthing: In practice, this IBadInterop definition would come from some ABI header file, right?

kennykerr: Right.

PR #604, opened 2019-09-25
Update C++/WinRT to use variable templates internally

### LINKED ISSUE #571: C++/WinRT should use variable templates rather than type traits where possible
Variable templates are both simpler to read/write and less costly for the compiler to parse and store (and should thus reduce PCH size further). C++/WinRT relies heavily on type traits and predates variable templates. The only type traits that can be switched to variable templates are the GUIDs and names of WinRT types.

The branch `kennykerr-variable` has the changes to move GUIDs over to variable templates, but is hitting a C++ compiler bug that I'm working to resolve.

kennykerr:
C++/WinRT relies heavily on type traits and predates [variable templates](https://en.cppreference.com/w/cpp/language/variable_template). Variable templates are both simpler to read/write and less costly for the compiler to parse and store and should reduce PCH size further. This update doesn’t change or add any public features but overhauls the use of traits substantially to rely on variable templates instead. This work uncovered a bug/inefficiency in the C++ compiler when dealing with certain types of variable templates. As such, this change increases the size of the test PCH by about 2%. This is acceptable at the moment, given the stream of optimizations that I’ve introduced over the last few weeks to reduce PCH size substantially. However, preliminary tests on a future C++ compiler indicate a roughly 10% reduction in PCH size when combined with these changes to C++/WinRT. 

Since it is virtually impossible to reason about this kind of change, this update also includes substantially more test coverage for the constexpr variable templates used by generics types. This ensures that all of the changes in this PR are verified using static assertions rather exhaustively (generic_types.cpp takes a while to compile).

Fixes #571


src/tool/cppwinrt/cppwinrt/code_writers.h
```diff
@@ -373,20 +358,14 @@ namespace xlang
 
         if (empty(generics))
         {
-            auto format = R"(    template <> struct name<%>
-    {
-        static constexpr auto & value{ L"%.%" };
-    };
+            auto format = R"(    template <> constexpr auto& name_v<%>{ L"%.%" };
```
DefaultRyan: Should these constexpr variable templates be declared `inline`?

kennykerr: No, constexpr implies inline.

DefaultRyan: That only applies to static member variables, not namespace-scope variables. Namespace-scope variables like `name_v` are going to get duplicated in every translation unit that includes this code, unless they are given the `inline` treatment.

This was the original reason for adopting `name_of<T>()` and `guid_of<T>()` as functions rather than constexpr template variables - our version of MSVC at the time didn't support inline variables, and the bloat was causing our customers pain.

kennykerr: Hmm, I'll double check what's going on here.

kennykerr: Indeed: https://github.com/microsoft/xlang/pull/605

PR #608, opened 2019-09-26
Clang and GCC conformance improvements for C++/WinRT

kennykerr:


src/tool/cppwinrt/strings/base_identity.h
```diff
@@ -454,7 +454,7 @@ namespace winrt::impl
 
     template <typename T>
 #ifdef __clang__
-    static const auto value
```
DefaultRyan: Should this be inline? How about constexpr?

kennykerr: It can't be constexpr because Clang refuses to treat __uuidof as a constant expression, but I'll make it inline.

src/tool/cppwinrt/strings/base_meta.h
```diff
@@ -121,11 +121,15 @@ namespace winrt::impl
     };
 
     template <typename T>
+#ifdef __IUnknown_INTERFACE_DEFINED__
 #ifdef __clang__
     const guid guid_v{ __uuidof(T) };
 #else
     constexpr guid guid_v{ __uuidof(T) };
 #endif
+#else
```
DefaultRyan: I think these should all be inline as well.

kennykerr: Right.

src/tool/cppwinrt/strings/base_identity.h
kennykerr: Fixed.

src/tool/cppwinrt/strings/base_meta.h
kennykerr: Fixed.

PR #628, opened 2020-01-13
Updating parsing to handle xml manifest files

EricJohnson327:
Updating parsing to handle xml manifest files.


src/winrt/activation/catalog.cpp
```diff
@@ -4,6 +4,10 @@
 #include <roapi.h>
 #include <winstring.h>
 #include <rometadataresolution.h>
+#include <combaseapi.h>
+#include <wrl.h>
```
kennykerr: Ideally you should use something like C++/WinRT's `com_ptr` rather than WRL.

src/winrt/activation/catalog.cpp
```diff
@@ -4,6 +4,10 @@
...
+#include <msxml6.h>
```
kennykerr: Ideally you should use XmlLite rather than msxml.

src/winrt/activation/catalog.cpp
EricJohnson327: This is shipping back to RS2, will that still work?  Scott made it sound like we couldn't use WinRT, but I may be confusing different concepts.

kennykerr: If you're only depending on the Windows SDK here that may not work.

src/winrt/activation/catalog.cpp
EricJohnson327: Isn't XmlLite asynchronous?  Async will cause issues since we need to have this completed early.

kennykerr: No, it is not. https://docs.microsoft.com/en-us/archive/msdn-magazine/2007/april/xmllite-a-small-and-fast-xml-parser-for-native-c

src/winrt/activation/catalog.cpp
```diff
@@ -62,26 +66,66 @@ static unordered_map < wstring, shared_ptr<component> > g_types;
...
+            {
+                this_component->threading_model = ABI::Windows::Foundation::ThreadingType::ThreadingType_STA;
+            }
+            else if (wcsicmp(L"free", value) == 0)
+            {
+                this_component->threading_model = ABI::Windows::Foundation::ThreadingType::ThreadingType_MTA;
+            }
+            else
+            {
+                this_component->threading_model = ABI::Windows::Foundation::ThreadingType::ThreadingType_BOTH;
```
kythant: Do we need to check to see if it specifically says "both"?  Not sure if the language in the parser will restrict any other keywords and if it doesn't we could have anything in there and it will default to both

src/winrt/activation/catalog.cpp
```diff
@@ -62,26 +66,66 @@ static unordered_map < wstring, shared_ptr<component> > g_types;
...
-    this_component->metadata_path = converter.from_bytes(u8_metadata_path);
-
-    // each line after is an activatable class. No apartment-specific registration. Implicitly "both".
-    string u8_class_name;
-    while (getline(manifest_file, u8_class_name))
+    VARIANT_BOOL vbResult = VARIANT_FALSE;
+
+    ComPtr<IXMLDOMDocument2> xmlDoc;
+    RETURN_IF_FAILED(CoCreateInstance(CLSID_DOMDocument, NULL, CLSCTX_INPROC_SERVER, IID_PPV_ARGS(xmlDoc.GetAddressOf())));
+    HRESULT hr = xmlDoc->load(_variant_t(manifest_path), &vbResult);
```
kythant: RETURN_IF_FAILED here

src/winrt/activation/catalog.cpp
BenJKuhn: sorry, dinosaur here. I steered them towards MSXML, but that would be a much better alternative.

src/winrt/example/TestFiles/demo2.manifest
```diff
@@ -0,0 +1,11 @@
+﻿<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
+<assembly xmlns="urn:schemas-microsoft-com:asm.v1" manifestVersion="1.0">
+  <file name="test_component.dll">
+    <comClass
+        clsid="test_component.Class"
+        threadingModel="Both">
+
+      <progid>"test_component.Class"</progid>
```
jonwis: Quotes shouldn't be required within the `<progid/>` element.

EricJohnson327: Sorry, wasn't actually using the progid in my test.  Holdover from copy/paste.  Will remove from test manifest.

src/winrt/activation/catalog.cpp
EricJohnson327: Switched over to XmlLite

src/winrt/activation/catalog.cpp
BenJKuhn: The goal is to mirror the 19H1 support. Review the corresponding OS code and follow suit.

src/winrt/activation/catalog.cpp
```diff
@@ -70,26 +76,57 @@ static unordered_map < wstring, shared_ptr<component> > g_types;
...
+    RETURN_IF_FAILED(xmlReader->SetInput(fileStream.Get()));
+    const WCHAR* fileName;
+    while (S_OK == xmlReader->Read(&nodeType))
     {
-        std::wstring wide_class_name = converter.from_bytes(u8_class_name);
-        g_types[wide_class_name] = this_component;
+        if (nodeType == XmlNodeType_Element)
+        {
+            RETURN_IF_FAILED((xmlReader->GetLocalName(&localName, nullptr)));
+            if (localName && _wcsicmp(localName, L"file") == 0)
```
kennykerr: The pointer check isn't needed as GetLocalName will return a valid string in all cases (even if it's an empty string).

src/winrt/test/CMakeLists.txt
```diff
@@ -24,9 +24,19 @@ add_custom_command(TARGET regfree_winrt_test POST_BUILD
     ${CMAKE_CURRENT_BINARY_DIR}
 )
 
+add_custom_command(TARGET regfree_winrt_test POST_BUILD
```
kythant: Should move this to example directory's cmake file

src/winrt/activation/catalog.cpp
EricJohnson327: Spec made it sound like both was default if none supplied, didn't realize ThreadingModel attribute was required.  Also, was using "apartment" and "free" which are COM values.  Updated to "sta" and "mta".

src/winrt/activation/catalog.cpp
```diff
@@ -70,26 +76,61 @@ static unordered_map < wstring, shared_ptr<component> > g_types;
...
+                RETURN_IF_FAILED(xmlReader->GetValue(&fileName, nullptr));
+                this_component->module_path = fileName;
+            }
+            else if (_wcsicmp(localName, L"activatableClass") == 0)
+            {
+                const WCHAR* threadingModel;
+                RETURN_IF_FAILED(xmlReader->MoveToAttributeByName(L"threadingModel", nullptr));
+                RETURN_IF_FAILED(xmlReader->GetValue(&threadingModel, nullptr));
+
+                if (wcsicmp(L"sta", threadingModel) == 0)
```
BenJKuhn: This will use the CRT application locale, which can be changed, and could lead to difficult-to-diagnose failures. Consider usign wcsicmp_l passing in the default C locale explicitly.

PR #631, opened 2020-01-25
Ejohn/undocked winrt activation/metadata

EricJohnson327:
Adding support for RoGetMetaDataFile and RoResolveNamespace detours.


src/UndockedRegFreeWinRT/UndockedRegFreeWinRT/catalog.cpp
```diff
@@ -42,15 +41,15 @@ struct component
...
             if (handle == nullptr)
             {
-                return HRESULT_FROM_WIN32(GetLastError());;
+                return HRESULT_FROM_WIN32(GetLastError());
             }
             this->get_activation_factory = (activation_factory_type)GetProcAddress(handle, "DllGetActivationFactory");
             if (this->get_activation_factory == nullptr)
             {
-                return HRESULT_FROM_WIN32(GetLastError());;
+                return HRESULT_FROM_WIN32(GetLastError());
```
kennykerr: You're leaking the library handle.

src/UndockedRegFreeWinRT/UndockedRegFreeWinRT/catalog.cpp
```diff
@@ -172,32 +173,44 @@ HRESULT WinRTGetMetadataFile(
...
+        metaDataImport != nullptr ||
+        typeDefToken != nullptr)
+    {
+        return E_NOTIMPL;
+    }
 
-    // algorithm: search our metadata, find best match if any.
-    // search system metedata. 
-    // Take longest string match excluding extension (which should be winmd, but not clear it's required to be).
+    HSTRING substring;
```
kennykerr: You're leaking the HSTRING.

src/UndockedRegFreeWinRT/UndockedRegFreeWinRT/dllmain.cpp
```diff
@@ -29,12 +68,12 @@ VOID CALLBACK EnsureMTAInitializedCallBack
...
-thread that is not been CoInitialize. COM treats this thread as a implicit MTA and
-when we call CoGetObjectContext on it we implicitily initialized the MTA.
+/* 
+In the context callback call to the MTA apartment, there is a bug that prevents COM 
+from automatically initializing MTA remoting. It only allows NTA to be intialized 
+outside of the NTA and blocks all others. The workaround for this is to spin up another 
+thread that is not been CoInitialize. COM treats this thread as a implicit MTA and 
+when we call CoGetObjectContext on it we implicitily initialized the MTA. 
 */
 HRESULT EnsureMTAInitialized()
```
kennykerr: Wouldn't it be simpler to use CoIncrementMTAUsage?

EricJohnson327: No idea.  @kythant any ideas?

kennykerr: CoIncrementMTAUsage is what C++/WinRT uses to ensure the MTA is present. This function was introduced in Windows 8, so no good if you're targeting Windows 7 but it looks like you're calling all kinds of other Windows 8 functions as well.

kythant: COM bug in CoIncrementMTAUsage is preventing it from initializing MTA remoting and the callback below was failing with RPC_E_DISCONNECTED error. Thus the reason for this implicit MTA hack.

src/UndockedRegFreeWinRT/UndockedRegFreeWinRT/catalog.cpp
```diff
@@ -14,43 +14,50 @@
...
 using namespace std;
 using namespace Microsoft::WRL;
+using namespace Microsoft::WRL::Wrappers;
 
 // TODO: Components won't respect COM lifetime. workaround to get them in the COM list?
 
 extern "C"
 {
-    typedef HRESULT(__stdcall* activation_factory_type)(HSTRING, IActivationFactory**);
+    typedef HRESULT (__stdcall * activation_factory_type)(HSTRING, IActivationFactory**) ;
```
kythant: Weird spacing.

src/UndockedRegFreeWinRT/UndockedRegFreeWinRT/catalog.cpp
```diff
@@ -161,7 +170,7 @@ HRESULT WinRTGetActivationFactory(
         return component_iter->second->GetActivationFactory(activatableClassId, iid, factory);
     }
     return REGDB_E_CLASSNOTREG;
-
+
```
kythant: Unnecessary spacing?

PR #633, opened 2020-01-30
Disable detours for 19h1 and above. 

kythant:
Check via Windows.Foundation.UniversalApiContract whether we should be disabling the detours.
If Windows.Foundation.UniversalApiContract version 8 is present that means we're on 19h1 and above. 
Also added special function so that we can enable it for the tests. 

Cross MTA activation test is also fixed by making our detour library aggresively fallback to the true APIs which is probably what we should be doing anyway to not break existing scenarios.


kythant:
https://www.reddit.com/r/Eyebleach/

because the check is ewwwww

kythant:
> https://www.reddit.com/r/Eyebleach/
> 
> because the check is ewwwww

Was originally going to use build number but the APIs to access it was deprecated or can't be used here

src/UndockedRegFreeWinRT/UndockedRegFreeWinRT/dllmain.cpp
```diff
@@ -285,8 +287,15 @@ HRESULT ExtRoLoadCatalog()
     return WinRTLoadComponent(manifestPath.c_str());
 }
 
+
 BOOL WINAPI DllMain(HINSTANCE /*hmodule*/, DWORD reason, LPVOID /*lpvReserved*/)
 {
+    BOOL present = false;
+    // if version 8 is present, that means we're on 19h1 build 18362 and above and we want to disable detour.
+    if (RoIsApiContractMajorVersionPresent(L"Windows.Foundation.UniversalApiContract", 8, &present) == S_OK && present)
```
kennykerr: When was this function introduced?

src/UndockedRegFreeWinRT/UndockedRegFreeWinRT/dllmain.cpp
```diff
@@ -191,10 +195,36 @@ HRESULT WINAPI RoActivateInstanceDetour(HSTRING activatableClassId, IInspectable
     return S_OK;
 }
 
+VERSIONHELPERAPI IsWindowsVersionOrGreaterEx(WORD wMajorVersion, WORD wMinorVersion, WORD wServicePackMajor, WORD wBuildNumber)
+{
+    OSVERSIONINFOEXW osvi = { sizeof(osvi), 0, 0, 0, 0, {0}, 0, 0 };
```
jaykrell: This can just be `OSVERSIONINFOEXW osvi = { sizeof(osvi) };`
which is nicer imho. Subjective. Gcc might warn. Visual C++ will not.
If there is a union in there, it might not zero it all, but I don't think you need it to anyway.

src/UndockedRegFreeWinRT/UndockedRegFreeWinRT/extwinrt.h
```diff
@@ -0,0 +1,4 @@
+__pragma(once)
```
jaykrell: That's new style to me. Why not `#pragma once`?

src/UndockedRegFreeWinRT/Nuget/readme.txt
```diff
@@ -1,10 +1,37 @@
...
 The Microsoft.Windows.UndockedRegFreeWinrt NuGet package
-enabling you consum Windows Runtime classes registry free!
+enabling you consume Windows Runtime classes registry free!
 ========================================================================
 
-Insert details here
-  
+Installing this package will enable non-packaged desktop applications 
+to leverage user-defined Windows Runtime types via the use of the 
+funsion manifest.
```
jaykrell: `funsion`? This is a typo, of a code name..

jaykrell:
I think *slightly* more explanation is needed.
My first guess, was the newer OS disallows Detours.
But my second guess, is that newer OS implements this for you, which would be quite nice.

BenJKuhn:
> I think _slightly_ more explanation is needed.
> My first guess, was the newer OS disallows Detours.
> But my second guess, is that newer OS implements this for you, which would be quite nice.

Option two is correct. This is essentially patching in a feature that was created in 19H1, but would be useful to have on older versions of Windows as well. This is a good write-up about the original feature: 
https://blogs.windows.com/windowsdeveloper/2019/04/30/enhancing-non-packaged-desktop-apps-using-windows-runtime-components/

This is a shim that provides approximately the same behavior on versions of Windows prior to the introduction of OS support.

PR #635, opened 2020-03-02
prep abi tool for nuget release

Scottj1s:


Scottj1s:
@dunhor  - just the help strings in main.cpp

src/tool/abi/main.cpp
```diff
@@ -21,9 +21,57 @@ auto output_directory(reader const& args)
...
+
+static constexpr option options[]
+{
+    { "input", 0, option::no_max, "<spec>", "Windows metadata to include in generation" },
+    { "reference", 0, option::no_max, "<spec>", "Windows metadata to reference from generation" },
+    { "output", 0, 1, "<path>", "Location of generated headers" },
+    { "include", 0, option::no_max, "<prefix>", "One or more prefixes to include in input" },
+    { "exclude", 0, option::no_max, "<prefix>", "One or more prefixes to exclude from input" },
+    { "verbose", 0, 0, {}, "Show detailed progress information" },
+    { "ns-prefix", 0, 1, "<always|optional|never>", "Set namespace prefixing preference (default: never)" },
```
dunhor: >Set namespace prefixing preference (default: never) [](start = 53, length = 51)

Should probably call out that the prefix is the `ABI::` namespace prefix. E.g. maybe
> Controls whether or not type names are prepended with the 'ABI' namespace

For comparison, here's all that midl says: "`Prepend the 'ABI' namespace to all types`"

src/tool/abi/main.cpp
```diff
@@ -21,9 +21,57 @@ auto output_directory(reader const& args)
...
+    { "enum-class", 0, 0, {}, "Use 'MIDL_ENUM', rather than 'enum'" },
+    { "lowercase-include-guard", 0, 0, {}, "Generate lowercase include guards" },
```
dunhor: Perhaps specify that this is for compat with SDK headers. E.g. maybe
> Generate lowercase include guards for compatibility with Windows SDK headers
For reference, it's primarily a work around since published headers are always converted to lower case

PR #648, opened 2020-05-20
Use ComPtr to fix activation factory leaks.

### LINKED ISSUE #647: Undocked RegFree WinRT leaks activation factories
https://github.com/microsoft/xlang/blob/a7bac3174a96b020c087df857301437274de6c47/src/UndockedRegFreeWinRT/UndockedRegFreeWinRT/dllmain.cpp#L202

jlaanstra:
Fixes #647

### DIFF
--- a/src/UndockedRegFreeWinRT/UndockedRegFreeWinRT/dllmain.cpp
+++ b/src/UndockedRegFreeWinRT/UndockedRegFreeWinRT/dllmain.cpp
@@ -199,7 +199,7 @@ HRESULT WINAPI RoActivateInstanceDetour(HSTRING activatableClassId, IInspectable
     // Activate in current apartment
     if (location == ActivationLocation::CurrentApartment)
     {
-        IActivationFactory* pFactory;
+        Microsoft::WRL::ComPtr<IActivationFactory> pFactory;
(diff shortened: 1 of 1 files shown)


src/UndockedRegFreeWinRT/UndockedRegFreeWinRT/dllmain.cpp
```diff
@@ -199,7 +199,7 @@ HRESULT WINAPI RoActivateInstanceDetour(HSTRING activatableClassId, IInspectable
     // Activate in current apartment
     if (location == ActivationLocation::CurrentApartment)
     {
-        IActivationFactory* pFactory;
+        Microsoft::WRL::ComPtr<IActivationFactory> pFactory;
```
ChrisGuzak, line 202: is there a reason you are not using `winrt::com_ptr`?
consider removing the `p` prefix on the members that indicate raw pointer.

BenJKuhn: This code doesn't have a WIL dependency (it has a local definition for RETURN_IF_FAILED). It probably should be updated to use WIL, but that's a larger change not related to the bug fix at hand. I've filed a new issue: https://github.com/microsoft/xlang/issues/650

PR #653, opened 2020-06-08
Use WIL NugetPackage for RegFree WinRT

kythant:
Add WIL Nuget Package to dependencies and remove it's owned RETURN_IF_FAILED macro.

### DIFF
--- a/src/UndockedRegFreeWinRT/UndockedRegFreeWinRT/catalog.cpp
+++ b/src/UndockedRegFreeWinRT/UndockedRegFreeWinRT/catalog.cpp
@@ -15,7 +15,7 @@
 #include <codecvt>
 #include <locale>
 #include "catalog.h"
-
+ 
 using namespace std;
 using namespace Microsoft::WRL;
 using namespace Microsoft::WRL::Wrappers;
--- a/src/UndockedRegFreeWinRT/UndockedRegFreeWinRT/UndockedRegFreeWinRT.vcxproj
+++ b/src/UndockedRegFreeWinRT/UndockedRegFreeWinRT/UndockedRegFreeWinRT.vcxproj
@@ -176,6 +176,7 @@
   </ItemDefinitionGroup>
   <ItemGroup>
     <None Include="cpp.hint" />
+    <None Include="packages.config" />
     <None Include="winrtact.def" />
   </ItemGroup>
   <ItemGroup>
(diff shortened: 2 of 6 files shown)


src/UndockedRegFreeWinRT/UndockedRegFreeWinRT/catalog.cpp
```diff
@@ -193,20 +193,19 @@ HRESULT WinRTGetMetadataFile(
...
+        if (SUCCEEDED(WindowsCompareStringOrdinal(substring.get(), HStringReference(L"Windows").Get(), &result)) && (result == 0))
         {
             // Let original RoGetMetadataFile handle Windows namespace
             return REGDB_E_CLASSNOTREG;
         }
     }
 
     DWORD metaDataFilePathsCount = 0;
-    HSTRING* metaDataFilePaths;
+    wil::unique_cotaskmem_array_ptr<wil::unique_hstring> metaDataFilePaths;
```
DefaultRyan, line 208: This solves two leaks (the array being leaked, as well as all of the elements), but now it's going to overrelease the string that's returned.

```
*metaDataFilePath = metaDataFilePaths[bestMatch];
metaDataFilePaths[bestMatch] = nullptr;
```

DefaultRyan, line 208: Also, using this removes the need for `metaDataFilePathsCount`, and adds the ability to use range-based for, or iterators.

src/UndockedRegFreeWinRT/UndockedRegFreeWinRT/catalog.cpp
```diff
@@ -193,20 +193,19 @@ HRESULT WinRTGetMetadataFile(
     {
         return E_NOTIMPL;
     }
-
-    Microsoft::WRL::Wrappers::HString substring;
-    if (SUCCEEDED(WindowsSubstringWithSpecifiedLength(name, 0, 7, substring.GetAddressOf())))
+    wil::unique_hstring substring;
+    if (SUCCEEDED(WindowsSubstringWithSpecifiedLength(name, 0, 7, &substring)))
```
DefaultRyan, line 197: Probably allocates a new string. Since the length is known, why not just copy it into a local wchar_t[8], and then you can call `CompareStringOrdinal` on the raw strings instead of `WindowsCompareStringOrdinal`.

src/UndockedRegFreeWinRT/UndockedRegFreeWinRT/catalog.cpp
```diff
@@ -194,27 +194,22 @@ HRESULT WinRTGetMetadataFile(
         return E_NOTIMPL;
     }
 
-    Microsoft::WRL::Wrappers::HString substring;
-    if (SUCCEEDED(WindowsSubstringWithSpecifiedLength(name, 0, 7, substring.GetAddressOf())))
+    wchar_t szFolderPath[8];
+    StringCchPrintf(szFolderPath, _countof(szFolderPath), WindowsGetStringRawBuffer(name, nullptr));
+    if (CompareStringOrdinal(szFolderPath, 8, L"Windows", 8, false) == CSTR_EQUAL)
```
kythant, line 199: I have a feeling we should be comparing "Windows." not "Windows". Or else we might be avoiding namespaces that has "WindowsSomething" and that's not valid.

DefaultRyan, line 199: Nit: If the input strings are null-terminated (which I thought they are), it's typically more appropriate to supply -1 as both lengths.

PR #659, opened 2020-06-22
Update AbiWinRT semantics to match CppWinRT

### LINKED ISSUE #654: Make ABI header consumption easy
With Project Reunion and other components being moved out of the OS, there needs to be a method for them to consume ABI headers for customers that have already done so.

It could be the responsibility of every component author to ship headers, but in the case of WinUI, we have dependencies on types defined in other headers, and so there isn't always a clear line on which headers we need to include in our nuget for them to build. As more components are lifted, I'd imagine that this exercise would become more confusing.

Rather than putting the onerous on each component author to generate the headers and include them, the `Microsoft.Windows.AbiWinRT` nuget package should behave the same way that `Microsoft.Windows.CppWinRT` does, and sniff out the references and generate the appropriate headers.

stevenbrix:
This change makes the `Microsoft.Windows.AbiWinRT` package behave similarly to `Microsoft.Windows.CppWinRT`. When a projects adds a nuget reference to `Microsoft.Windows.AbiWinRT`, the references will be sniffed out and the appropriate header files will be generated. In order to support older projects that haven't switch to modern idl, I added the ability to generate the .idl files from the referened winmd by invoking `winmdidl` before we run midl. This is only needed because of the bug described here https://github.com/microsoft/xlang/issues/654#issuecomment-647632160

Note that I removed anything from cpp/winrt that was related to `XamlMetadataProvider`, or had anything to do with `Page` or `ApplicationDefinition`, like the heap enforcement target.

## Usage

|  Property |  Description |
|-----------|---------------|
| `AbiWinRTAddAbiNamespacePrefix` | Adds the `ABI::` prefix to all generated headers. Defaults to `true`. |
| `AbiWinRTEnsureSDKHeaderCompat` | Passes the `-lowercase-include-guard` switch to `abi.exe` to enable compat with SDK headers. Defaults to `false`. |

Fixes #654


src/package/abi/Microsoft.Windows.AbiWinRT.targets
```diff
@@ -6,8 +6,500 @@ Copyright (C) Microsoft Corporation. All rights reserved.
...
+        <AbiWinRTUseHardlinksIfPossible Condition="'$(AbiWinRTUseHardlinksIfPossible)' == ''">false</AbiWinRTUseHardlinksIfPossible>
+        <AbiWinRTWriteOnlyWhenDifferent Condition="('$(AbiWinRTWriteOnlyWhenDifferent)' == '') And (('$(MSBuildToolsVersion)' == 'Current') Or ('$(MSBuildToolsVersion)' &gt;= '15'))">true</AbiWinRTWriteOnlyWhenDifferent>
+        <AbiWinRTWriteOnlyWhenDifferent Condition="'$(AbiWinRTWriteOnlyWhenDifferent)' != 'true'">false</AbiWinRTWriteOnlyWhenDifferent>
+        <AbiWinRTPath Condition="'$(AbiWinRTPath)'==''">$(MSBuildThisFileDirectory)..\..\bin\</AbiWinRTPath>
+
+        <!-- Default modern idl semantics to false considering that most users probably have not made the switch to the modern idl syntax -->
+        <AbiWinRTModernIDL Condition="'$(AbiWinRTModernIDL)'==''">false</AbiWinRTModernIDL>
+
+        <!-- Only generate the platform and component projection if using modern idl. Otherwise, allow the legacy
+             toolchain to kick in, as well as use the existing references from the Windows SDK -->
```
Scottj1s, line 30: update comments copied from c++/winrt

src/package/abi/Microsoft.Windows.AbiWinRT.targets
```diff
@@ -6,8 +6,500 @@ Copyright (C) Microsoft Corporation. All rights reserved.
...
+        <AbiWinRTAddAbiNamespacePrefix Condition="'$(AbiWinRTAddAbiNamespacePrefix)'==''">true</AbiWinRTAddAbiNamespacePrefix>
+        <AbiWinRTCommandAddAbiNamespacePrefix Condition="'$(AbiWinRTAddAbiNamespacePrefix)'=='true'">-ns-prefix always</AbiWinRTCommandAddAbiNamespacePrefix>
+
+        <!-- If not generating the platform headers, then default to ensure that there is compat -->
+        <AbiWinRTEnsureSDKHeaderCompat Condition="'$(AbiWinRTEnsureSDKHeaderCompat)'=='' and '$(AbiWinRTEnablePlatformProjection)'=='false'">true</AbiWinRTEnsureSDKHeaderCompat>
+        <AbiWinRTCommandEnsureSDKHeaderCompat Condition="'$(AbiWinRTEnsureSDKHeaderCompat)'=='true'">-lowercase-include-guard</AbiWinRTCommandEnsureSDKHeaderCompat>
+
+        <GeneratedFilesDir Condition="'$(GeneratedFilesDir)' == ''">$(IntDir)Generated Files\</GeneratedFilesDir>
+        <!--Override SDK's uap.props setting to ensure version-matched headers-->
+        <AbiWinRT_IncludePath>$(GeneratedFilesDir)</AbiWinRT_IncludePath>
```
Scottj1s, line 45: >AbiWinRT_IncludePath [](start = 9, length = 20)

review unused properties and remove

src/package/abi/Microsoft.Windows.AbiWinRT.targets
```diff
@@ -6,8 +6,500 @@ Copyright (C) Microsoft Corporation. All rights reserved.
...
+      <Midl Condition="'$(AbiWinRTModernIDL)' != 'false'">
+        <AdditionalMetadataDirectories Condition="'%(AdditionalMetadataDirectories)' == '' And '$(WindowsSDK_MetadataFoundationPath)' != ''">$(WindowsSDK_MetadataFoundationPath);%(AdditionalMetadataDirectories)</AdditionalMetadataDirectories>
+        <AdditionalMetadataDirectories Condition="'%(AdditionalMetadataDirectories)' == '' And '$(WindowsSDK_MetadataFoundationPath)' == ''">$(WindowsSDK_MetadataPath);%(AdditionalMetadataDirectories)</AdditionalMetadataDirectories>
+        <AdditionalOptions>%(AdditionalOptions) /nomidl</AdditionalOptions>
+      </Midl>
+      <Link>
+        <AdditionalDependencies Condition="'$(AbiWinRTLibs)' != 'false'">%(AdditionalDependencies);WindowsApp.lib</AdditionalDependencies>
+      </Link>
+    </ItemDefinitionGroup>
+
```
Scottj1s, line 503: remove all this stuff that's cppwinrt carryover

src/package/abi/Microsoft.Windows.AbiWinRT.targets
stevenbrix: Hey... I wrote this comment myself :)

src/package/abi/Microsoft.Windows.AbiWinRT.targets
stevenbrix: Sure. For my own curiosity, why was this here? I wasn't 100% sure what `CppWinRTLibs` was for, is it for desktop projects that also want to link against WindowsApp.lib to access WinRT APIs?

Scottj1s: I'd say if you're not sure what it is, remove it.  You probably only need ~100 lines in this file.  I don't think you need to be configuring other tools like cl, midlrt, mdmerge, etc.  And you don't need all the platform/component projection targeting.

---
In reply to: [443886013](https://github.com/microsoft/xlang/pull/659#discussion_r443886013) [](ancestors = 443886013)

src/package/abi/Microsoft.Windows.AbiWinRT.IdlGen.targets
```diff
@@ -0,0 +1,83 @@
...
+          Returns="@(AbiWinRTDirectWinMDReferencesForIdl);@(AbiWinRTDirectWinMDReferenceDirs)">
+    <ItemGroup>
+      <_AbiWinRTDirectWinMDReferencesForIdl Remove="@(_AbiWinRTDirectWinMDReferencesForIdl)" />
+      <_AbiWinRTDirectWinMDReferencesForIdl Include="@(Reference)" Condition="'%(Reference.Extension)' == '.winmd' and '%(Reference.IsSystemReference)' != 'true'" />
+      <AbiWinRTDirectWinMDReferencesForIdl Remove="@(AbiWinRTDirectWinMDReferencesForIdl)"/>
+      <AbiWinRTDirectWinMDReferencesForIdl Include="@(_AbiWinRTDirectWinMDReferencesForIdl)">
+        <WinMDPath>%(FullPath)</WinMDPath>
+      </AbiWinRTDirectWinMDReferencesForIdl>
+
+      <!-- Generate list of reference directories to pass into winmdidl.exe -->
```
Scottj1s, line 24: tool name is abi.exe

src/package/abi/Microsoft.Windows.AbiWinRT.IdlGen.targets
```diff
@@ -0,0 +1,83 @@
...
+
+  <!--Build reference IDL from WinMD project references and dynamic library project references
+      Note that Condition is evaluated before DependsOnTargets are run -->
+  <Target Name="AbiWinRTMakeReferenceIdl"
+          Condition="'@(AbiWinRTDirectWinMDReferencesForIdl)' != ''"
+          AfterTargets="GetAbiWinRTDirectWinMDReferencesForIdl"
+          Inputs="@(AbiWinRTDirectWinMDReferencesForIdl)"
+          Outputs="@(_ExpectedGeneratedIdl)">
+    <PropertyGroup>
+      <WindowsSdkToolLocation Condition="'$(WindowsSdkToolLocation)'==''">$(WDKBinRoot)\$(PreferredToolArchitecture)</WindowsSdkToolLocation>
```
Scottj1s, line 57: this is for OS builds - you can remove it and all references to it

src/package/abi/Microsoft.Windows.AbiWinRT.props
```diff
@@ -0,0 +1,58 @@
...
+    <PropertyGroup>
+        <!-- Only do this for MSBuild versions below 16.0
+             as it is since done automatically, see https://github.com/microsoft/msbuild/pull/3605-->
+        <MSBuildAllProjects Condition="'$(MSBuildToolsVersion)'  &lt;= '15'">$(MSBuildAllProjects);$(MSBuildThisFileFullPath)</MSBuildAllProjects>
+        
+        <UsingAbiWinRT>true</UsingAbiWinRT>
+    </PropertyGroup>
+
+     <PropertyGroup>
+        <PreferredToolArchitecture>x64</PreferredToolArchitecture>
```
Scottj1s, line 17: this is cppwinrt-specific

src/package/abi/Microsoft.Windows.AbiWinRT.props
```diff
@@ -0,0 +1,58 @@
...
+        <CanReferenceWinRT>true</CanReferenceWinRT>
+        <IsNativeLanguage>true</IsNativeLanguage>
```
Scottj1s, line 19: why is this needed?

src/package/abi/Microsoft.Windows.AbiWinRT.props
```diff
@@ -0,0 +1,58 @@
...
+        <!-- This causes VS to add the platform WinMD references for the target SDK -->
+        <WinMDAssembly>true</WinMDAssembly>
+        <!--Set a value to prevent SDK's uap.props from setting it and pointing to the wrong headers-->
+        <AbiWinRT_IncludePath>PreventSdkUapPropsAssignment</AbiWinRT_IncludePath>
```
Scottj1s, line 23: make sure you test consuming projects to ensure they don't have colliding SDK ABI headers - that's why this was added for cppwinrt

src/package/abi/Microsoft.Windows.AbiWinRT.props
```diff
@@ -0,0 +1,58 @@
...
+    </PropertyGroup>
+
+    <ItemDefinitionGroup>
+        <ClCompile>
+            <CompileAsWinRT Condition="'%(ClCompile.CompileAsWinRT)' == ''">false</CompileAsWinRT>
+        </ClCompile>
+        <Midl Condition="'$(AbiWinRTModernIDL)' != 'false'">
```
Scottj1s, line 30: can you remove AbiWinRTModernIDL and only add it if necessary later (cppwinrt needed it) - just assume modern is all that's supported?

Scottj1s: the scenarios for the abi nuget are likely to be much narrower than for cppwinrt, so the need for this degree of flexibility is much lower.

src/package/abi/Microsoft.Windows.AbiWinRT.targets
```diff
@@ -6,8 +6,493 @@ Copyright (C) Microsoft Corporation. All rights reserved.
...
+             as it is since done automatically, see https://github.com/microsoft/msbuild/pull/3605-->
+        <MSBuildAllProjects Condition="'$(MSBuildToolsVersion)'  &lt;= '15'">$(MSBuildAllProjects);$(MSBuildThisFileFullPath)</MSBuildAllProjects>
     </PropertyGroup>
 
-</Project>
+    <PropertyGroup>
+        <AbiWinRTVerbosity Condition="'$(AbiWinRTVerbosity)' == ''">normal</AbiWinRTVerbosity>
+        <AbiWinRTCommandVerbosity Condition="'$(AbiWinRTVerbosity)' == 'high'">-verbose</AbiWinRTCommandVerbosity>
+        <AbiWinRTProjectWinMD>$(OutDir)$(RootNamespace).winmd</AbiWinRTProjectWinMD>
+        <AbiWinRTMergedDir>$(IntDir)Merged\</AbiWinRTMergedDir>
```
Scottj1s, line 18: how does all this configuration interact with the cppwinrt nuget?  seems like they would collide badly

src/package/abi/Microsoft.Windows.AbiWinRT.targets
```diff
@@ -6,8 +6,493 @@ Copyright (C) Microsoft Corporation. All rights reserved.
...
+        <AbiWinRTUnmergedDir>$(IntDir)Unmerged\</AbiWinRTUnmergedDir>
+        <AbiWinRTSkipUnchangedFiles Condition="'$(AbiWinRTSkipUnchangedFiles)' == ''">true</AbiWinRTSkipUnchangedFiles>
```
Scottj1s, line 20: skip unchanged files and hardlinks were made optional for cppwinrt to support older versions of msbuild.  I'd recommend removing all this flexibility and asserting your minimal version requirement

src/package/abi/Microsoft.Windows.AbiWinRT.props
```diff
@@ -0,0 +1,58 @@
...
+            <EnableWindowsRuntime>true</EnableWindowsRuntime>
+            <MetadataFileName>$(IntDir)Unmerged\%(Filename).winmd</MetadataFileName>
+            <GenerateClientFiles Condition="'%(Midl.GenerateClientFiles)'==''">None</GenerateClientFiles>
```
Scottj1s, line 33: cppwinrt doesn't support proxy/stub generation, which is why these are all disabled.  I don't think you want to do that for abi.exe

src/package/abi/Microsoft.Windows.AbiWinRT.targets
```diff
@@ -6,8 +6,493 @@ Copyright (C) Microsoft Corporation. All rights reserved.
...
+        <AbiWinRTPath Condition="'$(AbiWinRTPath)'==''">$(MSBuildThisFileDirectory)..\..\bin\</AbiWinRTPath>
+
+        <!-- Default modern idl semantics to false considering that most users probably have not made the switch to the modern idl syntax -->
+        <AbiWinRTModernIDL Condition="'$(AbiWinRTModernIDL)'==''">false</AbiWinRTModernIDL>
+
+        <!-- Only generate the platform and component projection if using modern idl. Otherwise, allow the legacy
+             toolchain to kick in, as well as use the existing references from the Windows SDK -->
+        <AbiWinRTEnableComponentProjection Condition="'$(AbiWinRTEnableComponentProjection)' == '' and '$(AbiWinRTModernIdl)'!='false'">true</AbiWinRTEnableComponentProjection>
+        <AbiWinRTEnablePlatformProjection Condition="'$(AbiWinRTEnablePlatformProjection)' == '' and '$(AbiWinRTModernIdl)'!='false'">true</AbiWinRTEnablePlatformProjection>
+        <AbiWinRTEnableReferenceProjection Condition="'$(AbiWinRTEnableReferenceProjection)' == '' and '$(AbiWinRTModernIdl)'!='false'">true</AbiWinRTEnableReferenceProjection>
```
Scottj1s, line 31: cppwinrt makes a distinction between platform, component, and reference projections which abi.exe should not do.  there should only be a single combined platform/reference projection here (component is a unique cppwinrt concept implying authoring support, not just consuming a third party lib).  and there should be a build assertion to prevent contamination with SDK abi headers.  maybe a static assert or pragma detect mismatch would help

src/package/abi/Microsoft.Windows.AbiWinRT.targets
```diff
@@ -6,8 +6,493 @@ Copyright (C) Microsoft Corporation. All rights reserved.
...
+
+        <!-- Default ABI headers to always be prepended with the ABI namespace -->
+        <AbiWinRTAddAbiNamespacePrefix Condition="'$(AbiWinRTAddAbiNamespacePrefix)'==''">true</AbiWinRTAddAbiNamespacePrefix>
+        <AbiWinRTCommandAddAbiNamespacePrefix Condition="'$(AbiWinRTAddAbiNamespacePrefix)'=='true'">-ns-prefix always</AbiWinRTCommandAddAbiNamespacePrefix>
+
+        <!-- If not generating the platform headers, then default to ensure that there is compat -->
+        <AbiWinRTEnsureSDKHeaderCompat Condition="'$(AbiWinRTEnsureSDKHeaderCompat)'=='' and '$(AbiWinRTEnablePlatformProjection)'=='false'">true</AbiWinRTEnsureSDKHeaderCompat>
```
Scottj1s, line 38: maybe this allows combining with sdk headers?  if so, then the suggested build assert could be conditional

src/package/abi/Microsoft.Windows.AbiWinRT.targets
```diff
@@ -6,8 +6,493 @@ Copyright (C) Microsoft Corporation. All rights reserved.
...
+        <AbiWinRTCommandEnsureSDKHeaderCompat Condition="'$(AbiWinRTEnsureSDKHeaderCompat)'=='true'">-lowercase-include-guard</AbiWinRTCommandEnsureSDKHeaderCompat>
+
+        <GeneratedFilesDir Condition="'$(GeneratedFilesDir)' == ''">$(IntDir)Generated Files\</GeneratedFilesDir>
+        <!--TEMP: Override NuGet SDK's erroneous setting in uap.props -->
```
Scottj1s, line 42: this may no longer be needed - test with nuget SDK distro (and/or ask @manodasanW )

src/package/abi/Microsoft.Windows.AbiWinRT.targets
```diff
@@ -6,8 +6,493 @@ Copyright (C) Microsoft Corporation. All rights reserved.
...
+        <!-- Note: Before* targets run before Compute* targets. -->
+        <ComputeMidlInputsTargets>
+            $(ComputeMidlInputsTargets);AbiWinRTSetMidlReferences;
+        </ComputeMidlInputsTargets>
+        <AfterMidlTargets>
+            $(AfterMidlTargets);
+            GetAbiWinRTMdMergeInputs;
+            AbiWinRTMergeProjectWinMDInputs;
+            GetResolvedWinMD;
+            AbiWinRTCopyWinMDToOutputDirectory;
```
Scottj1s, line 60: this all feels to 'intrusive' - shouldn't this package just perform a 'read-only' scan of a project's winmds (direct, referenced, etc) and create headers?

src/package/abi/Microsoft.Windows.AbiWinRT.targets
```diff
@@ -6,8 +6,493 @@ Copyright (C) Microsoft Corporation. All rights reserved.
...
+        </AfterMidlTargets>
+        <ResolveAssemblyReferencesDependsOn>
+            $(ResolveAssemblyReferencesDependsOn);GetAbiWinRTProjectWinMDReferences;AbiWinRTMarkStaticLibrariesPrivate;
+        </ResolveAssemblyReferencesDependsOn>
+        <!-- Note: Before* targets run before Compute* targets. -->
+        <BeforeClCompileTargets>
+            $(BeforeClCompileTargets);AbiWinRTMakeProjections;
```
Scottj1s, line 67: all ClCompile stuff should be removed - cppwinrt specific

src/package/abi/Microsoft.Windows.AbiWinRT.targets
```diff
@@ -6,8 +6,493 @@ Copyright (C) Microsoft Corporation. All rights reserved.
...
+        </BeforeClCompileTargets>
+        <CleanDependsOn>
+            $(CleanDependsOn);AbiWinRTClean
+        </CleanDependsOn>
+
+    </PropertyGroup>
+
+    <!-- For a static library we don't want the winmd/lib/pdb to be packaged -->
+    <PropertyGroup Condition="'$(ConfigurationType)' == 'StaticLibrary'">
```
Scottj1s, line 76: remove

src/package/abi/Microsoft.Windows.AbiWinRT.targets
```diff
@@ -6,8 +6,493 @@ Copyright (C) Microsoft Corporation. All rights reserved.
...
+        <IncludeCopyWinMDArtifactsOutputGroup>false</IncludeCopyWinMDArtifactsOutputGroup>
+    </PropertyGroup>
+
+    <Target Name="AbiWinRTVerifyKitVersion" Condition="'$(AbiWinRTOverrideSDKReferences)' != 'true'">
```
Scottj1s, line 80: remove all references to AbiWinRTOverrideSDKReferences (cppwinrt carryover for os build support)

Scottj1s, line 80: I'd remove this target - can always add msbuild assertions later if needed

src/package/abi/Microsoft.Windows.AbiWinRT.targets
```diff
@@ -6,8 +6,493 @@ Copyright (C) Microsoft Corporation. All rights reserved.
...
+            <_FilesToDelete Include="$(AbiWinRTUnmergedDir)**"/>
+        </ItemGroup>
+        <Delete Files="@(_FilesToDelete)"/>
+    </Target>
+
+    <!-- Target used only to evaluate AbiWinRTGenerateWindowsMetadata if it doesn't already have a value -->
+    <Target Name="AbiWinRTComputeGenerateWindowsMetadata"
+            DependsOnTargets="GetAbiWinRTProjectWinMDReferences;$(AbiWinRTComputeGenerateWindowsMetadataDependsOn)">
+
+        <PropertyGroup>
```
Scottj1s, line 106: this nuget package is "consume only" - referenced by clients of winmds - all static lib concerns should go

src/package/abi/Microsoft.Windows.AbiWinRT.targets
```diff
@@ -6,8 +6,493 @@ Copyright (C) Microsoft Corporation. All rights reserved.
...
+                <Implementation>$(WinMDImplementationPath)$(TargetName)$(TargetExt)</Implementation>
+                <FileType>winmd</FileType>
+                <WinMDFile>true</WinMDFile>
+                <ProjectType>$(ConfigurationType)</ProjectType>
+            </WinMDFullPath>
+
+            <Link Remove="@(Link)" Condition="'%(Link.DeleteSoon)' == 'true'" />
+        </ItemGroup>
+
+        <!-- Add C++/WinRT primary WinMD to the WinMDFullPath if AbiWinRTGenerateWindowsMetadata is true -->
```
Scottj1s, line 160: update all c++/winrt comments

src/package/abi/Microsoft.Windows.AbiWinRT.targets
```diff
@@ -6,8 +6,493 @@ Copyright (C) Microsoft Corporation. All rights reserved.
...
+
+        <Message Text="GetResolvedWinMD: @(WinMDFullPath->'%(FullPath)')" Importance="$(AbiWinRTVerbosity)"/>
+    </Target>
+
+    <!-- Static library reference WinMDs are merged into the project WinMD that
+         references it and might have the same name because they often share namespace. 
+         Therefore they shouldn't be copied to the output folder
+         because they might override files in the output folder with the
+         same name, causing missing types. -->
+    <Target Name="AbiWinRTMarkStaticLibrariesPrivate"
```
Scottj1s, line 181: remove

src/package/abi/Microsoft.Windows.AbiWinRT.targets
```diff
@@ -6,8 +6,493 @@ Copyright (C) Microsoft Corporation. All rights reserved.
...
+        </ItemGroup>
+        <Message Text="AbiWinRTDirectWinMDReferences: @(AbiWinRTDirectWinMDReferences->'%(WinMDPath)')" Importance="$(AbiWinRTVerbosity)"/>
+    </Target>
+
+    <!--Get direct WinMD project references for projections, IDL processing, and AppX packaging-->
+    <Target Name="GetAbiWinRTProjectWinMDReferences"
+            DependsOnTargets="ResolveProjectReferences;$(GetAbiWinRTProjectWinMDReferencesDependsOn)"
+            Returns="@(AbiWinRTStaticProjectWinMDReferences);@(AbiWinRTDynamicProjectWinMDReferences)">
+        <ItemGroup>
+            <!-- Get static library project references -->
```
Scottj1s, line 244: remove

src/package/abi/Microsoft.Windows.AbiWinRT.targets
```diff
@@ -6,8 +6,493 @@ Copyright (C) Microsoft Corporation. All rights reserved.
...
+    <Target Name="GetAbiWinRTMdMergeInputs"
+                DependsOnTargets="AbiWinRTResolveReferences;"
+                Returns="@(AbiWinRTMdMergeMetadataDirectories);@(AbiWinRTMdMergeInputs)">
+        <ItemGroup>
+            <_MdMergeInputs Remove="@(_MdMergeInputs)"/>
+            <_MdMergeInputs Include="@(Midl)">
+                <WinMDPath>%(Midl.OutputDirectory)%(Midl.MetadataFileName)</WinMDPath>
+                <MdMergeOutputFile>$(AbiWinRTProjectWinMD)</MdMergeOutputFile>
+            </_MdMergeInputs>
+            <!-- Static libraries don't mdmerge other static libraries.
```
Scottj1s, line 282: remove

src/package/abi/Microsoft.Windows.AbiWinRT.props
Scottj1s: now that I think of it, if this nuget is for consumption only, then all midl/mdmerge configuration should go away (like clcompile/link) - you should only be scanning for referenced winmds to generated headers

src/package/abi/Microsoft.Windows.AbiWinRT.targets
stevenbrix: Some amount of `ClCompile` stuff is needed, this is for generating the headers. You also need to update the include paths as well.

src/package/abi/Microsoft.Windows.AbiWinRT.IdlGen.targets
stevenbrix: which one/?

src/package/abi/Microsoft.Windows.AbiWinRT.props
stevenbrix: I've removed this since it isn't needed then. I don't see any collision, the files generated by abiwinrt are consumed

src/package/abi/Microsoft.Windows.AbiWinRT.props
stevenbrix: it's not, i've removed it

src/package/abi/Microsoft.Windows.AbiWinRT.targets
stevenbrix: which build assert?

src/package/abi/Microsoft.Windows.AbiWinRT.targets
stevenbrix: Doesn't look like it is

src/package/abi/Microsoft.Windows.AbiWinRT.targets
stevenbrix: yup, i've updated it since we talked and removed this

src/package/abi/Microsoft.Windows.AbiWinRT.targets
stevenbrix: this nuget provides an error when it detects cpp/winrt nuget package is also being used

src/package/abi/Microsoft.Windows.AbiWinRT.props
stevenbrix: it's not

src/package/abi/Microsoft.Windows.AbiWinRT.props
stevenbrix: going to just resolve all these midl related ones since we agreed a different nuget package is appropriate for that

src/package/abi/Microsoft.Windows.AbiWinRT.targets
stevenbrix: Does the component scenario not make sense for abi.exe since midlrt generates the same headers?

stevenbrix: how do you prevent contamination with sdk abi headers? we could probably remove the sdk abi header location from `ClCompile.AdditionalIncludeDirectories`?

src/package/abi/Microsoft.Windows.AbiWinRT.IdlGen.targets
Scottj1s: n.m - misread this

---
In reply to: [453766996](https://github.com/microsoft/xlang/pull/659#discussion_r453766996) [](ancestors = 453766996)

src/package/abi/Microsoft.Windows.AbiWinRT.targets
stevenbrix: actually I don't think we can do that, since wrl headers are also in that location

src/package/abi/Microsoft.Windows.AbiWinRT.IdlGen.targets
stevenbrix: this was for running winmidl, but is no longer part of this change

src/package/abi/Microsoft.Windows.AbiWinRT.targets
Scottj1s: talk to @manodasanW - he had to tweak the SDK Nuget to prevent collisions with the uap.props additional include dirs

src/package/abi/Microsoft.Windows.AbiWinRT.targets
Scottj1s: can't remember where that comment was - but somewhere you could assert (cause a build warning/error) that ensures SDK WRL headers don't accidentally mix with ABIWinRT headers

src/package/abi/Microsoft.Windows.AbiWinRT.targets
Scottj1s: Yes, additional include path is fine (but nothing else - no /await, /std:c++17, etc)

stevenbrix: gotcha, sweet - yeah i've removed all of that

src/package/abi/Microsoft.Windows.AbiWinRT.targets
stevenbrix: The way I did it was by pre-pending `GeneratedFilesDir` to the `ClCompile.AdditionalIncludeDirectories`. I'll follow up with Mano to see what he suggests. I'd love for an actual assert that isn't dependent on MSBuild import order.

src/package/abi/Microsoft.Windows.AbiWinRT.targets
```diff
@@ -6,8 +6,155 @@ Copyright (C) Microsoft Corporation. All rights reserved.
...
+        <BeforeClCompileTargets>
+            $(BeforeClCompileTargets);AbiWinRTMakeProjections;
+        </BeforeClCompileTargets>
+        <CleanDependsOn>
+            $(CleanDependsOn);AbiWinRTClean
+        </CleanDependsOn>
+
+    </PropertyGroup>
+
+    <!-- Using both cpp/winrt and abi/winrt together is unsupported. -->
```
Scottj1s, line 49: C++/WinRT (or cppwinrt, but not both)

src/tool/abi/readme.md
```diff
@@ -0,0 +1,16 @@
+## The ABI/WinRT language projection
+
+The Application Binary Interface (ABI) projection generates header files for consuming WinRT APIs that are more familiar to headers generated from COM .idl files.
+
+**Note: This tool is for legacy codebases and not recommended for new development. C++ developers should instead use https://github.com/microsoft/cppwinrt, which supports standard c++ and the latest c++ language features.**
```
Scottj1s, line 5: capitalize C++

src/tool/abi/readme.md
```diff
@@ -0,0 +1,16 @@
+## The ABI/WinRT language projection
+
+The Application Binary Interface (ABI) projection generates header files for consuming WinRT APIs that are more familiar to headers generated from COM .idl files.
```
Scottj1s, line 3: "more familiar" -> "similar"

src/package/abi/Microsoft.Windows.AbiWinRT.targets
```diff
@@ -46,7 +46,7 @@ Copyright (C) Microsoft Corporation. All rights reserved.
 
     </PropertyGroup>
 
-    <!-- Using both cpp/winrt and abi/winrt together is unsupported. -->
+    <!-- Using both c++/winrt and abi/winrt together is unsupported. -->
     <Target Name="AbiWinRTValidateNoCppWinRT">
         <Error Condition="'$(CppWinRTPackage)' != ''" Text="Abi/WinRT and Cpp/WinRT create competing projections and are not compatible.
```
Scottj1s, line 51: and this

stevenbrix: 🎵 Scott's got, the eye of the tiger... 🎵

PR #676, opened 2020-07-14
Fix RoResolveNamespace not looking at exe directory on RS5 as fallback. 

kythant:
Types were failing to resolve on RS5 because RoResolveNamespace did not look at the exe file path as fallback by default. Supplying the exe file path to windowsMetaData dir only looks for types with the Windows prefix. The fix was to also supply the exe file path to the package graph parameter. 

Also fix xml parsing to look for 'name' on ActivatableClass tag instead of 'clsid'.

### DIFF
--- a/src/UndockedRegFreeWinRT/UndockedRegFreeWinRT/dllmain.cpp
+++ b/src/UndockedRegFreeWinRT/UndockedRegFreeWinRT/dllmain.cpp
@@ -302,11 +302,12 @@ HRESULT WINAPI RoResolveNamespaceDetour(
     DWORD* subNamespacesCount,
     HSTRING** subNamespaces)
 {
+    HSTRING packageGraphDirectory[1];
+    packageGraphDirectory[0] = Microsoft::WRL::Wrappers::HStringReference(exeFilePath.c_str()).Get();
     HRESULT hr = TrueRoResolveNamespace(name, Microsoft::WRL::Wrappers::HStringReference(exeFilePath.c_str()).Get(),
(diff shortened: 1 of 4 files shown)


src/UndockedRegFreeWinRT/UndockedRegFreeWinRT/dllmain.cpp
```diff
@@ -302,11 +302,12 @@ HRESULT WINAPI RoResolveNamespaceDetour(
...
     HSTRING** subNamespaces)
 {
+    HSTRING packageGraphDirectory[1];
+    packageGraphDirectory[0] = Microsoft::WRL::Wrappers::HStringReference(exeFilePath.c_str()).Get();
     HRESULT hr = TrueRoResolveNamespace(name, Microsoft::WRL::Wrappers::HStringReference(exeFilePath.c_str()).Get(),
-        packageGraphDirsCount, packageGraphDirs,
+        1, packageGraphDirectory,
         metaDataFilePathsCount, metaDataFilePaths,
         subNamespacesCount, subNamespaces);
-
```
jonwis, line 309: This references a temporary created that's destroyed immediately at the `.Get();`

Instead, use:

```c++
auto pathReference = Microsoft::WRL::Wrappers::HStringReference(exeFilePath.c_str());
HSTRING packageGraphDirectories[] = { pathReference.Get() };
```

And then use `ARRAYSIZE(packageGraphDirectories), packageGraphDirectories` as the parameter pair below.

PR #679, opened 2020-07-15
adding midlrt nuget package

### LINKED ISSUE #678: decouple modern idl generation from cpp/winrt
some customers that would like to use modern idl but aren't using cpp/winrt don't have a good way of moving to the newer idl syntax, which is nicer, and has compilation time improvements, without an entire rewrite of their code base. the logic that lives in the `Microsoft.Windows.CppWinRT` nuget package that handles gather the .winmd references and passing them into midlrt should not be something that only ships with C++/WinRT, and made available to other customers (WinUI and Win2D being two of them)

stevenbrix:
Fixes #678 

This PR takes the logic from the `Microsoft.Windows.CppWinRT` nuget package and decouples it from that package, so that other consumers, like Win2D and WinUI, can use it without having to switch all their code over to using C++/WinRT. Ideally once this package is in, the C++/WinRT package removes it's custom logic and takes a dependency on this package.

/cc @Scottj1s

### DIFF
--- a/src/package/abi/LICENSE
+++ b/src/package/abi/LICENSE
@@ -1,4 +1,4 @@
-WinRT ABI Header Generation Tool
+WinRT Metadata Generation Utility Package
 
 Copyright (c) Microsoft Corporation
 All rights reserved. 
--- a/src/package/midlrt/Microsoft.Windows.MidlRT.IdlGen.targets
+++ b/src/package/midlrt/Microsoft.Windows.MidlRT.IdlGen.targets
@@ -1,0 +1,84 @@
+<!--
+***********************************************************************************************
+Copyright (C) Microsoft Corporation. All rights reserved.
+***********************************************************************************************
+-->
+<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+    <!--
+        This file contains targets for generating IDL files from winmd references. This are only
+        used when the consumer hasn't switched to Modern IDL.
+    -->
+    
+  <!--
+    If not using the newer modern idl semantics, generate the .idl for any referenced winmd files.
+    We need a special target here, because the other Get*References  depends on ResolveReferences
+    being run first. However, MIDL runs before ResolveReferences, so this target uses the References
+    item, not the ReferencePath item.
+  -->
+  <Target Name="GetMidlRTDirectWinMDReferencesForIdl"
+          BeforeTargets="Midl"
+          Returns="@(MidlRTDirectWinMDReferencesForIdl);@(MidlRTDirectWinMDReferenceDirs)">
+    <ItemGroup>
+      <_MidlRTDirectWinMDReferencesForIdl Remove="@(_MidlRTDirectWinMDReferencesForIdl)" />
+      <_MidlRTDirectWinMDReferencesForIdl Include="@(Reference)" Condition="'%(Reference.Extension)' == '.winmd' and '%(Reference.IsSystemReference)' != 'true'" />
+      <MidlRTDirectWinMDReferencesForIdl Remove="@(MidlRTDirectWinMDReferencesForIdl)"/>
+      <MidlRTDirectWinMDReferencesForIdl Include="@(_MidlRTDirectWinMDReferencesForIdl)">
+        <WinMDPath>%(FullPath)</WinMDPath>
+      </MidlRTDirectWinMDReferencesForIdl>
+
+      <!-- Generate list of reference directories to pass into winmdidl.exe -->
+      <_MidlRTDirectWinMDReferenceDirs Include="@(MidlRTDirectWinMDReferencesForIdl->'%(RootDir)%(Directory)')"/>
+    </ItemGroup>
+
+    <!-- Remove any duplicates of file paths (caused by a nuget containing mulitple .winmd files) -->
+    <RemoveDuplicates Inputs="@(_MidlRTDirectWinMDReferenceDirs)">
+      <Output TaskParameter="Filtered" ItemName="MidlRTDirectWinMDReferenceDirs"/>
+    </RemoveDuplicates>
+    <Message Text="MidlRTDirectWinMDReferencesForIdl: @(MidlRTDirectWinMDReferencesForIdl->'%(WinMDPath)')" Importance="$(MidlRTVerbosity)"/>
+  </Target>
+
+    <!-- 
+      Gather the outputs from the target if previously executed. If it hasn't been executed, then this list
+      will be empty and the target will run.
+    -->
+  <Target Name="_GatherExpectedGeneratedFiles" BeforeTargets="MidlRTMakeReferenceIdl" Returns="@(_ExpectedGeneratedIdl)">
+    <ItemGroup>
+      <_ExpectedGeneratedIdl Include="$(GeneratedFilesDir)\*.idl"/>
+    </ItemGroup>
+  </Target>
+
+  <!--Build reference IDL from WinMD project references and dynamic library project references
+      Note that Condition is evaluated before DependsOnTargets are run -->
+  <Target Name="MidlRTMakeReferenceIdl"
+          Condition="'@(MidlRTDirectWinMDReferencesForIdl)' != ''"
+          AfterTargets="GetMidlRTDirectWinMDReferencesForIdl"
+          Inputs="@(MidlRTDirectWinMDReferencesForIdl)"
+          Outputs="@(_ExpectedGeneratedIdl)">
+    <PropertyGroup>
+      <WindowsSdkToolLocation Condition="'$(WindowsSdkToolLocation)'==''">$(WDKBinRoot)\$(PreferredToolArchitecture)</WindowsSdkToolLocation>
+    </PropertyGroup>
+    <ItemGroup>
+      <_MidlRTRefInputs Remove="@(_MidlRTRefInputs)"/>
+      <_MidlRTRefInputs Include="@(MidlRTDirectWinMDReferencesForIdl)"/>
+
+      <_MidlRTMetadataDirInputs Include="$(WindowsSDK_UnionMetadataPath)"/>
+      <_MidlRTMetadataDirInputs Include="@(MidlRTDirectWinMDReferenceDirs)"/>
+    </ItemGroup>
+
+    <!-- Since the backslash is an escape character, we need to append a '.' to the end of each path so that they are properly formed -->
+    <PropertyGroup>
+      <_MidlRTWinmdIdlParams>/outdir:&quot;$(GeneratedFilesDir).&quot;</_MidlRTWinmdIdlParams>
+      <_MidlRTWinmdIdlParams>$(_MidlRTWinmdIdlParams) @(_MidlRTMetadataDirInputs->'/metadata_dir:&quot;%(Identity).&quot;', ' ')</_MidlRTWinmdIdlParams>
+    </PropertyGroup>
+
+    <MakeDir Directories="$(GeneratedFilesDir)" />
+    <Exec Command="&quot;$(WindowsSdkToolLocation)\winmdidl.exe&quot; $(_MidlRTWinmdIdlParams) %(_MidlRTRefInputs.Identity)"
+          Condition="'@(_MidlRTRefInputs)' != ''"/>
+  </Target>
+
+  <ItemDefinitionGroup>
+    <Midl>
+      <AdditionalIncludeDirectories>%(AdditionalIncludeDirectories);$(GeneratedFilesDir)</AdditionalIncludeDirectories>
+    </Midl>
+  </ItemDefinitionGroup>
+</Project>
--- a/src/package/midlrt/Microsoft.Windows.MidlRT.props
+++ b/src/package/midlrt/Microsoft.Windows.MidlRT.props
@@ -1,0 +1,42 @@
+<!--
+***********************************************************************************************
+Copyright (C) Microsoft Corporation. All rights reserved.
+***********************************************************************************************
+-->
+<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+
+    <PropertyGroup>
+        <!-- Only do this for MSBuild versions below 16.0
+             as it is since done automatically, see https://github.com/microsoft/msbuild/pull/3605-->
+        <MSBuildAllProjects Condition="'$(MSBuildToolsVersion)'  &lt;= '15'">$(MSBuildAllProjects);$(MSBuildThisFileFullPath)</MSBuildAllProjects>
+        
+        <UsingMidlRTPackage>true</UsingMidlRTPackage>
+    </PropertyGroup>
+
+     <PropertyGroup>
+        <PreferredToolArchitecture>x64</PreferredToolArchitecture>
+        <CanReferenceWinRT>true</CanReferenceWinRT>
+        <IsNativeLanguage>true</IsNativeLanguage>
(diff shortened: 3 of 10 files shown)


src/package/abi/LICENSE
```diff
@@ -1,4 +1,4 @@
-WinRT ABI Header Generation Tool
+WinRT Metadata Generation Utility Package
```
stevenbrix, line 1: oops, this is the wrong file to update

DefaultRyan:
One thing surprised me. It sounded like you implied that non-modern IDL can't/won't import winmds. Is that true? It seems like I should know, but you're the one that's testing this. :)

src/package/midlrt/readme.md
```diff
@@ -0,0 +1,26 @@
+## The Windows Runtime (WinRT) API Genaration tool
```
DefaultRyan, line 1: Typo - Generation

stevenbrix:
> One thing surprised me. It sounded like you implied that non-modern IDL can't/won't import winmds. Is that true? It seems like I should know, but you're the one that's testing this. :)

It's a bit confusing, like what constitutes "modern IDL"? When I think about it, there are generally two things that come to mind:
1. Nicer syntax that is closer to C#
2. Faster compilation times via referencing winmd files

I know that you can have the modern syntax without referencing winmd files (WinUI does this today), but it seems that having 2 without 1 has some problems. I know the documentation says you can move from MIDL2.0 to MIDL3.0 syntax, but I wasn't able to get it to work.

src/package/midlrt/Microsoft.Windows.MidlRT.IdlGen.targets
```diff
@@ -0,0 +1,115 @@
+<!--
+***********************************************************************************************
+Copyright (C) Microsoft Corporation.
+***********************************************************************************************
+-->
+<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <!--
+    This file contains targets for generating IDL files from winmd references. This are only
```
Scottj1s, line 8: This are

src/package/midlrt/Microsoft.Windows.MidlRT.IdlGen.targets
```diff
@@ -0,0 +1,115 @@
...
+
+  <Target Name="_GatherExpectedGeneratedFiles"
+          Condition="'@(MidlRTDirectWinMDReferencesForIdl)' != ''"
+          BeforeTargets="MidlRTMakeReferenceIdl"
+          Returns="@(_ExpectedGeneratedIdl)">
+    <ReadLinesFromFile File="$(_MidlRTIdlGenTrackerFile)" >
+      <Output TaskParameter="Lines" ItemName="_ExpectedGeneratedIdl"/>
+    </ReadLinesFromFile>
+
+    <!-- If no files we're read, then we haven't run this target before, so include a fake so the MidlRTMakeReferenceIdl target will execute -->
```
Scottj1s, line 70: were

src/package/midlrt/Microsoft.Windows.MidlRT.IdlGen.targets
```diff
@@ -0,0 +1,115 @@
...
+
+  <!--Build reference IDL from WinMD project references and dynamic library project references
+      Note that Condition is evaluated before DependsOnTargets are run -->
+  <Target Name="MidlRTMakeReferenceIdl"
+          Condition="'@(MidlRTDirectWinMDReferencesForIdl)' != ''"
+          AfterTargets="GetMidlRTDirectWinMDReferencesForIdl"
+          Inputs="@(MidlRTDirectWinMDReferencesForIdl)"
+          Outputs="@(_ExpectedGeneratedIdl)">
+    <PropertyGroup>
+      <WindowsSdkToolLocation Condition="'$(WindowsSdkToolLocation)'==''">$(WDKBinRoot)\$(PreferredToolArchitecture)</WindowsSdkToolLocation>
```
Scottj1s, line 84: why is this using a WDK property?

src/package/midlrt/Microsoft.Windows.MidlRT.props
```diff
@@ -0,0 +1,42 @@
...
+
+    <PropertyGroup>
+        <!-- Only do this for MSBuild versions below 16.0
+             as it is since done automatically, see https://github.com/microsoft/msbuild/pull/3605-->
+        <MSBuildAllProjects Condition="'$(MSBuildToolsVersion)'  &lt;= '15'">$(MSBuildAllProjects);$(MSBuildThisFileFullPath)</MSBuildAllProjects>
+        
+        <UsingMidlRTPackage>true</UsingMidlRTPackage>
+    </PropertyGroup>
+
+     <PropertyGroup>
```
Scottj1s, line 16: I doubt any of these are needed

src/package/midlrt/Microsoft.Windows.MidlRT.props
```diff
@@ -0,0 +1,42 @@
...
+        <PreferredToolArchitecture>x64</PreferredToolArchitecture>
+        <CanReferenceWinRT>true</CanReferenceWinRT>
+        <IsNativeLanguage>true</IsNativeLanguage>
+        <!-- This causes VS to add the platform WinMD references for the target SDK -->
+        <WinMDAssembly>true</WinMDAssembly>
+    </PropertyGroup>
+
+    <ItemDefinitionGroup>
+        <ClCompile>
```
Scottj1s, line 25: remove?

src/package/midlrt/Microsoft.Windows.MidlRT.targets
```diff
@@ -0,0 +1,339 @@
...
+        <ResolveAssemblyReferencesDependsOn>
+            $(ResolveAssemblyReferencesDependsOn);GetMidlRTProjectWinMDReferences;MidlRTMarkStaticLibrariesPrivate;
+        </ResolveAssemblyReferencesDependsOn>
+        <CleanDependsOn>
+            $(CleanDependsOn);MidlRTClean
+        </CleanDependsOn>
+
+    </PropertyGroup>
+
+    <!-- For a static library we don't want the winmd/lib/pdb to be packaged -->
```
Scottj1s, line 47: remove?

src/package/midlrt/Microsoft.Windows.MidlRT.targets
```diff
@@ -0,0 +1,339 @@
...
+    <Target Name="ComputeGetResolvedWinMD"
+            Condition="'$(MidlRTGenerateWindowsMetadata)' == ''">
+        <!-- If MidlRTGenerateWindowsMetadata is not defined, compute it.-->
+        <!-- We use Calltarget, so we don't run anything including DependsOnTargets
+             targets if $(MidlRTGenerateWindowsMetadata) already has a value.-->
+        <CallTarget Targets="MidlRTComputeGenerateWindowsMetadata" />
+    </Target>
+
+    <!-- This target overrides the GetResolvedWinMD target used to resolve the WinMD for native projects
+         so it is aware of the C++/WinRT generated WinMD.
```
Scottj1s, line 91: any comments mentioning "C++/WinRT" should be reviewed.

src/package/midlrt/Microsoft.Windows.MidlRT.targets
```diff
@@ -0,0 +1,339 @@
...
+                DependsOnTargets="MidlRTResolveReferences;"
+                Returns="@(MidlRTMdMergeMetadataDirectories);@(MidlRTMdMergeInputs)">
+        <ItemGroup>
+            <_MdMergeInputs Remove="@(_MdMergeInputs)"/>
+            <_MdMergeInputs Include="@(Midl)">
+                <WinMDPath>%(Midl.OutputDirectory)%(Midl.MetadataFileName)</WinMDPath>
+                <MdMergeOutputFile>$(MidlRTProjectWinMD)</MdMergeOutputFile>
+            </_MdMergeInputs>
+            <!-- Static libraries don't mdmerge other static libraries.
+                 Instead they are passed as independent inputs for the component projection. -->
```
Scottj1s, line 231: "component projection" ?

src/package/midlrt/Microsoft.Windows.MidlRT.targets
```diff
@@ -0,0 +1,339 @@
...
+
+    <!--Merge project-generated WinMDs and project-referenced static library WinMDs into project WinMD-->
+    <Target Name="MidlRTMergeProjectWinMDInputs"
+            DependsOnTargets="Midl;GetMidlRTMdMergeInputs;$(MidlRTMergeProjectWinMDInputsDependsOn)"
+            Inputs="$(MSBuildAllProjects);@(MidlRTMdMergeInputs)"
+            Outputs="@(_MdMergedOutput);$(IntDir)mdmerge.rsp">
+        <PropertyGroup>
+            <!--Note: MidlRTNamespaceMergeDepth supersedes MidlRTMergeDepth-->
+            <_MdMergeDepth Condition="'$(MidlRTNamespaceMergeDepth)' != ''">-n:$(MidlRTNamespaceMergeDepth)</_MdMergeDepth>
+            <_MdMergeDepth Condition="'$(_MdMergeDepth)' == ''">$(MidlRTMergeDepth)</_MdMergeDepth>
```
Scottj1s, line 290: since this is new, don't need to carry the c++/winrt baggage (delete MidlRTMergeDepth)

src/package/midlrt/Microsoft.Windows.MidlRT.targets
```diff
@@ -0,0 +1,339 @@
...
+    <!--Ctrl+F7 (selected file) midl compilation support-->
+    <Target Name="MidlRTSetSelectMidlReferences" BeforeTargets="SelectMidl" DependsOnTargets="MidlRTSetMidlReferences" />
+
+    <!--Merge project-generated WinMDs and project-referenced static library WinMDs into project WinMD-->
+    <Target Name="MidlRTMergeProjectWinMDInputs"
+            DependsOnTargets="Midl;GetMidlRTMdMergeInputs;$(MidlRTMergeProjectWinMDInputsDependsOn)"
+            Inputs="$(MSBuildAllProjects);@(MidlRTMdMergeInputs)"
+            Outputs="@(_MdMergedOutput);$(IntDir)mdmerge.rsp">
+        <PropertyGroup>
+            <!--Note: MidlRTNamespaceMergeDepth supersedes MidlRTMergeDepth-->
```
Scottj1s, line 288: delete note and MidlRTMergeDepth

src/package/midlrt/readme.md
```diff
@@ -0,0 +1,26 @@
+## The Windows Runtime (WinRT) API Generation tool
```
Scottj1s, line 1: update title

src/package/midlrt/Microsoft.Windows.MidlRT.targets
```diff
@@ -0,0 +1,339 @@
...
+        <MSBuildAllProjects Condition="'$(MSBuildToolsVersion)'  &lt;= '15'">$(MSBuildAllProjects);$(MSBuildThisFileFullPath)</MSBuildAllProjects>
+    </PropertyGroup>
+
+    <PropertyGroup>
+        <MidlRTProjectWinMD>$(OutDir)$(RootNamespace).winmd</MidlRTProjectWinMD>
+        <MidlRTMergedDir>$(IntDir)Merged\</MidlRTMergedDir>
+        <MidlRTUnmergedDir>$(IntDir)Unmerged\</MidlRTUnmergedDir>
+
+        <ModernMidlRT Condition="'$(ModernMidlRT)'==''">true</ModernMidlRT>
+
```
Scottj1s, line 20: to be safe, you should normalize:
        <ModernMidlRT Condition="'$(ModernMidlRT)' != 'true'">false</ModernMidlRT>

otherwise comparisons with 'false' below might fail

src/package/midlrt/Microsoft.Windows.MidlRT.targets
stevenbrix: Gotcha, sounds good!

src/package/midlrt/Microsoft.Windows.MidlRT.IdlGen.targets
stevenbrix: i'll have to take a look, i think this is what's set in uap.props that points to where winmdidl is

src/package/midlrt/Microsoft.Windows.MidlRT.props
stevenbrix: remove what? sorry, i can't really tell where your comment is :/

src/package/midlrt/Microsoft.Windows.MidlRT.props
stevenbrix: I like having the `UsingFoo` properties just because they can be used by consuming projects.

src/package/midlrt/readme.md
stevenbrix: to what exactly? i thought this was a fairly good description of what MIDL does?

stevenbrix: it's API generation, not projection generation

src/package/midlrt/Microsoft.Windows.MidlRT.IdlGen.targets
stevenbrix: it was, but it looks like it was removed in the latest sdk, so i'll change this

src/package/midlrt/Microsoft.Windows.MidlRT.props
Scottj1s: oh - remove the ClCompile item config

src/package/midlrt/readme.md
Scottj1s: oh n.m. - thought this had been copied from the ABI tool

src/package/midlrt/Microsoft.Windows.MidlRT.props
Scottj1s: how many of these are needed (were some copied from cppwinrt?):

```
        <PreferredToolArchitecture>x64</PreferredToolArchitecture>
        <CanReferenceWinRT>true</CanReferenceWinRT>
        <IsNativeLanguage>true</IsNativeLanguage>
        <!-- This causes VS to add the platform WinMD references for the target SDK -->
        <WinMDAssembly>true</WinMDAssembly>
```

src/package/midlrt/Microsoft.Windows.MidlRT.props
stevenbrix: oh yes, gotcha!

src/package/midlrt/Microsoft.Windows.MidlRT.targets
stevenbrix: good catch - we don't need `MidlRTGenerateWindowsMetadata` because that's the only thing this package is for :)

src/package/midlrt/Microsoft.Windows.MidlRT.props
stevenbrix: `CanReferenceWinRT` is what VS uses to make sure your project can reference a .winmd - seems important to have. You could argue this would be set by projection packages - but it doesn't hurt

`WinMDAssembly`  is what VS uses to know this project produces a winmd - seems reasonable

`PreferredToolArchitecture` - we use preferred tool architecture in this nuget package

looks like we don't need  `IsNativeLanguage`

Issue #687, opened 2020-08-12
Enable Undocked Regfree WinRT to work on Win8.1 and Win7


AdamBraden:
The detours library that enables undocked regfree winrt could in theory work down to Win7 - take a closer look to see if this is possible - for both Win8.1 and Win7. 

Win8.1 currently errors with Error 0x8007109A – This operation is only valid in the context of an app container.

Win7 fails to load the component - possibly due to api set issues.

kythant:
Windows 8.1
0x8007109A error is caused by the app container check that exists in windows 8.1 but not on later versions.
This can be fixed by, setting the linker option of your component that you're activating to /APPCONTAINER:NO
https://docs.microsoft.com/en-us/cpp/build/reference/appcontainer-windows-store-app?view=vs-2019
https://docs.microsoft.com/en-us/windows/uwp/debug-test-perf/windows-app-certification-kit-tests#appcontainercheck

Windows 7
Because the RoAPIs are missing on windows 7 and the project is based on detouring, it does not seem worthwhile to design a whole new version of the project just for win7 once you take into account that there is no customer demand for win7 support.

PR #693, opened 2020-09-03
RoGetMetaDataFile to find non-windows type in exe directory 

kythant:
Fix RoGetMetaDataFile so that it find non windows type definition in the WinMD that is in the executable's directory. 

This fixes cross apartment activations on RS5


src/UndockedRegFreeWinRT/UndockedRegFreeWinRT/typeresolution.cpp
```diff
@@ -0,0 +1,657 @@
...
+                        }
+                    }
+                }
+                else if (hr == CLDB_E_RECORD_NOTFOUND)
+                {
+                    hr = RO_E_METADATA_NAME_NOT_FOUND;
+                }
+            }
+
+            if ((hr == RO_E_METADATA_NAME_NOT_FOUND) &&
```
EricJohnson327, line 146: Is the entire point of this block to selectively update the RO_E_METADATA_NAME_NOT_FOUND result to RO_E_METADATA_NAME_IS_NAMESPACE?  No issue with it, just making sure I understand correctly.

src/UndockedRegFreeWinRT/UndockedRegFreeWinRT/typeresolution.cpp
```diff
@@ -0,0 +1,657 @@
...
+        __in IMetaDataDispenserEx* pMetaDataDispenser,
+        __in PCWSTR pszFullName,
+        __in PCWSTR pszDirectoryPath,
+        __out_opt HSTRING* phstrMetaDataFilePath,
+        __deref_opt_out_opt IMetaDataImport2** ppMetaDataImport,
+        __out_opt mdTypeDef* pmdTypeDef)
+    {
+        HRESULT hr;
+
+        wchar_t szCandidateFilePath[MAX_PATH + 1] = { 0 };
```
EricJohnson327, line 219: In GetProcessExeDirInitOnceCallback() you use unique_process_heap_string to account for large filenames.  Do you need to do that here too?

src/UndockedRegFreeWinRT/UndockedRegFreeWinRT/typeresolution.cpp
```diff
@@ -0,0 +1,657 @@
...
+
+        for (int i = elementIndex - 1; i >= 0; i--)
+        {
+            _arFilePaths[i + 1] = _arFilePaths[i];
+        }
+
+        _arFilePaths[0] = pszFoundFilePath;
+    }
+
+    HRESULT MetaDataImportersLRUCache::RemoveLeastRecentlyUsedItemIfListIsFull()
```
EricJohnson327, line 639: If the list is full, will we not check all metadata?

src/UndockedRegFreeWinRT/UndockedRegFreeWinRT/typeresolution.cpp
kythant: I think because were not calling GetModuleFileNameW here that it will be fine?

src/UndockedRegFreeWinRT/UndockedRegFreeWinRT/typeresolution.cpp
kythant: No it will right? This is basically guaranteeing for AddNewFilePathToList that it can add a new file path and the list won't be full.

src/UndockedRegFreeWinRT/UndockedRegFreeWinRT/typeresolution.cpp
kythant: Looks like it :)

src/UndockedRegFreeWinRT/UndockedRegFreeWinRT/typeresolution.cpp
EricJohnson327: Couldn't it in theory add enough new files that the first one get's pushed out?  Meaning we wouldn't then search that file?  Is this even a scenario of concern?  I'm not sure how many files the list can hold, but it's likely plenty.

kythant: It should only hold up to g_dwMetaDataImportersLRUCacheSize. It's based on least recently used (LRU). So in theory, GetMetaDataImporter should be removing the least used one if the size is full. But if it's get called again on the winmd that was removed, it should readd it to the cache by calling GetNewMetaDataImporter on it.

EricJohnson327 (approved):

src/UndockedRegFreeWinRT/UndockedRegFreeWinRT/typeresolution.h
```diff
@@ -0,0 +1,143 @@
...
+#include "Synchapi.h"
+#include <unordered_map>
+
+#define METADATA_FILE_EXTENSION L"winmd"
+#define METADATA_FILE_PATH_FORMAT L"%s%s."  METADATA_FILE_EXTENSION
+#define METADATA_FILE_SEARCH_FORMAT L"%s%s*."  METADATA_FILE_EXTENSION
+
+namespace UndockedRegFreeWinRT
+{
+    // This number comes from (c_cchTypeNameMax = 512) in //depot/fbl_ie_dev1/com/WinRT/buildtools/mdmerge/MetadataScope.h
```
DefaultRyan, line 20: This looks like some pretty inside information. Probably shouldn't do that.

kythant: Oh yea let me comb through the headers to make sure

src/UndockedRegFreeWinRT/UndockedRegFreeWinRT/typeresolution.cpp
```diff
@@ -0,0 +1,653 @@
...
+        __out_opt mdTypeDef* pmdTypeDef)
+    {
+        HRESULT hr = HRESULT_FROM_WIN32(APPMODEL_ERROR_NO_PACKAGE);
+        UINT32 dwPackagesCount = 0;
+        UINT32 dwBufferLength = 0;
+
+        PCWSTR exeDir = nullptr;  // Never freed; owned by process global.
+        RETURN_IF_FAILED(GetProcessExeDir(&exeDir));
+
+        hr = FindTypeInDirectoryWithNormalization(
```
brialmsft, line 435: Would it make sense to restrict exe dir resolutions to only apply to unpackaged apps? That's how it's done in the OS-provided API, and I think consistency with that behavior should be our default approach. Packaged apps can have their exe on their package graph, which will allow type resolution against adjacent WinMDs.

src/UndockedRegFreeWinRT/UndockedRegFreeWinRT/typeresolution.h
```diff
@@ -0,0 +1,143 @@
...
+    static const UINT32 g_uiMaxTypeName = 512;
```
DefaultRyan, line 20: I don't see why this needs to be in the header. It's only used in the cpp implementation.

kythant: Good point

brialmsft (changes requested):

src/UndockedRegFreeWinRT/UndockedRegFreeWinRT/typeresolution.h
kythant: Moved a bunch of these to the cpp implementation per suggestion

src/UndockedRegFreeWinRT/UndockedRegFreeWinRT/typeresolution.cpp
```diff
@@ -0,0 +1,653 @@
+#pragma once
+
+#include <RoMetadataApi.h>
+#include <hstring.h>
+#include <atlbase.h>
```
DefaultRyan, line 5: ATL? Is there really a need to take a dependency on that? WRL and WIL provide much more modern tools for doing what you want.

kythant: I think removed it and I just forgot to remove this header.

src/UndockedRegFreeWinRT/UndockedRegFreeWinRT/typeresolution.cpp
```diff
@@ -0,0 +1,653 @@
...
+    HRESULT FindTypeInMetaDataFile(
+        __in IMetaDataDispenserEx* pMetaDataDispenser,
+        __in PCWSTR pszFullName,
+        __in PCWSTR pszCandidateFilePath,
+        __in TYPE_RESOLUTION_OPTIONS resolutionOptions,
+        __deref_opt_out_opt IMetaDataImport2** ppMetaDataImport,
+        __out_opt mdTypeDef* pmdTypeDef)
+    {
+        HRESULT hr = S_OK;
+        CComPtr<IMetaDataImport2> spMetaDataImport;
```
DefaultRyan: CComPtr looks like an ATL type. You should prefer to use WRL or WIL instead.

src/UndockedRegFreeWinRT/UndockedRegFreeWinRT/catalog.cpp
```diff
@@ -177,18 +178,11 @@ HRESULT WinRTGetMetadataFile(
...
+    mdTypeDef* typeDefToken)
 {
-    if (metaDataDispenser != nullptr ||
-        metaDataImport != nullptr ||
-        typeDefToken != nullptr)
-    {
-        return E_NOTIMPL;
-    }
-
     wchar_t szFolderPath[9];
```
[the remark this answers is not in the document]
kythant: renamed to folderPrefix

src/UndockedRegFreeWinRT/UndockedRegFreeWinRT/typeresolution.cpp
kythant: It was! I just did while trying to remove the header.

src/UndockedRegFreeWinRT/UndockedRegFreeWinRT/typeresolution.cpp
```diff
@@ -0,0 +1,653 @@
...
+
+        RETURN_IF_WIN32_BOOL_FALSE(InitOnceExecuteOnce(&ProcessExeDirInitOnce, GetProcessExeDirInitOnceCallback, nullptr, nullptr));
+
+        // The cache has been successfully populated by the InitOnce, so we can just use it directly.
+        *path = g_cachedProcessExeDir.get();
+        return S_OK;
+    }
+
+    HRESULT FindTypeInMetaDataFile(
+        __in IMetaDataDispenserEx* pMetaDataDispenser,
```
DefaultRyan, line 55: If you're going to use SAL annotations, use the publicly documented updated annotations.
```
__in -> _In_
__deref_opt_out_opt -> _COM_Outptr_opt_result_maybenull_
__out_opt -> _Out_opt__
```

src/UndockedRegFreeWinRT/UndockedRegFreeWinRT/catalog.cpp
```diff
@@ -197,7 +191,53 @@ HRESULT WinRTGetMetadataFile(
...
+        }
+        return UndockedRegFreeWinRT::ResolveThirdPartyType(
+            spMetaDataDispenser.Get(),
+            pszFullName,
+            metaDataFilePath,
+            metaDataImport,
+            typeDefToken);
+    }
+
     wil::unique_cotaskmem_array_ptr<wil::unique_hstring> metaDataFilePaths;
```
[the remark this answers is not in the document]
kythant: I think I can do work here to make this flow better for sure.

src/UndockedRegFreeWinRT/UndockedRegFreeWinRT/typeresolution.cpp
kythant: Are SAL annotations still useful here in github though?

src/UndockedRegFreeWinRT/UndockedRegFreeWinRT/typeresolution.cpp
kythant: That depends on how it works down level to rs2 and windows 8.1.
https://github.com/microsoft/xlang/issues/687

Would the old os versions be able to handle it for the packaged apps?

brialmsft: Yes, packaged type resolution was the original goal of these APIs, and its behavior has been mostly unchanged since Win8. Unpackaged type resolution was the new thing added with reg-free WinRT, so it's the only thing that you should need to undock and take downlevel via detours.

kythant: Alright then it makes sense to add the check then.

brialmsft (approved):

PR #694, opened 2020-09-04
Simplify Initialization Process of RegFreeWinRT on C#

kythant:


src/UndockedRegFreeWinRT/mwinrtact/mwinrtact.cs
```diff
@@ -0,0 +1,15 @@
+﻿using System.Runtime.InteropServices;
+
+namespace Microsoft.Windows
+{
+    public static class UndockedRegFreeWinrt
+    {
+        [DllImport("winrtact.dll")]
```
AdamBraden, line 7: Do we want a failed HRESULT to throw in this case? If so, we need PreserveSig=false on the DllImport
https://docs.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.dllimportattribute.preservesig?view=netcore-3.1

kythant: I think if something goes wrong with initializing the detours and loading the catalog. We want to silently fail?

kythant: @BenJKuhn What do you think?

BenJKuhn: The app has to opt into this functionality, and if the manifests don't load, the app is likely to break in strange ways. So I think an exception here are reasonable and appropriate.

src/UndockedRegFreeWinRT/Nuget/Microsoft.Windows.UndockedRegFreeWinrt.nuspec
```diff
@@ -24,6 +24,7 @@
     <file src="$winrtactx64debug$" target="runtimes\x64\debug\"/>
     <file src="$winrtactx86release$" target="runtimes\x86\release\"/>
     <file src="$winrtactx86debug$" target="runtimes\x86\debug\"/>
+    <file src="$mwinrtactx64release$" target="lib\netstandard2.0"/>
```
BenJKuhn, line 27: is there a reason for not including an x86 version as well? Or is this really "any cpu"?

kythant: I don't really understand how it's working but when I tested it, it seems to work on any CPU. I don't know if I am doing this nuspec part properly by targeting "lib\netstandard2.0". If you know someone who is expert in this, let me know.

BenJKuhn: I believe this is what you want. Targeting .Net standard 2.0 will provide compat with a wide range of .NET targets, if I understand correctly. You can find more about it here: https://docs.microsoft.com/en-us/nuget/reference/target-frameworks.

BenJKuhn (approved):

PR #695, opened 2020-09-07
Send unraisable Python exceptions to the proper handler

zooba:
Exceptions raised during event handlers are _practically_ never going to reappear in Python code (the exceptions are those directly triggered by other Python code, such as `StringMap.map_changed`, but we don't have a great way to identify those).

Instead, [PyErr_WriteUnraisable](https://docs.python.org/3/c-api/exceptions.html#c.PyErr_WriteUnraisable) will format an "Exception ignored in ..." message by default, and can be hooked by developers to handle it specially. It also clears the error state after reporting, which prevents superfluous errors if we happen to end up back in a Python thread.


zooba:
I bet we could throw a better exception than `hresult_invalid_argument` but I don't know what it would be... `E_FAIL`? Make up our own HRESULT?

BenJKuhn (approved):

BenJKuhn:
> I bet we could throw a better exception than `hresult_invalid_argument` but I don't know what it would be... `E_FAIL`? Make up our own HRESULT?

From what you're describing, would E_UNEXPECTED be appropriate?

zooba:
> From what you're describing, would E_UNEXPECTED be appropriate?

Maybe? My understanding was always that E_UNEXPECTED was a "this should never happen" kind of error code, while this situation is more of a generic failure that has nowhere to propagate.

BenJKuhn (approved):

PR #697, opened 2020-09-11
[draft] Side-by-side Windows Runtime Components spec

BenJKuhn:
This is a draft specification based on the basic design concepts described here: https://github.com/microsoft/ProjectReunion/issues/153

This is a high-level design which will require feature work in multiple projects to fully implement. Detailed designs of changes required in related projects will be developed using normal maintenance processes for each project, and will not be fully specified here. However, this design document will provide a sufficient description of the desired experience to design the changes in each project, and ensure that they result in a coherent and high-quality developer experience.

I'm opening this pull request with open issues to encourage feedback and contributions from people working on related projects.

### DIFF
--- a/docs/winrt/WinRT SxS Components.md
+++ b/docs/winrt/WinRT SxS Components.md
@@ -1,1 +1,153 @@
-## Placeholder
+# Side-by-Side Versioned Components
+
+## Status
+Proposed: This document describes a proposed design that has not been implemented.
+
+## Background
+The Windows Runtime type system was largely designed with the needs of opertating system components in mind. Although applications can define their own API types using the Windows Runtime type system, the feature set for doing so is somewhat basic. Versioning strategies are only vaguely defined, and no provisions are made for side-by-side support of a component. As the community has developed more sophisticated libraries over time, these deficiencies are becoming more challenging, and further support is needed.
+
+Today, the Windows Runtime only supports binary-compatible versioning. That is, if a an application or component is developed with a dependency on version N of another component, version N+1 *can* be developed in such a way that code designed to use version N can leverage version N+1 without changes or recompilation of the client component. Alternately, a component author can ignore binary compatbility (which comes with trade-offs) and require that applications use a single, specific version of the component. 
+
+Up to this point, this has been largely sufficient. The single largest "component" is the Windows operating system, which has processes in place to ensure binary compatbility. Other components have typically been distributed as part of an MSIX packaged application. he expectation is that the application developer can manage the dependency graph. This has largely been true up to this point. However, authors of new components are recognizing that they will be consumed in more complex dependency graphs, where the application may have dependencies on multiple versions of the component, and the component author does not want to be locked into a binary-stable model. This implies the need for some form of side by side support.
+
+## Terminology
+* Cross-talk - any ambiguity or contention caused by the presence of two highly similar pieces of code that are not intended to interact, such as two versions of the same component.
+
+## Requirements
+Side-by-side support implies, in it's simplest form, that an application can consume two versions of the same component with overlapping type names and functionality, and within reasonable bounds, operate normally.
+
+### Goals
+* Enable specific activation of a version of a component without the potential for inadvertently activating an alternate version.
+* Enable developer experiences in projections that maintain a natural and familiar feel
+* To the extent possible, avoid cluttering the developer experience with version information
+* Ensure compatibility with existing tools and projections. Specifically, JavaScript and C++ /CX support, although not under active development, remain important aspects of the Windows Runtime ecosystem. These projections should not be prevented from consuming components designed to support side-by-side versioning.
+* Ensure that actively maintained projections and tooling provide a simple, clean experience.
+
+### Non Goals
+* Automatically support side-by-side for all components. Since side-by-side support requires component authors to implement components in such a manner as to avoid cross-talk, component authors should opt-in only when components guarantee the ability to safely share process space with other instances of the same components. This includes avoiding or versioning use of global resources such as named objects, etc.
+
+## Design
+The design proposed in this specification is a refinement on the model proposed here: https://github.com/microsoft/ProjectReunion/issues/153.
+
+In simplest terms, the design presented here is built around two key aspects:
+1) To ensure compatibility with existing tooling, side-by-side components are marked as such using a Windows Runtime custom attribute (TODO: As annotated on a contract? Type? Where?).
+2) The metadata that describes types intended for side-by-side use contains an extra layer of namespace that expresses the version of the component.
+3) Projections are responsible for masking this versioning information if they understand the version data.
+4) If a projection does not have versioning support, the version appears in the projected namespace used for the component as if it had been explicitly stated. 
+
+### Authoring in MIDL3
+
+The simulated namespace must follow normal identifier grammar for backward compatibility. Grammar for Windows Runtime types is defined here:
+https://docs.microsoft.com/en-us/uwp/winrt-cref/winrt-type-system
+
+MIDL 3 grammar for a versioned component will follow the form:
+
+```
+[component_version(1)]
+component Widgets {
+    namespace Contoso.Widgets {
+        runtime class Doodad;
+    }
+}
+```
+
+The version is specified as a single integral value. The value is approximately equivalent to the major version number in a semver version number. That is, version 1 is not compatible with version 2, etc. For compatible versions, prefer contract versioning. Contract versions, by contrast, are like a semver minor version. Contract version 2 is compatible with contract version 1.
+
+In this example, this is an update to Widgets which introduces non-breaking changes. This would be equivalent to semantic vesion 1.2
+
+```
+[component_version(1)]
+component Widgets {
+
+    [contract(Contoso.Widgets, 2)]
+    namespace Contoso.Widgets {
+
+        [contractversion(2.0)]
+        apicontract WidgetContract{};
+
+        [contract(Contoso.Widgets.WidgetContract, 1)]
+        runtime_class Doodad;
+
+        [contract(Contoso.Widgets.WidgetContract, 2)]
+        runtime_class Thingamabob;
+    }
+}
+```
+
+There is one non-obvious aspect to the example above. Notice that the contract is defined within the Widgets component. This means that the metadata name of the contract is actually Widgets_1.Contoso.Widgets.WidgetContract. MIDLRT will be updated to allow you to omit the contract "namespace" when referencing a contract defined within the same component. Although the names are similar under the covers, and look the same in MIDL3 text, Widgets_1.Contosos.Widgets.WidgetContract has no version relationship with Widgets_2.Contosos.Widgets.WidgetContract. They are two distinct and unrelated contracts.
+
+### Metadata Representation
+
+In the metadata, the MIDL above translates into a custom attribute component_vesion, as well as a namespace path element. To an unenlightened projection the metadata will appear as if the type Doodad contained an additional path element at the root. The MIDL 3 equivalent would be:
+
+```
+namespace Widgets_1.Contoso.Widgets {
+        runtime_class Doodad;
+}
+```
+
+Version 2 of the same component will end up with a different implicit namespace in metadata:
+
+```
+namespace Widgets_1.Contoso.Widgets {
+        runtime_class Doodad;
+}
+```
+
+### ABI & Platform support
+At the ABI, types will self-identify using the component prefix namespace. For example, the IInspectable::GetRuntimeClassName method for Doodad will return the string "Widgets_1.Contoso.Widgets.Doodad". This ensure that metadata resolution algorithms and metadata-based marshalling support in all versions of Windows that support it will be able to uniquely correlate runtime classes with metadata definitions, without risk of type confusion.
+
+The component-qualified type name will also be used in calculations for pinterfaces GUID calculations. This ensures that marshaling and dynamic casting via QueryInterface will not result in type confusion.
+
+Component authors that use explicit GUIDs for interfaces are encouraged, but not required to generate new GUIDs if the binary layout of an interface changes across major versions. 
+
+Type registrations will include the component namespace element in the registered type name. This ensures that activtion of different versions is unambiguous, and that an application can include more than one version of the component. Call to RoGetActivationFactory or RoActivateInstance should use the component-qualified name>
+
+### The C++ Developer Experience
+Developers using older versions of C++ /WinRT will see the qualified namespace name. Developers using enlightened versions of C++ /WinRT will use the components as if the component namespace does not exist. This ensures that only minimal work is required to migrate from version 1 to version 2 of a component. Migration work can focus on breaking changes across versions, rather than mechanical changes required to update references to version-specific names.
+
+Open Design Issues: Should it be a goal to allow a single C++ /WinRT consumer to work with two versions of a component within the same compilation unit?
+
+### The C# Developer Experience
+Clients using .NET Core 3.1 and earlier, as well as .NET Framework 4 will be able to use components, but will refer to types using component-qualified names.
+
+Clients using the .NET 5 projection will not use component-qualified names. Instead, the projections will be differentiated by assembly name.
+
+#### Consuming
+
+Open Design Issue: How do the projection tools and build system handle uniquely named assemblies for different versions of the same component?
+
+#### Authoring
+
+Open Design Issue: How do the projection tools and build system handle uniquely named assemblies for different versions of the same component? Can the single source file authoring experience provide a high-quality experience for implementing components?
+
+### ABITool & MIDLRT header creation support
+and implicitly calculated interface names.
+
+### Design Patterns
+
+#### Cross-version interoperability
+In some cases, interactions between objects in different versions of a componentmay be necessary. In these cases, binary-stable interfaces may still be used. To avoid duplicate definition errors in some projections or languages, these interfaces should still be defined within the component. It's critical that these interfaces retain binary-compatible (ideally identical) definitions, transitively across all types referenced from the interface. Changes to the interface could result in stack or heap corruption.
+
+```
+// original version
+[component_version(2)]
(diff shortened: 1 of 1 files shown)


docs/winrt/WinRT SxS Components.md
```diff
@@ -1 +1,153 @@
...
+#### Authoring
+
+Open Design Issue: How do the projection tools and build system handle uniquely named assemblies for different versions of the same component? Can the single source file authoring experience provide a high-quality experience for implementing components?
+
+### ABITool & MIDLRT header creation support
+and implicitly calculated interface names.
+
+### Design Patterns
+
+#### Cross-version interoperability
```
jonwis, line 129: Another strategy is to define a specific "component" that is about cross-component interop. Like:

```c#
[component_interop]
component WidgetInterop {
    namespace Contoso.Widgets.Interop {
        interface IDoodadInterop {
            Guid Id { get; };
        }
    }
}

[component_version(1)]
component Widgets {
    namespace Contoso.Widgets {
        runtimeclass Widget : Contoso.Widgets.Interop.IDoodadInterop {
            void DoThing();
        }
    }
}

[component_version(2)]
component Widgets {
    namespace Contoso.Widgets {
        runtimeclass Widget : Contoso.Widgets.Interop.IDoodadInterop {
            String GetCake(UInt32 minLayers);
        }
    }
}
```

The rule is that [component_interop] is effectively always version 1 and cannot declare a version 2 of itself.

If you want to prevent inter-component crosstalk, you could add a `[component_interop(WidgetInterop)]` to your `component Widgets` entries, specifically making just those types available.

Or the specific `runtimeclass` gets a new `[interop(WidgetInterop.Contoso.Widgets.IDoodadInterop)]` attribute such that the IDoodadInterop is not directly projected on the type, but projections know to inject this "cloaked" interface for casting purposes.

docs/winrt/WinRT SxS Components.md
```diff
@@ -1 +1,153 @@
...
+
+Clients using the .NET 5 projection will not use component-qualified names. Instead, the projections will be differentiated by assembly name.
+
+#### Consuming
+
+Open Design Issue: How do the projection tools and build system handle uniquely named assemblies for different versions of the same component?
+
+#### Authoring
+
+Open Design Issue: How do the projection tools and build system handle uniquely named assemblies for different versions of the same component? Can the single source file authoring experience provide a high-quality experience for implementing components?
```
jonwis, line 122: Also for authoring, I _think_ your proposal would let C++ developers produce a single DLL that implements multiple versions of the component, which would be pretty keen from a back/forward compat POV.

I could completely break my interface in v2.0 of the component, but still have a single type that effectively implements both the v1.5 and v2.0 interfaces.

docs/winrt/WinRT SxS Components.md
```diff
@@ -1 +1,153 @@
...
+
+Up to this point, this has been largely sufficient. The single largest "component" is the Windows operating system, which has processes in place to ensure binary compatbility. Other components have typically been distributed as part of an MSIX packaged application. he expectation is that the application developer can manage the dependency graph. This has largely been true up to this point. However, authors of new components are recognizing that they will be consumed in more complex dependency graphs, where the application may have dependencies on multiple versions of the component, and the component author does not want to be locked into a binary-stable model. This implies the need for some form of side by side support.
+
+## Terminology
+* Cross-talk - any ambiguity or contention caused by the presence of two highly similar pieces of code that are not intended to interact, such as two versions of the same component.
```
jonwis, line 14: Nit on markdown formatting - I think you need a blank line here for some previewers to work right.

docs/winrt/WinRT SxS Components.md
BenJKuhn: It depends a bit on the details of how C++ allocates "good names". if they're aliased through namespaces and sufficient knobs are provided, it's possible that a single binary or possibly even a single compilation unit could use multiple versions.

docs/winrt/WinRT SxS Components.md
BenJKuhn: avoiding duplicated metadata is still desirable. That will need to be handled with care in this situation. Thinking about this more, it may be more practical for a component author to deliver versioned components with a shared version-stable dependency. If "shared" portions are in defined from multiple components, we may need to teach some tools to de-dup and flag unexpected inconsistencies.

docs/winrt/WinRT SxS Components.md
```diff
@@ -1 +1,153 @@
...
+
+Component authors that use explicit GUIDs for interfaces are encouraged, but not required to generate new GUIDs if the binary layout of an interface changes across major versions. 
+
+Type registrations will include the component namespace element in the registered type name. This ensures that activtion of different versions is unambiguous, and that an application can include more than one version of the component. Call to RoGetActivationFactory or RoActivateInstance should use the component-qualified name>
+
+### The C++ Developer Experience
+Developers using older versions of C++ /WinRT will see the qualified namespace name. Developers using enlightened versions of C++ /WinRT will use the components as if the component namespace does not exist. This ensures that only minimal work is required to migrate from version 1 to version 2 of a component. Migration work can focus on breaking changes across versions, rather than mechanical changes required to update references to version-specific names.
```
BenJKuhn, line 107: Capturing a note from a side conversation. C++ clients should have "safeties" like pragma detect_mismatch to prevent accidental cross-talk between incompatible versions. This could be turned off in build settings for clients that want to explicitly work with both, and do so with appropriate care.

docs/winrt/WinRT SxS Components.md
```diff
@@ -1 +1,153 @@
...
+
+Open Design Issues: Should it be a goal to allow a single C++ /WinRT consumer to work with two versions of a component within the same compilation unit?
+
+### The C# Developer Experience
+Clients using .NET Core 3.1 and earlier, as well as .NET Framework 4 will be able to use components, but will refer to types using component-qualified names.
```
BenJKuhn, line 112: @jeffstall has suggested that it's still desirable to do better than this for .NET Framework and .NET Core developers as well. We'll investigate this further and see what options are available.

docs/winrt/WinRT SxS Components.md
```diff
@@ -1 +1,153 @@
...
+```
+namespace Widgets_1.Contoso.Widgets {
+        runtime_class Doodad;
+}
+```
+
+Version 2 of the same component will end up with a different implicit namespace in metadata:
+
+```
+namespace Widgets_1.Contoso.Widgets {
```
AaronRobinsonMSFT, line 92: Should this be `Widgets_2`?

docs/winrt/WinRT SxS Components.md
```diff
@@ -1 +1,153 @@
...
+        runtime_class Doodad;
+}
+```
+
+### ABI & Platform support
```
AaronRobinsonMSFT, line 97: ~~This ensure~~ -> This will ensure

docs/winrt/WinRT SxS Components.md
```diff
@@ -1 +1,153 @@
...
+At the ABI, types will self-identify using the component prefix namespace. For example, the IInspectable::GetRuntimeClassName method for Doodad will return the string "Widgets_1.Contoso.Widgets.Doodad". This ensure that metadata resolution algorithms and metadata-based marshalling support in all versions of Windows that support it will be able to uniquely correlate runtime classes with metadata definitions, without risk of type confusion.
+
+The component-qualified type name will also be used in calculations for pinterfaces GUID calculations. This ensures that marshaling and dynamic casting via QueryInterface will not result in type confusion.
```
AaronRobinsonMSFT, line 100: pinterfaces?

AaronRobinsonMSFT: ~~marshaling~~ -> marshalling

docs/winrt/WinRT SxS Components.md
```diff
@@ -1 +1,153 @@
...
+
+Component authors that use explicit GUIDs for interfaces are encouraged, but not required to generate new GUIDs if the binary layout of an interface changes across major versions. 
+
+Type registrations will include the component namespace element in the registered type name. This ensures that activtion of different versions is unambiguous, and that an application can include more than one version of the component. Call to RoGetActivationFactory or RoActivateInstance should use the component-qualified name>
```
AaronRobinsonMSFT, line 104: ~~activtion~~ -> activation

AaronRobinsonMSFT: There is a '>' at the end of the line. Is that intended?

docs/winrt/WinRT SxS Components.md
```diff
@@ -1 +1,153 @@
...
+
+Open Design Issue: How do the projection tools and build system handle uniquely named assemblies for different versions of the same component? Can the single source file authoring experience provide a high-quality experience for implementing components?
+
+### ABITool & MIDLRT header creation support
+and implicitly calculated interface names.
+
+### Design Patterns
+
+#### Cross-version interoperability
+In some cases, interactions between objects in different versions of a componentmay be necessary. In these cases, binary-stable interfaces may still be used. To avoid duplicate definition errors in some projections or languages, these interfaces should still be defined within the component. It's critical that these interfaces retain binary-compatible (ideally identical) definitions, transitively across all types referenced from the interface. Changes to the interface could result in stack or heap corruption.
```
AaronRobinsonMSFT, line 130: ~~componentmay~~ -> component may

docs/winrt/WinRT SxS Components.md
```diff
@@ -1 +1,153 @@
-## Placeholder
\ No newline at end of file
+# Side-by-Side Versioned Components
+
+## Status
+Proposed: This document describes a proposed design that has not been implemented.
+
+## Background
+The Windows Runtime type system was largely designed with the needs of opertating system components in mind. Although applications can define their own API types using the Windows Runtime type system, the feature set for doing so is somewhat basic. Versioning strategies are only vaguely defined, and no provisions are made for side-by-side support of a component. As the community has developed more sophisticated libraries over time, these deficiencies are becoming more challenging, and further support is needed.
```
jkotas, line 7:
```suggestion
The Windows Runtime type system was largely designed with the needs of operating system components in mind. Although applications can define their own API types using the Windows Runtime type system, the feature set for doing so is somewhat basic. Versioning strategies are only vaguely defined, and no provisions are made for side-by-side support of a component. As the community has developed more sophisticated libraries over time, these deficiencies are becoming more challenging, and further support is needed.
```

docs/winrt/WinRT SxS Components.md
```diff
@@ -1 +1,153 @@
...
+
+Today, the Windows Runtime only supports binary-compatible versioning. That is, if a an application or component is developed with a dependency on version N of another component, version N+1 *can* be developed in such a way that code designed to use version N can leverage version N+1 without changes or recompilation of the client component. Alternately, a component author can ignore binary compatbility (which comes with trade-offs) and require that applications use a single, specific version of the component.
```
jkotas, line 9:
```suggestion
Today, the Windows Runtime only supports binary-compatible versioning. That is, if a an application or component is developed with a dependency on version N of another component, version N+1 *can* be developed in such a way that code designed to use version N can leverage version N+1 without changes or recompilation of the client component. Alternately, a component author can ignore binary compatibility (which comes with trade-offs) and require that applications use a single, specific version of the component. 
```

docs/winrt/WinRT SxS Components.md
```diff
@@ -1 +1,153 @@
...
+
+### The C++ Developer Experience
+Developers using older versions of C++ /WinRT will see the qualified namespace name. Developers using enlightened versions of C++ /WinRT will use the components as if the component namespace does not exist. This ensures that only minimal work is required to migrate from version 1 to version 2 of a component. Migration work can focus on breaking changes across versions, rather than mechanical changes required to update references to version-specific names.
+
+Open Design Issues: Should it be a goal to allow a single C++ /WinRT consumer to work with two versions of a component within the same compilation unit?
+
+### The C# Developer Experience
+Clients using .NET Core 3.1 and earlier, as well as .NET Framework 4 will be able to use components, but will refer to types using component-qualified names.
+
+Clients using the .NET 5 projection will not use component-qualified names. Instead, the projections will be differentiated by assembly name.
```
jkotas, line 114: It may worth noting that this is different from the default .NET versioning model. Consider:
- Package A references package B v1
- The application project references package A and package B v2
The result is that package A will load B v2 at runtime.

docs/winrt/WinRT SxS Components.md
```diff
@@ -1 +1,153 @@
...
+
+Up to this point, this has been largely sufficient. The single largest "component" is the Windows operating system, which has processes in place to ensure binary compatbility. Other components have typically been distributed as part of an MSIX packaged application. he expectation is that the application developer can manage the dependency graph. This has largely been true up to this point. However, authors of new components are recognizing that they will be consumed in more complex dependency graphs, where the application may have dependencies on multiple versions of the component, and the component author does not want to be locked into a binary-stable model. This implies the need for some form of side by side support.
```
jkotas, line 11:
```suggestion
Up to this point, this has been largely sufficient. The single largest "component" is the Windows operating system, which has processes in place to ensure binary compatibility. Other components have typically been distributed as part of an MSIX packaged application. he expectation is that the application developer can manage the dependency graph. This has largely been true up to this point. However, authors of new components are recognizing that they will be consumed in more complex dependency graphs, where the application may have dependencies on multiple versions of the component, and the component author does not want to be locked into a binary-stable model. This implies the need for some form of side by side support.
```

jkotas, line 11:
```suggestion
Up to this point, this has been largely sufficient. The single largest "component" is the Windows operating system, which has processes in place to ensure binary compatbility. Other components have typically been distributed as part of an MSIX packaged application. The expectation is that the application developer can manage the dependency graph. This has largely been true up to this point. However, authors of new components are recognizing that they will be consumed in more complex dependency graphs, where the application may have dependencies on multiple versions of the component, and the component author does not want to be locked into a binary-stable model. This implies the need for some form of side by side support.
```

jkotas, line 11: The versioning for Windows may be worth drilling into more. It has been a simple: Keep it binary compatible. If it cannot be binary compatible, give it a different name. I do not understand why this simple model does not work as a solution that this proposal is trying to solve.

BenJKuhn: This will remain true for Windows. There are a growing number of Windows Runtime components that are shipped outside of windows as well. Owners of some of those components don't want to reset user codebases to make breaking changes across versions, and are willing to do the work to ensure side-by-side compatibility. This makes it easier to write client code that can easily be reused across versions with little or no changes. If you read carefully between the lines, this effectively is giving everything a new name under the hood, but providing the ability to paper over it for the sake improving code portability and mitigating the cost of version migration.

Binary compatibility offers many, many benefits, but for some scenarios there are good reasons why this may be unavoidable.

docs/winrt/WinRT SxS Components.md
BenJKuhn: I'm using a markdown analyzer (and now a spell checker too), but it didn't flag that. Do you have a favorite that you prefer (ideally for vscode)?

docs/winrt/WinRT SxS Components.md
BenJKuhn: yes, thanks

docs/winrt/WinRT SxS Components.md
BenJKuhn: Fixed

docs/winrt/WinRT SxS Components.md
BenJKuhn: Fixed

docs/winrt/WinRT SxS Components.md
BenJKuhn: Fixed

BenJKuhn: Fixed

docs/winrt/WinRT SxS Components.md
BenJKuhn: Fixed

docs/winrt/WinRT SxS Components.md
BenJKuhn: Not in the US, according to OED: https://www.lexico.com/definition/marshal

My recollection is that Microsoft typically uses the single L form at least within COM docs, but I've been pretty inconsistent. I've standardized on one L.

docs/winrt/WinRT SxS Components.md
BenJKuhn: Fixed

docs/winrt/WinRT SxS Components.md
jkotas: >  If you read carefully between the lines, this effectively is giving everything a new name under the hood

Right, this is giving everything a new name using magic. My point is that you can just tell people to give things new name when they choose to break, and do not build the magic.

docs/winrt/WinRT SxS Components.md
jonwis: For VSCode, I just use "Prettier - Code Formatter" which would have also converted the \* into \-, so YMMV.

docs/winrt/WinRT SxS Components.md
BenJKuhn: I think these are workable suggestions but not necessary to achieve the key requirements this feature is trying to address. The requirement can be met using existing tech, but could be cleaner with better tooling support. Can we table this aspect and treat it as a separate proposal / discussion once this design is settled?

docs/winrt/WinRT SxS Components.md
BenJKuhn: Hmm, this could best be achieved by the component author by enabling the versioned namespace in the projection. C# for .NET 5 would have some challenges with this, I think, because of an optimization for .NET -> .NET calls, but C++ and perhaps Rust could handle this gracefully, as could the older .NET projection, which would still just see these as different types.

PR #698, opened 2020-09-16
Implement Python guid/uuid conversion

### LINKED ISSUE #662: NotImplementedError while using Winrt in Python
Hi,

We are using the Windows.Devices.Bluetooth.GenericAttributeProfile and we are trying to start gatt service for heart rate.
We are able to find the apis which need to be used and calling the api to get the heart rate uuid and facing error like NotImplementedError. Which means in winrt this api functionality is not implemented?, I don't think so the method of calling is wrong. Could you help me get the uuids?

import winrt.windows.devices.bluetooth.genericattributeprofile as wdbgatt
gatt_uuids = wdbgatt.GattServiceUuids
heartrate_uuid = gatt_uuids.get_heart_rate()

o/p:   NotImplementedError

Regards,
Vinoth P B

### LINKED ISSUE #664: new GATT service using the python/winrt Windows.Devices.Bluetooth.GenericAttributeProfile Namespace
Hi,

I am tring to create new gatt service in python/winrt module using the GattServiceProvider Class as per the
document we have to do following steps

CreateAsync(Guid) --> Which will create the new gatt service based one the guid that we are pssing.
Next the from the result of the CreateAsync(guid) we will be having the service_provider using that service provider
we need to perform startadvertising and stopadvertising. Here we have option to pass GattServiceProviderAdvertisingParameters so we can set this boolean varaibles Understood the total concept using the LE explorer application as reference
https://github.com/microsoft/BluetoothLEExplorer/blob/master/GattServicesLibrary/GattServicesLibrary/GenericGattService.cs

Below is my code

import winrt.windows.devices.bluetooth.genericattributeprofile as wdbgatt

def add_advertisement_status_changed(sender, args):

    print('status changed:', args.status)

async def gatt_advertise():

    gatt_service = wdbgatt.GattServiceProvider
    ad_param = wdbgatt.GattServiceProviderAdvertisingParameters()
    ad_param.is_discoverable = True
    ad_param.is_connectable = True
    mfg_uuid = uuid.uuid1()   # This will returns some random guid
    print('uuid:', mfg_uuid)
    result = await gatt_service.create_async(mfg_uuid)
    provider = result.service_provider
    provider.add_advertisement_status_changed(add_advertisement_status_changed)
    provider.start_advertising(ad_param)
    time.sleep(60)  # work around so publish advertisements
    provider.stop_advertising()

When I am trying to run the code receiving runtime error
o/p: RuntimeError: Not implemented

I am following the right procedure but still receiving the Not implemented error.
Could help me to start_advertising based on gatt service?

Regards,
Vinoth P B
--

dlech:
This implements `struct converter<winrt::guid>` to provide marshaling between `xlang::guid` and the standard Python `uuid.UUID` type.

Fixes #662
Fixes #664


dlech:
Hi @BenJKuhn, could you please take a look?

BenJKuhn:
Just curious, how are you building & validating the projection? This project got parked for a bit, and the local dev machine build process was lost. I reconstructed a bit of a hack to get things moving again, but didn't check it in since it wasn't really portable / repeatable. If you have something, or found something, that works well, it would be great if you can share.

src/test/python/testpropertyvalue.py
```diff
@@ -1,6 +1,8 @@
 import sys
 sys.path.append("./generated")
 
+from uuid import UUID
```
BenJKuhn, line 4: If you don't import UUID, can you still import winrt.windows.foundation successfully? Or would it only fail if you use create_guid or other UUID APIs without importing it?

src/tool/python/strings/pybase.h
```diff
@@ -671,6 +671,63 @@ namespace py
         }
     };
 
+    template <>
+    struct converter<winrt::guid>
+    {
+        static PyObject* convert(winrt::guid value) noexcept
+        {
+            PyObject* valueAsBytes = PyBytes_FromStringAndSize((char*)&value, sizeof(value));
+            PyObject* uuidModule = PyImport_ImportModule("uuid");
```
BenJKuhn, line 680: How is the performance of this call in a loop? Is this something that should be lifted our or cached in some way? Could it be static / InitOnceExecuteOnce? Or is it fast enough once loaded that it's not a concern?

dlech:
I buit the project using the scripts in the `src\package\pywinrt\projection\` folder. IIRC, I ran `build.ps1` once and then `generate.ps1` after making changes. I temporarily changed `$pyinclude` in `generate.ps` to get `Windows.Devices.Bluetooth` included without building the full projection.

I couldn't figure out how to get the unit tests to run (gave some sort of DLL import error), but I was able to manually import the generated package Python and ran the new unit test in this PR line by line in a REPL.

src/test/python/testpropertyvalue.py
dlech: The convert function will import the `uuid` module if needed so users don't need to import it.

src/tool/python/strings/pybase.h
dlech: The way I understand it, `PyImport_ImportModule` will import the module if it is not already imported, otherwise it returns immediately. So in a loop, there should be no performance hit.

dlech:
I almost forgot... one tricky part was getting the power shell environment set up:

```PowerShell
> Import-Module '${env:ProgramFiles(x86)}\Microsoft Visual Studio\2019\Community\Common7\Tools\Microsoft.VisualStudio.DevShell.dll'
> Enter-VsDevShell -VsInstallPath '${env:ProgramFiles(x86)}\Microsoft Visual 
Studio\2019\Community\' -DevCmdArguments '-arch=x64'
```

BenJKuhn (approved):

PR #702, opened 2020-09-19
Implementing Embedded Manifest - Draft

kythant:
Draft of embedded Manifest
Very barebones implementation of embedded manifest that looks for an embedded manifest in the exe.

Inside dllmain.cpp, ExtRoLoadCatalog is where the bare minimum "sequencing" is happening 
Currently, it calls LoadFromEmbeddedManifest with the path of the exe. If the embedded manifest is not found, it calls LoadFromSxSManifest which looks for the exename.exe.manifest

There is no functionality for handling dll embedded manifests.

### DIFF
--- a/src/UndockedRegFreeWinRT/UndockedRegFreeWinRT/catalog.cpp
+++ b/src/UndockedRegFreeWinRT/UndockedRegFreeWinRT/catalog.cpp
@@ -82,18 +82,32 @@ struct component
 
 static unordered_map<wstring, shared_ptr<component>> g_types;
 
-HRESULT WinRTLoadComponent(PCWSTR manifest_path)
+HRESULT WinRTLoadComponentFromFilePath(PCWSTR manifestPath)
 {
     ComPtr<IStream> fileStream;
-    ComPtr<IXmlReader> xmlReader;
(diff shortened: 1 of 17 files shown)


src/UndockedRegFreeWinRT/UndockedRegFreeWinRT/catalog.cpp
```diff
@@ -82,18 +82,32 @@ struct component
 
 static unordered_map<wstring, shared_ptr<component>> g_types;
 
-HRESULT WinRTLoadComponent(PCWSTR manifest_path)
+HRESULT WinRTLoadComponentFromFilePath(PCWSTR manifestPath)
 {
     ComPtr<IStream> fileStream;
-    ComPtr<IXmlReader> xmlReader;
+    RETURN_IF_FAILED(SHCreateStreamOnFileEx(manifestPath, STGM_READ, FILE_ATTRIBUTE_NORMAL, false, nullptr, &fileStream));
```
BenJKuhn, line 88: This is a desktop-only API, which will limit where this library can be used. This isn't really intended for UWP scenarios, but is this the first time we would break this compat?
https://docs.microsoft.com/en-us/windows/win32/api/shlwapi/nf-shlwapi-shcreatestreamonfileex

kythant: Discussed offline. We're not introducing a new dependency here and we've always been using xmllite that relies on this so we should be okay.

PR #707, opened 2020-09-30
Embedded Manifest

kythant:
Implemented Embedded Manifests for exe and dll

To enable dll embedded manifests, I call into CreateActCtxW and QueryActCtxW and iterate through the manifests assemblies.

I've also implemented an embedded manifest tests which works exactly like the UndockedRegFreeWinRTTest.exe but the exe manifest is embedded and it has a dependentAssembly on the EmbeddedTestComponent. It looks at the manifest that is embedded in EmbeddedTestComponent.dll and activates the classes from it. 

There are also structural improvements to the project. 

I know this PR is big and the default GitHub may not be sufficient for it so it will be helpful to filter by .cpp extension so you can just look at the important bits.


src/UndockedRegFreeWinRT/UndockedRegFreeWinRT/catalog.cpp
```diff
@@ -81,72 +82,373 @@ struct component
...
 static unordered_map<wstring, shared_ptr<component>> g_types;
+static unordered_set<wstring> g_assemblies;
 
-HRESULT WinRTLoadComponent(PCWSTR manifest_path)
+HRESULT LoadFromSxSManifest(std::wstring path)
+{
+    return WinRTLoadComponentFromFilePath(path.c_str());
+}
+
+HRESULT LoadFromEmbeddedManifest(std::wstring path)
```
jonwis, line 92: There's a feature of CreateActCtxW that you can point it at a DLL or exe and it'll compute the activation context for it, including all the probing logic listed here.  Then there are APIs like https://docs.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-queryactctxw and https://docs.microsoft.com/en-us/windows/win32/api/winnt/ns-winnt-activation_context_assembly_detailed_information that let you find the list of assembly manifests used to construct the activation context, which you can then load & parse.

Activation contexts are also cacheable so you could keep a list of those somewhere.

src/UndockedRegFreeWinRT/UndockedRegFreeWinRT/catalog.cpp
```diff
@@ -81,72 +82,373 @@ struct component
...
+        return ERROR_FILE_NOT_FOUND;
+    }
+    DWORD length = SizeofResource(handle, hrsc);
+    void* data = LockResource(embeddedManifest);
+    if (!data)
+    {
+        return ERROR_FILE_NOT_FOUND;
+    }
+    std::string result = std::string((char*)data, length);
+    return WinRTLoadComponentFromString(result.c_str());
```
jonwis, line 137: Maybe instead of copying the data here, WinRTLoadComponentFromString should take a `string_view` ?

src/UndockedRegFreeWinRT/UndockedRegFreeWinRT/catalog.cpp
```diff
@@ -81,72 +82,373 @@ struct component
...
+    {
+        // I notice in testing, the original sxs allows this number to be also 1. 
+        resource = 2;
+    }
+    else
+    {
+        return COR_E_ARGUMENT;
+    }
+
+    HMODULE handle = LoadLibraryExW(path.c_str(), nullptr, LOAD_LIBRARY_AS_DATAFILE_EXCLUSIVE);
```
jonwis, line 114: Use WIL's `unique_hmodule` to not leak this HMODULE.

src/UndockedRegFreeWinRT/UndockedRegFreeWinRT/catalog.cpp
```diff
@@ -81,72 +82,373 @@ struct component
...
     ComPtr<IStream> fileStream;
-    ComPtr<IXmlReader> xmlReader;
+    RETURN_IF_FAILED(SHCreateStreamOnFileEx(manifestPath, STGM_READ, FILE_ATTRIBUTE_NORMAL, false, nullptr, &fileStream));
+    return ParseRootManifestFromXmlReaderInput(fileStream.Get());;
+}
+
+HRESULT WinRTLoadComponentFromString(PCSTR xmlStringValue)
+{
+    ComPtr<IStream> xmlStream = nullptr;
+    xmlStream.Attach(SHCreateMemStream(reinterpret_cast<const BYTE*>(xmlStringValue), strlen(xmlStringValue) * sizeof(CHAR)));
```
jonwis, line 150: Does this make another copy of the incoming data? I can't quite parse what MSDN's saying.

src/UndockedRegFreeWinRT/UndockedRegFreeWinRT/catalog.cpp
```diff
@@ -81,72 +82,373 @@ struct component
...
+}
+
+HRESULT ParseFileTag(ComPtr<IXmlReader> xmlReader)
+{
+    HRESULT hr = S_OK;
+    XmlNodeType nodeType;
+    LPCWSTR localName = nullptr;
+    LPCWSTR fileName = nullptr;
+    hr = xmlReader->MoveToAttributeByName(L"name", nullptr);
+    if (hr != S_OK)
```
jonwis, line 198: `RETURN_HR_IF` ?

src/UndockedRegFreeWinRT/UndockedRegFreeWinRT/catalog.cpp
```diff
@@ -81,72 +82,373 @@ struct component
...
+            const WCHAR* pwszLocalName;
+            const WCHAR* pwszValue;
+            if (FAILED(hr = xmlReader->GetLocalName(&pwszLocalName, NULL)))
+            {
+                wprintf(L"Error getting local name, error is %08.8lx", hr);
+                return HRESULT_FROM_WIN32(ERROR_SXS_MANIFEST_PARSE_ERROR);
+            }
+            if (FAILED(hr = xmlReader->GetValue(&pwszValue, NULL)))
+            {
+                wprintf(L"Error getting value, error is %08.8lx", hr);
```
jonwis, line 258: Should these wprintfs be fprintfs to stderr?

src/UndockedRegFreeWinRT/UndockedRegFreeWinRT/catalog.cpp
```diff
@@ -81,72 +82,373 @@ struct component
...
+    {
+        return HRESULT_FROM_WIN32(ERROR_SXS_MANIFEST_PARSE_ERROR);
+    }
+    else
+    {
+        while (TRUE)
+        {
+            const WCHAR* pwszLocalName;
+            const WCHAR* pwszValue;
+            if (FAILED(hr = xmlReader->GetLocalName(&pwszLocalName, NULL)))
```
jonwis, line 251: Consider using `if (FAILED_LOG(...))` which integrates with WIL's error path so you don't need the wprintfs at all.

src/UndockedRegFreeWinRT/UndockedRegFreeWinRT/catalog.cpp
```diff
@@ -81,72 +82,373 @@ struct component
...
+    auto component_iter = g_types.find(activatableClass);
+    if (component_iter != g_types.end()) 
+    {
+        return HRESULT_FROM_WIN32(ERROR_SXS_DUPLICATE_ACTIVATABLE_CLASS);
+    }
+    g_types[activatableClass] = this_component;
+    return S_OK;
+}
 
+HRESULT ParseDependentAssemblyTag(ComPtr<IXmlReader> xmlReader)
```
jonwis, line 316: Should these methods be void and use throwing instead of "return error" ? Or should there be a try/catch at the top level since you're mixing C++ constructs and non-throwing constructs?

src/UndockedRegFreeWinRT/UndockedRegFreeWinRT/catalog.cpp
```diff
@@ -81,72 +82,373 @@ struct component
...
+{
+    XmlNodeType nodeType;
+    LPCWSTR localName = nullptr;
```
jonwis, line 319: `PCWSTR`.  Also, for consistently, pick either `const WCHAR*` or `PCWSTR`. You've got a mix of both in this change.

src/UndockedRegFreeWinRT/UndockedRegFreeWinRT/catalog.cpp
```diff
@@ -81,72 +82,373 @@ struct component
...
+        {
+            RETURN_IF_FAILED((xmlReader->GetLocalName(&localName, nullptr)));
+            if (_wcsicmp_l(localName, L"assemblyIdentity", locale) == 0)
+            {
+                RETURN_IF_FAILED(ParseAssemblyIdentityTag(xmlReader, true));
+            }
+        }
+        else if (nodeType == XmlNodeType_EndElement)
+        {
+            RETURN_IF_FAILED((xmlReader->GetLocalName(&localName, nullptr)));
```
jonwis, line 333: Extra parens

src/UndockedRegFreeWinRT/UndockedRegFreeWinRT/catalog.cpp
```diff
@@ -81,72 +82,373 @@ struct component
...
+    {
+        // Don't fail if we don't find the assembly, in the realworld, manifests may have 
+        // dependentAssembly ref to WinSxS assemblies which this code won't find.
+        SearchAssembly(dependentAssemblyFileName);
+    }
+    return hr;
+}
+
+
+HRESULT SearchAssembly(std::wstring dependentAssemblyFileName)
```
jonwis, line 374: wstring_view ? Or `PCWSTR` since that's mostly what you're calling it woth?

src/UndockedRegFreeWinRT/UndockedRegFreeWinRT/catalog.cpp
```diff
@@ -81,72 +82,373 @@ struct component
...
+    }
+    if (hr == COR_E_FILENOTFOUND || hr == HRESULT_FROM_WIN32(ERROR_PATH_NOT_FOUND))
+    {
+        hr = SearchAssemblyWithLang(dependentAssemblyFileName, lang.substr(0, 2));
+        if (hr == S_OK)
+        {
+            return hr;
+        }
+    }
+    // This wasn't clear in the docs but it seems to fallback to en-US first
```
jonwis, line 398: It falls back through the languages the user has installed. Please don't hardcode en-US here. Another reason to use CreateActCtxW which already follows the right pattern.

src/UndockedRegFreeWinRT/UndockedRegFreeWinRT/catalog.cpp
```diff
@@ -81,72 +82,373 @@ struct component
...
+            {
+                return hr;
+            }
+        }
+    }
+    // Finally search for a no language version
+    return SearchAssemblyWithLang(dependentAssemblyFileName, L"");
+}
+
+HRESULT SearchAssemblyWithLang(std::wstring dependentAssemblyFileName, std::wstring lang)
```
jonwis, line 419: `std::wstring const&` - the caller already has a `std::wstring` and passing by-value induces an unnecessary copy.

src/UndockedRegFreeWinRT/UndockedRegFreeWinRT/catalog.cpp
```diff
@@ -81,72 +82,373 @@ struct component
...
+{
+    if (!lang.empty())
+    {
+        lang += L"\\";
+    }
+    HRESULT hr = S_OK;
+    PCWSTR exeFilePath = nullptr;
+    UndockedRegFreeWinRT::GetProcessExeDir(&exeFilePath);
+    std::wstring path = std::wstring(exeFilePath) + L"\\" + lang;
```
jonwis, line 428: Should this use the `std::` path types instead of hand-munging them?

jonwis (changes requested):

src/UndockedRegFreeWinRT/UndockedRegFreeWinRT/catalog.cpp
kythant: Deleted this method

src/UndockedRegFreeWinRT/UndockedRegFreeWinRT/catalog.cpp
kythant: I am returning errors such as ERROR_SXS_MANIFEST_PARSE_ERROR and ERROR_SXS_DUPLICATE_ACTIVATABLE_CLASS so I don't really want it to be void. Maybe just a try catch at the top level.

src/UndockedRegFreeWinRT/UndockedRegFreeWinRT/catalog.cpp
kythant: You're talking about SHCreateMemStream right? Yea the MSDN isn't so clear about it.

src/UndockedRegFreeWinRT/UndockedRegFreeWinRT/catalog.cpp
kythant: Wasn't sure how to use unique_hmodule with LoadLibraryExW so I just use a wil::scope_exit

src/UndockedRegFreeWinRT/UndockedRegFreeWinRT/dllmain.cpp
```diff
@@ -352,13 +352,91 @@ void RemoveHooks()
...
+        return HRESULT_FROM_WIN32(GetLastError());
+    }
     std::wstring manifestPath(filePath);
-    manifestPath += L".manifest";
+    HANDLE hActCtx = INVALID_HANDLE_VALUE;
+    auto exit = wil::scope_exit([&]
+    {
+        if (hActCtx != INVALID_HANDLE_VALUE)
+        {
+            ReleaseActCtx(hActCtx);
```
jonwis, line 365: Consider adding a wil::unique_actctx maybe?

src/UndockedRegFreeWinRT/UndockedRegFreeWinRT/dllmain.cpp
```diff
@@ -352,13 +352,91 @@ void RemoveHooks()
...
+    });
+    wil::unique_hmodule exeModule;
+    RETURN_IF_WIN32_BOOL_FALSE(GetModuleHandleExW(0, nullptr, &exeModule));
+    ACTCTXW acw = { sizeof(acw) };
+    acw.lpSource = manifestPath.c_str();
+    acw.hModule = exeModule.get();
+    acw.lpResourceName = MAKEINTRESOURCEW(1);
+    acw.dwFlags = ACTCTX_FLAG_HMODULE_VALID | ACTCTX_FLAG_RESOURCE_NAME_VALID;
+
+    hActCtx = CreateActCtxW(&acw);
```
jonwis, line 376: `RETURN_LAST_ERROR_IF(!hActCtx)`

src/UndockedRegFreeWinRT/UndockedRegFreeWinRT/dllmain.cpp
```diff
@@ -352,13 +352,91 @@ void RemoveHooks()
...
+    (void)QueryActCtxW(0,
+        hActCtx,
+        nullptr,
+        ActivationContextDetailedInformation,
+        nullptr,
+        0,
+        &bufferSize);
+    RETURN_HR_IF(HRESULT_FROM_WIN32(GetLastError()), bufferSize == 0);
+
+    actCtxInfo = (PACTIVATION_CONTEXT_DETAILED_INFORMATION) new BYTE[bufferSize];
```
jonwis, line 394: Maybe use std::vector or `auto buffer = make_unique<BYTE[]>(bufferSize);` and then use `reinterpret_cast<PACTCTXDETAILEDINFO>(buffer.get())`

jonwis: (and if you do so, wrap the function with a function-level try to catch any failures.)

src/UndockedRegFreeWinRT/UndockedRegFreeWinRT/catalog.cpp
kythant: Should I go with this solution?

    try
    {
        return ParseRootManifestFromXmlReaderInput(fileStream.Get());
    }
    catch(...)
    {
        LOG_CAUGHT_EXCEPTION();
        return ERROR_SXS_MANIFEST_PARSE_ERROR;
    }

src/UndockedRegFreeWinRT/UndockedRegFreeWinRT/dllmain.cpp
```diff
@@ -352,13 +352,91 @@ void RemoveHooks()
...
+        (void)QueryActCtxW(0,
+            hActCtx,
+            &index,
+            AssemblyDetailedInformationInActivationContext,
+            nullptr,
+            0,
+            &bufferSize);
+        RETURN_HR_IF(HRESULT_FROM_WIN32(GetLastError()), bufferSize == 0);
+
+        asmInfo = (PACTIVATION_CONTEXT_ASSEMBLY_DETAILED_INFORMATION) new BYTE[bufferSize];
```
jonwis, line 422: Same comment as above. Maybe make a helper method, like:

```c++
template<typename T> wistd::unique_ptr<T> make_sized_struct(size_t size) {
    wistd::unique_ptr<T> buffer(reinterpret_cast<T*>(::operator new (size)));
}
```

... or something

jonwis (approved):
Thanks!

EricJohnson327 (approved):

Issue #709, opened 2020-10-08
Build and release winrt package for Python 3.9


dlech:
Python 3.9 is nearing the stable release. It would be nice to have a new release built for this in the near future.

There are also a few outstanding pull requests that would be nice to be merged before releasing.

Also, it would be nice to release for multiple Python versions so that users get the same package version no matter which version of Python they are running.

The current `setup.py` lists Python 3.7, 3.8 and 3.9 however the current release on PyPi is only for Python 3.8 and previous releases are only for Python 3.7.

BenJKuhn:
https://github.com/actions/virtual-environments/issues/1740

BenJKuhn:
Just posting an update. It sounds like ADO will have updated Python 3.9 support enabled early next week (based on comments in the thread linked above), so I should be able to publish the new packages shortly after that.

Thanks,

Ben

dany74q:
Hey @BenJKuhn ! Any chance we have an ETA on having this awesome lib on 3.9 ?

Thanks !

dany74q:
`winrt 1.0.20330.1` is now available for python 3.7 - 3.9, that's awesome, thanks @BenJKuhn !

Kyle-Payne:
Is there an ETA on when winrt will be updated for Python 3.10?

dlech:
The project has moved. Please use https://pypi.org/project/winsdk/ instead.

Kyle-Payne:
Dlech, ok, so if an application like gimme-aws-creds is using winrt then it needs to be updated to use winsdk instead. Correct?

dlech:
yes.

PR #710, opened 2020-10-08
build and package for Py 3.7, 3.8, 3.9

dlech:
Not sure if this is the right way to do it, but this hopefully addresses #709.

### DIFF
--- a/src/package/pywinrt/projection/AzurePipelinesTemplates/job-build-projection.yml
+++ b/src/package/pywinrt/projection/AzurePipelinesTemplates/job-build-projection.yml
@@ -7,15 +7,33 @@ jobs:
   strategy:
     maxParallel: 10
     matrix:
-      Release_x86:
+      Release_x86_py37:
         buildArchitecture: 'x86'
         buildConfiguration: 'Release'
-      Release_x64:
+        pythonVersion: '3.7.x'
+      Release_x64_py37:
         buildArchitecture: 'amd64'
         buildConfiguration: 'Release'
+        pythonVersion: '3.7.x'
+      Release_x86_py38:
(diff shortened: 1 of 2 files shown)


BenJKuhn:
I have a feeling this is going to take just a bit more work to get building / working smoothly. Let me get a 3.9 package out this week (which will include your other contributions) then see if I can work through this to produce updates packages for 3.7+ with all the changes.

BenJKuhn:
the change as suggested here will create valid wheels that will install in the various versions of python, but they're unlikely to load or run correctly. Basically, the python interface is not compatible across versions, and the linked DLL name changes with each version. This change will build versioned packages, but put the same binaries in each one.

You can read more here:
https://docs.python.org/3/c-api/stable.html

I'm not sure that I can change the version that I use on the build agent on the fly, so I'm going to change around the pipeline definition to allow the version to be controlled when the pipeline is run. Then I can just trigger runs for any source-compatible versions that we continue to support, one per version. Note that this only applies to the projection, not the tool itself.

BenJKuhn:
I made significant progress here:
https://github.com/microsoft/xlang/tree/benkuhn/version-enhancements

It looks like the task UsePythonVersion@0 in steps-build-projection.yml isn't running (and hasn't been for a long time, we just got lucky on versions). The conditional doesn't seem to be working, but I have confirmed the parameter is set right. Enough debugging for tonight though. I'll try to get back to this later this weekend.

BenJKuhn:
@dlech I have something that seems to be working so far. It's loosely based on what you started here, but ensures that the published artifacts don't step on each other. I'm going to close out this PR, and start a fresh one from my working branch. I'll add you to that.

PR #712, opened 2020-10-10
Updating build to target python 3.9

### LINKED ISSUE #709: Build and release winrt package for Python 3.9

BenJKuhn:
This is a minimal update needed to produce a PyPi package for Python 3.9, which just released a few days ago.
resolves #709

### DIFF
--- a/src/package/pywinrt/projection/AzurePipelinesTemplates/job-build-projection.yml
+++ b/src/package/pywinrt/projection/AzurePipelinesTemplates/job-build-projection.yml
@@ -17,5 +17,5 @@ jobs:
   - template: steps-build-projection.yml
     parameters:
       architecture: $(buildArchitecture)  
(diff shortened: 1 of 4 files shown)


kennykerr (approved):

jonwis (approved):
Looks OK; at some point we should figure out whether we need to support building 3.8 and 3.9, depending on how long they're in market together.

PR #713, opened 2020-10-14
Fix a bug with upper case DLL extension and added Embedded Manifest Test in c#

kythant:
Fix a bug with upper case DLL extension and added Embedded Manifest Test in c#.

The code looks at the extension of the path from the asmInfo but it only looks for ".dll" but the extension was in uppercase.

### DIFF
--- a/src/UndockedRegFreeWinRT/UndockedRegFreeWinRT/catalog.cpp
+++ b/src/UndockedRegFreeWinRT/UndockedRegFreeWinRT/catalog.cpp
@@ -14,6 +14,7 @@
 #include <codecvt>
 #include <locale>
 #include <RoMetadataApi.h>
+#include <algorithm>
 #include "catalog.h"
 #include "TypeResolution.h"
 
@@ -82,6 +83,13 @@ struct component
 
 static unordered_map<wstring, shared_ptr<component>> g_types;
 
+std::wstring_view wstr_tolower(std::wstring& s) {
+    std::transform(s.begin(), s.end(), s.begin(),
+        [](unsigned char c) { return std::tolower(c); } // correct
+    );
+    return s;
+}
+
 HRESULT LoadManifestFromPath(std::wstring path)
 {
     if (path.size() < 4)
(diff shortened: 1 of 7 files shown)


src/UndockedRegFreeWinRT/UndockedRegFreeWinRT/catalog.cpp
```diff
@@ -82,13 +83,21 @@ struct component
 
 static unordered_map<wstring, shared_ptr<component>> g_types;
 
+std::wstring_view wstr_tolower(std::wstring& s) {
```
DefaultRyan, line 86: This doesn't seem like the right design (accept a non-const `wstring`, modify it in-place, and return a `wstring_view` to the arg. Either return void, or return the `wstring`.

BenJKuhn (approved):

src/UndockedRegFreeWinRT/UndockedRegFreeWinRT/catalog.cpp
DefaultRyan: Better yet, since this is only going to be used for the extension, just use it to return a new string containing just the extension (like everything from the '.' to the end).

src/UndockedRegFreeWinRT/UndockedRegFreeWinRT/catalog.cpp
```diff
@@ -82,13 +83,21 @@ struct component
...
+    std::transform(s.begin(), s.end(), s.begin(),
+        [](unsigned char c) { return std::tolower(c); } // correct
```
DefaultRyan, line 88: This will break with Unicode filenames, consider using FindStringOrdinal instead. But since we're only looking at the extension, it's probably OK.

Just a little bit of fragility that you're converting the entire string, instead of just extracting the extension. If you change your function from converting the entire string, and instead, extract and `tolower` only the extension, then that's much better, since you know the extension should *only* be '.dll' or '.exe', which won't get corrupted by `tolower`.

DefaultRyan (changes requested):

src/UndockedRegFreeWinRT/UndockedRegFreeWinRT/catalog.cpp
kythant: I don't think using the string returned substr() of path means I am passing in the entire string is it? It should still only be the extension. Yea I can remake this method into a more specific GetExtensionLowerCase that takes in the full path and returns a new string with just the extension so it doesn't get misused. 

In the next PR :)

PR #715, opened 2020-10-26
Change pywinrt to build & package wheels for all supported python versions in one build operation.

BenJKuhn:
@dlech, here's the updated PR with the full set of changes. I'm still waiting for a full run end-to-end, and will need to check the published artifacts after that. Assuming that all looks good, this will build version 3.7 and 3.8 packages, and I can stage up 3.9 after this goes in. It should be far more straightforward with this change (just two files to touch). Not 100% optimized, but significantly better.


BenJKuhn:
Test build ran. pythonVersionSpec didn't expand in the artifact directories, and signing isn't pointing to the right place anymore, so need to fix those up before this is ready to go.

dlech:
Looks good to me, thanks.

sotteson1 (approved):

PR #724, opened 2021-02-02
Fix python version reference from native code DLLs

### LINKED ISSUE #717: Python/WinRT import error
```
In [1]: import winrt
---------------------------------------------------------------------------
ImportError                               Traceback (most recent call last)
<ipython-input-1-0eba3d78e1ae> in <module>
----> 1 import winrt

c:\program files\python38\lib\site-packages\winrt\__init__.py in <module>
      1 # WARNING: Please don't edit this file. It was generated by Python/WinRT v0.9.201125.1
      2
----> 3 from . import _winrt
      4 _winrt.init_apartment()
      5

ImportError: DLL load failed while importing _winrt: 找不到指定的模块。
```
Python Version: `Python 3.8.2 (tags/v3.8.2:7b3ab59, Feb 25 2020, 23:03:10) [MSC v.1916 64 bit (AMD64)]`
Python/WinRT Version: `winrt-1.0.20330.1-cp38-none-win_amd64.whl`

BenJKuhn:
CMake selection of python version uses "latest" that it can find by default, and isn't impacted by the pipeline task that configures the preferred python version. This changes the CMake build of the native code DLL to use specific versions for the binaries.

fixes #717

### DIFF
--- a/src/package/pywinrt/projection/AzurePipelinesTemplates/job-build-projection.yml
+++ b/src/package/pywinrt/projection/AzurePipelinesTemplates/job-build-projection.yml
@@ -10,32 +10,32 @@ jobs:
       Release_x86_py37:
         buildArchitecture: 'x86'
         buildConfiguration: 'Release'
-        pythonVersion: '3.7.x'
+        pythonVersion: '3.7'
         pythonArchitecture: 'x86'
       Release_x64_py37:
         buildArchitecture: 'amd64'
         buildConfiguration: 'Release'
-        pythonVersion: '3.7.x'
+        pythonVersion: '3.7'
         pythonArchitecture: 'x64'
       Release_x86_py38:
         buildArchitecture: 'x86'
         buildConfiguration: 'Release'
-        pythonVersion: '3.8.x'
+        pythonVersion: '3.8'
         pythonArchitecture: 'x86'
       Release_x64_py38:
         buildArchitecture: 'amd64'
         buildConfiguration: 'Release'
-        pythonVersion: '3.8.x'
+        pythonVersion: '3.8'
         pythonArchitecture: 'x64'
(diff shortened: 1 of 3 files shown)


src/package/pywinrt/projection/AzurePipelinesTemplates/job-build-projection.yml
```diff
@@ -10,32 +10,32 @@ jobs:
       Release_x86_py37:
         buildArchitecture: 'x86'
         buildConfiguration: 'Release'
-        pythonVersion: '3.7.x'
```
DefaultRyan, line 13: It's not clear to me why these were changed, but then again, I'm not an expert in this space.

What effect does this have, e.g., finding version 3.7.0 vs 3.7.1? I'm assuming you've tested it and it works as expected. Just trying to educate myself.

DefaultRyan (approved):

src/package/pywinrt/projection/AzurePipelinesTemplates/job-build-projection.yml
BenJKuhn: find_package in cmake doesn't use the .x syntax, but the YAML task requires .x. So I removed it from the variable value and added it back in where it's needed in the steps file.

BenJKuhn: Also, in find_package, "3.7 EXACT" will match 3.7.*

DefaultRyan: Yeah, the find_package EXACT bit was the one I was wondering about. I suspected it worked that way, but I feel unsafe making assumptions about CMake. :)

PR #743, opened 2021-06-06
python: implement buffer protocol for IBuffer

dlech:
This implements the CPython buffer protocol<sup>[1]</sup> for the `Windows.Storage.Streams.IBuffer` interface. This allows Python to access the buffer memory directly via `memoryview()`, the `struct` module and so on.

This is also a significant performance improvement for applications that just need to copy the data to/from the buffer.

[1]: https://docs.python.org/3/c-api/buffer.html

### DIFF
--- a/src/package/pywinrt/projection/readme.md
+++ b/src/package/pywinrt/projection/readme.md
@@ -109,6 +109,17 @@ async def get_current_latitude():
     return pos.coordinate.latitude
 ```
 
+### Buffer Protocol
+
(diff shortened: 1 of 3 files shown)


hcvarekamp:
Hey, is this still open? Are there any workarounds? I can´t figure out how to read an IBuffer in python. Thanks.

dlech:
I started a community fork at https://github.com/pywinrt that includes this fix and many, many others. Just replace the `winrt` package with `winsdk` and you will get that version.

PR #744, opened 2021-06-06
python: bump sdk version to 18362

### LINKED ISSUE #727: Missing classes and properties which were added in newer SDKs
When trying to use the Python module `winrt` to create toast notifications and handle their events, I noticed that the class `ToastActivatedEventArgs` is missing the `UserInput` property which was added in SDK version 18362.

Investigating a bit further, other classes and properties which where added in 18362 are also missing, for example `Windows.Devices.Input.PenDevice`.

Which SDK was used to build winrt? And would it be possible to update it?

dlech:
This is Windows 10 1903, which gives us 2 years worth of backwards compatibility.

Fixes #727

### DIFF
--- a/src/package/pywinrt/projection/AzurePipelinesTemplates/steps-build-projection.yml
+++ b/src/package/pywinrt/projection/AzurePipelinesTemplates/steps-build-projection.yml
@@ -22,9 +22,9 @@ steps:
 - script: |
     echo %ProjectionType%
     IF %ProjectionType%==full (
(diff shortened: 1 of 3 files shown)


jonwis (approved):

BenJKuhn (approved):

SamSchott:
Fancy cutting a new release with this?

dlech:
FYI, I have made some builds that include this plus a bunch of other bug fixes at https://github.com/dlech/xlang/actions/runs/1305498290.

PR #747, opened 2021-06-08
python: add event handler example to README

dlech:
I didn't see this explained anywhere, so this is based on just trying it out and seeing how it works.

### DIFF
--- a/src/package/pywinrt/projection/readme.md
+++ b/src/package/pywinrt/projection/readme.md
@@ -109,6 +109,87 @@ async def get_current_latitude():
     return pos.coordinate.latitude
 ```
 
+### Event Handlers
+
+Events in the Python projection are exposed as an `add_<event_name>` method
+for adding a handler and `remove_<event_name>` for removing a handler. The
+`add_<event_name>` methods return an event token object that must be passed to
+the corresponding `remove_<event_name>` method.
+
+Event callbacks will be called on a background WinRT thread. Consider using
+[call_soon_threadsafe] to ensure the handler is synchronized with the event loop.
+
+[call_soon_threadsafe]: https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.loop.call_soon_threadsafe
+
+```python
+import asyncio
+
+from winrt.windows.devices.bluetooth.advertisement import BluetoothLEAdvertisementWatcher
+
+
+async def scan():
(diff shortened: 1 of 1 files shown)


src/package/pywinrt/projection/readme.md
```diff
@@ -109,6 +109,87 @@ async def get_current_latitude():
...
+      received_queue.put_nowait(event_args)
+
+    stopped_future = event_loop.create_future()
+
+    # this event is expected *exactly* once, so we use a future to capture the result
+    def handle_stopped(sender, event_args):
+      stopped_future.set_result(event_args)
+
+    received_token = watcher.add_received(
+        lambda s, e: event_loop.call_soon_threadsafe(handle_received, s, e)
```
jonwis, line 147: Does this `call_soon_threadsafe` block when invoked? Some WinRT APIs expect the caller to "do something" with the `-EventArgs` object before returning.  For that we have https://docs.microsoft.com/en-us/uwp/api/Windows.Foundation.Deferral?view=winrt-20348 which lets a runtime tell the invoker "hey I'm not done handling this event yet, I'll let you know when I am."  Callers don't _always_ want to take out deferrals, but sometimes it's super important. Calling `Deferral.Complete()` ends the "logical" scope of the event invocation.

For a C++/WinRT example:

```c++
thing.PairingRequested([&](const auto& sender, DevicePairingRequestedEventArgs args)
{
    // Assume this is coming in on some background system thread generated by the device
    auto deferral = args.GetDeferral(); // Tell the sender we need a minute
    if (args.PairingKind() == DevicePairingKinds::ConfirmPinMatch)
    {
        // Here the event handler physically returns and the coroutine is suspended
        winrt::hstring userPin = co_await RequestPinFromUserOnUxThread();
        // Ok, we're back, the UX thread has given us a string pin
        args.Accept(userPin);
    }
    deferral.Complete(); // Inform the sender of the event we're really done with the event
});
```

jonwis (approved):

src/package/pywinrt/projection/readme.md
dlech: > Does this `call_soon_threadsafe` block when invoked?

No, it only schedules the callback, it does not wait for completion.

Thanks for the tip about Deferrals. Sounds like a good example to add eventually.

PR #751, opened 2021-06-26
python: fix leak of object pointers

### LINKED ISSUE #731: Python/WinRT memory leaks
I noticed that using `windows.media.control` can often lead to memory leaks, seemingly uncontrollable from python's side of things.

Example demonstration:
```python
from winrt.windows.media.control import GlobalSystemMediaTransportControlsSessionManager as MediaManager
from pympler.tracker import SummaryTracker
import asyncio
import gc


async def get_media_info():
    # Memory leaking calls
    sessions = await MediaManager.request_async()
    current_session = sessions.get_current_session()
    await current_session.try_get_media_properties_async()

    # Force garbage collect, does nothing
    gc.collect()

    return


if __name__ == '__main__':

    # Pause to check starting memory usage (I used task manager)
    # ~ 30 MB in my case
    input()
    
    # Track objects and sizes
    tracker = SummaryTracker()

    # 500 iterations is a good example amount
    for _ in range(500):
        # Leak memory
        asyncio.run(get_media_info())

        # Force garbage collect, does nothing
        gc.collect()

    # Print info on object count and size differences
    # Will always show ~ 1 MB of leaks, no matter how many iterations
    tracker.print_diff()

    # Pause to check ending memory usage (I used task manager)
    # ~ 60 MB with 500 iterations
    # ~ 330 MB with 5000 iterations
    # ~ 3.2 GB with 50000 iterations
    input()
```

Both `tracemalloc` and `pympler` do not detect these leaks, and python's built in garbage collector `gc` does not dispose of these items

dlech:
When a WinRT method returns a pointer, the python bindings were leaking a reference.

This solves it by deleting the pointer returned by `std::unique_ptr::release()`.

Basically, this patch changes the generated code from this:

```cpp
    static void _dealloc_IAsyncOperation(py::wrapper::Windows::Foundation::IAsyncOperation* self)
    {
        auto hash_value = self->obj->hash();
        py::wrapped_instance(hash_value, nullptr);
        self->obj.release();
    }
```

To this (just adds `delete`):

```cpp
    static void _dealloc_IAsyncOperation(py::wrapper::Windows::Foundation::IAsyncOperation* self)
    {
        auto hash_value = self->obj->hash();
        py::wrapped_instance(hash_value, nullptr);
        delete self->obj.release();
    }
```

I know enough about C++ to know this fix looks wrong but I don't know enough to know how to fix it properly.

For reference, `self->obj` is initialized here: 
https://github.com/microsoft/xlang/blob/b950b15d53cab2dddc78c65e783ef43042a2e8c9/src/tool/python/strings/pybase.h#L318-L350

Fixes #731

### DIFF
--- a/src/tool/python/code_writers.h
+++ b/src/tool/python/code_writers.h
@@ -325,7 +325,7 @@ self->obj%;
 
             w.write(format,
                 is_ptype(type) ? "self->obj->hash()" : "std::hash<winrt::Windows::Foundation::IInspectable>{}(self->obj)",
-                is_ptype(type) ? ".release()" : " = nullptr");
+                is_ptype(type) ? ".reset()" : " = nullptr");
         }
         w.write("}\n");
     }


jonwis:
The C++ thing would be to use `.reset()` on the `unique_ptr<T>`. That keeps the "lifeline" of the pointer within the control of the `unique_ptr<T>`.  It's _ok_ to use `delete p.release()` but it's more idiomatic to say `p.reset()` or `p = nullptr`.

dlech:
Thanks! I knew there had to be a nicer way. 😄

dlech:
I have updated the change so that it now generates code like this:

```cpp
static void _dealloc_IAsyncOperation(py::wrapper::Windows::Foundation::IAsyncOperation* self)
    {
        auto hash_value = self->obj->hash();
        py::wrapped_instance(hash_value, nullptr);
        delete self->obj.reset();
    }
```

PR #766, opened 2021-12-09
Fix a bug where the filename variable can get clobbered due to XmlReader invalidating the buffer

JesseCol:
The buffer returned by IXmlReader::GetValue gets invalidated after the XmlReader is used more in the future.  The fileName buffer was getting re-used in a scenario where there were lots of nodes to read.  The fix is simply to save it off to a local wstring.

### DIFF
--- a/src/UndockedRegFreeWinRT/UndockedRegFreeWinRT/dllmain.cpp
+++ b/src/UndockedRegFreeWinRT/UndockedRegFreeWinRT/dllmain.cpp
@@ -442,10 +442,12 @@ HRESULT ExtRoLoadCatalog()
 
 BOOL WINAPI DllMain(HINSTANCE hmodule, DWORD reason, LPVOID /*lpvReserved*/)
 {
+    /*
     if (IsWindows1019H1OrGreater())
     {
         return true;
     }
+    */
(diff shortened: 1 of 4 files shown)


src/UndockedRegFreeWinRT/UndockedRegFreeWinRT/dllmain.cpp
```diff
@@ -442,10 +442,12 @@ HRESULT ExtRoLoadCatalog()
 
 BOOL WINAPI DllMain(HINSTANCE hmodule, DWORD reason, LPVOID /*lpvReserved*/)
 {
+    /*
```
kythant, line 445: What's the reason for this change?

ocalvo:
The commend in dllmain.cpp was added to assist testing the change. It should had been removed.
I have created a new PR based on thos one here: https://github.com/microsoft/xlang/pull/767

JesseCol:
> The commend in dllmain.cpp was added to assist testing the change. It should had been removed. I have created a new PR based on this one [here](https://github.com/microsoft/xlang/pull/767)

Yep, sorry, my bad.  I'll close this PR out since Oscar created a new one.

PR #767, opened 2021-12-09
Fix a bug where the filename variable can get clobbered due to XmlReader invalidating the buffe

ocalvo:
Applying feedback over [this](https://github.com/microsoft/xlang/pull/766) PR

The buffer returned by IXmlReader::GetValue gets invalidated after the XmlReader is used more in the future. The fileName buffer was getting re-used in a scenario where there were lots of nodes to read. The fix is simply to save it off to a local wstring.


JesseCol (approved):

kythant (approved):

BenJKuhn (approved):

BenJKuhn:
fixes #767

PR #778, opened 2022-07-27
Convert ABI build from ADO UI-based build to YAML

BenJKuhn:
This is mostly an automatic conversion with just a few tweaks for readability, and a change to add a missing signing step. There's more that could be done here.


dunhor (approved):

BenJKuhn:
Going to check this one in, then will try out the onebranch template to see how it works as a separate change.

PR #780, opened 2022-09-22
Move to OS based dispenser

### LINKED ISSUE #779: UndockedRegFreeWinrt requires the netfx3 feature on demand unnecessarily
The use of `CoCreateInstance` here:
https://github.com/microsoft/xlang/blob/be9c9c45774f945efde6c2efeaa1edcecfa175ea/src/UndockedRegFreeWinRT/UndockedRegFreeWinRT/catalog.cpp#L379

requires that the Windows feature on demand `NetFX3` be present to function.  Creation of a `CLSID_CorMetaDataDispenser` can be achieved in the same manner as the OS using [MetaDataGetDispenser](https://learn.microsoft.com/en-us/windows/win32/api/rometadata/nf-rometadata-metadatagetdispenser).  This is part of the WinRT core, and so is always present.

JohnMcPMS:
Fixes #779

## Change
Use `MetaDataGetDispenser`, as the OS does, to leverage a built-in version of the `IMetaDataDispenser`.

## Validation
I have manually validated this change within another project.

### DIFF
--- a/src/UndockedRegFreeWinRT/ManifestParserTest/ManifestParserTest.vcxproj
+++ b/src/UndockedRegFreeWinRT/ManifestParserTest/ManifestParserTest.vcxproj
@@ -99,7 +99,7 @@
       <EnableCOMDATFolding>true</EnableCOMDATFolding>
       <OptimizeReferences>true</OptimizeReferences>
       <GenerateDebugInformation>true</GenerateDebugInformation>
-      <AdditionalDependencies>comsuppw.lib;shlwapi.lib;xmllite.lib;runtimeobject.lib;Pathcch.lib;%(AdditionalDependencies)</AdditionalDependencies>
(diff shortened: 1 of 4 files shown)


JohnMcPMS:
Additional validation: This change has been in use by WinAppSDK's fork: microsoft/windowsappsdk#1633

JohnMcPMS:
Anything I can do to help expedite review of this change?  It will be easier long term if we can pull it in through the official branch rather than making our own change downstream.

BenJKuhn (approved):

BenJKuhn (approved):

kythant:
image
S360 notice asks us to disable triggers from forks. This will need to be run manually.

JohnMcPMS:
I made another change here in the event that you wouldn't accept it with the test failures in the latest build.  As pointed out to me by @ryfu-msft , we had debugged this earlier in our usage.

For Release builds on x64, a void function won't always set `RAX` to 0 by the time it returns (I don't know that Debug builds have to, but they seem to in VS).  With `PreserveSig=false`, .Net is converting that errant `RAX` value into an HRESULT and throwing if it isn't 0.

PR #787, opened 2022-12-21
User/benkuhn/urfw buildconfig

BenJKuhn:
Update targets to allow custom build configurations to opt-inot RegFreeWinRT support by setting the URFW_Configuration property.

E.g. if you have a config called test:
    <Configurations>Debug;Release;Test</Configurations>

You’d add this below it:
    <URFW_Configuration Condition="'$(Configuration)' == 'Test'">Release</URFW_Configuration>

### DIFF
--- a/src/UndockedRegFreeWinRT/Nuget/Microsoft.Windows.UndockedRegFreeWinrt.targets
+++ b/src/UndockedRegFreeWinRT/Nuget/Microsoft.Windows.UndockedRegFreeWinrt.targets
@@ -1,5 +1,7 @@
 <Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   
+  <!-- To suport custom configruations, only test against Debug or !Debug. 
+    Any other configuration is assumed to be a release configuration -->
   <ItemDefinitionGroup>
     <Link>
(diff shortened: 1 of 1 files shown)


src/UndockedRegFreeWinRT/Nuget/Microsoft.Windows.UndockedRegFreeWinrt.targets
```diff
@@ -1,39 +1,43 @@
 <Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+
+  <PropertyGroup Condition="'$(URFW_Configuration)' == ''">
+    <URFW_Configuration>$(Configuration)</URFW_Configuration>
```
Scottj1s, line 4: could provide failsafe defaults here:
```
<URFW_Configuration Condition="'$(Configuration)' == 'Debug'">Debug</URFW_Configuration>
<URFW_Configuration Condition="'$(Configuration)' != 'Debug'">Release</URFW_Configuration>
```

BenJKuhn: the problem is I think this was causing problems with some projects. I had something before that didn't require an opt-in but it was failing tests. I'll explore just a bit more to see if I can figure out what was going wrong in that case.

PR #798, opened 2025-03-28
Onboard abi to static analysis with msbuild

DefaultRyan:
I'm onboarding the abi package to get static analysis, prefast, TSA bug filing, and all that goodness.

Problem is, it seems like a major pain to set up prefast to run on a non-msbuild workflow (e.g. cmake) and then merge all the output sarif files into a way that our other pipeline infrastructure can digest it.

So, I decided to reverse-engineer all the cmake build options we are using to create an msbuild version.

Bonus, I've offloaded the string literal processing (scrape a folder for *.h files and turn that into strings.h/strings.cpp) into an msbuild target. This is nice as it removes the need for a separate "prebuilt" project. (Technically, I probably could have simplified the target by using a Powershell script to do the scraping, but the custom target would have remained).

### DIFF
--- a/.pipelines/OneBranch.Official.yml
+++ b/.pipelines/OneBranch.Official.yml
@@ -47,8 +47,6 @@ extends:
         compiled:
           enabled: true
         tsaEnabled: true
-      prefast:
-        enabled: true
     
     nugetPublishing:
       feeds:
--- a/.pipelines/OneBranch.Official.yml
+++ b/.pipelines/OneBranch.Official.yml
@@ -37,8 +37,18 @@ extends:
       enabled: false
     
     globalSdl:
+      isNativeCode: true
+      asyncSdl:
+        enabled: true
       tsa:
-        enabled: false
+        enabled: true
+        useDynamicRouting: true
+      codeql:
+        compiled:
+          enabled: true
+        tsaEnabled: true
+      prefast:
+        enabled: true
     
     nugetPublishing:
       feeds:
--- a/src/tool/abi/abi.sln
+++ b/src/tool/abi/abi.sln
@@ -1,0 +1,31 @@
+﻿
+Microsoft Visual Studio Solution File, Format Version 12.00
+# Visual Studio Version 17
+VisualStudioVersion = 17.13.35913.81 d17.13
+MinimumVisualStudioVersion = 10.0.40219.1
+Project("{8BC9CEB8-8B4A-11D0-8D11-00A0C91BC942}") = "abi", "abi.vcxproj", "{F025D704-CA01-4C93-9097-D0C7BF9CCA07}"
+EndProject
+Global
+	GlobalSection(SolutionConfigurationPlatforms) = preSolution
+		Debug|x64 = Debug|x64
+		Debug|x86 = Debug|x86
+		Release|x64 = Release|x64
+		Release|x86 = Release|x86
+	EndGlobalSection
+	GlobalSection(ProjectConfigurationPlatforms) = postSolution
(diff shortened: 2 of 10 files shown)


.pipelines/OneBranch.Official.yml
```diff
@@ -37,8 +37,16 @@ extends:
...
+      asyncSdl:
+        enabled: true
       tsa:
-        enabled: false
+        enabled: true
+        useDynamicRouting: true
+      codeql:
+        compiled:
+          enabled: true
+        tsaEnabled: true
```
manodasanW, line 49: repeated?

manodasanW: nvm, this is different.

.pipelines/OneBranch.Official.yml
```diff
@@ -37,8 +37,16 @@ extends:
       enabled: false
     
     globalSdl:
+      isNativeCode: true
+      asyncSdl:
+        enabled: true
       tsa:
-        enabled: false
+        enabled: true
```
vineeththomasalex, line 44: Does this work?

.pipelines/OneBranch.Official.yml
```diff
@@ -37,8 +37,16 @@ extends:
       enabled: false
     
     globalSdl:
+      isNativeCode: true
```
vineeththomasalex, line 40: What does isNativeCode do? I'm having a hard time finding this in the docs

manodasanW (approved):

src/tool/abi/abi.sln
```diff
@@ -0,0 +1,31 @@
...
+	EndGlobalSection
+	GlobalSection(ProjectConfigurationPlatforms) = postSolution
+		{F025D704-CA01-4C93-9097-D0C7BF9CCA07}.Debug|x64.ActiveCfg = Debug|x64
+		{F025D704-CA01-4C93-9097-D0C7BF9CCA07}.Debug|x64.Build.0 = Debug|x64
+		{F025D704-CA01-4C93-9097-D0C7BF9CCA07}.Debug|x86.ActiveCfg = Debug|Win32
+		{F025D704-CA01-4C93-9097-D0C7BF9CCA07}.Debug|x86.Build.0 = Debug|Win32
+		{F025D704-CA01-4C93-9097-D0C7BF9CCA07}.Release|x64.ActiveCfg = Release|x64
+		{F025D704-CA01-4C93-9097-D0C7BF9CCA07}.Release|x64.Build.0 = Release|x64
+		{F025D704-CA01-4C93-9097-D0C7BF9CCA07}.Release|x86.ActiveCfg = Release|Win32
+		{F025D704-CA01-4C93-9097-D0C7BF9CCA07}.Release|x86.Build.0 = Release|Win32
```
vineeththomasalex, line 23: Do we need to think about arm64 or is that irrelevant?

vineeththomasalex (approved):

.pipelines/OneBranch.Official.yml
DefaultRyan: We'll kick off an official build and find out. At least the work is likely complete from our end.

src/tool/abi/abi.sln
DefaultRyan: The build pipeline for the abi tool is specifically only compiling x86. I'm not aware of any infrastructure that is consuming this on arm64, so irrelevant for now. Adding it to the msbuild solution is easy - the bigger pain would be beefing up all the rest of the build and packaging pipelines infrastructure.

.pipelines/OneBranch.Official.yml
DefaultRyan: Replied offline with a docs link.

